PR #8, opened 2011-05-02
Passthrough Extension added

navilan:
Hi,

I have added a passthrough extension similar to [asciidoc's passthrough syntax](http://www.methods.co.nz/asciidoc/userguide.html#X76).

I have added a couple of doctests as well. Not sure if it is sufficient for acceptance. Please let me know if there is something that is missing.

Thanks
Lakshmi.

### DIFF
--- a/markdown/extensions/passthrough.py
+++ b/markdown/extensions/passthrough.py
@@ -1,0 +1,89 @@
+#!/usr/bin/env python
+
+"""
+Passthrough extension for Python Markdown
+=========================================
+
+This extension adds passthrough Blocks to Python-Markdown. Passthrough blocks
+are not transformed by the markdown processor at all.
+
+    >>> import markdown
+    >>> text = '''
+    ... A paragraph before a *passthrough* block:
+    ...
+    ... +++++
+    ... *passed through*
+    ... +++++
+    ... '''
+    >>> html = markdown.markdown(text, extensions=['passthrough'])
+    >>> html
+    u'<p>A paragraph before a <em>passthrough</em> block:</p>\\n*passed through*'
+
+Works with safe_mode also (we check this because we are using the HtmlStash):
+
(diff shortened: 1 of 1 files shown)


waylan:
I'm inclined to deny this request. It seems to me that this should exist as a third party extension. Markdown already has the mechanisms in place to do this sort of thing. Raw HTML can already be passed through and markdown syntax can be escaped. For more complex stuff, it would seem that the specific implementor should be in the best position to develop a solution that works for his/her specific needs. We provide the hooks for that purpose.

Besides, I'm not really interested is adding more nonstandard markdown syntax that I will have to continue to support. The existing extensions (mostly) all implement common markdown syntax additions which are found in multiple implementations of Markdown (I've considered removing the few that don't). I'm not aware that asciidoc's passthrough syntax is used in any Markdown implementation.

navilan:
Hi,

Thanks for the detailed response. Makes perfect sense. 

I added the passthrough processor to circumvent a situation where Markdown's raw html pass through broke down. I will try to reproduce it with the smallest possible html when I get a chance.

Thanks Again
Lakshmi

waylan:
Thanks for understanding. And please do report any problems with our
raw html pass through.

On Mon, May 2, 2011 at 10:01 AM, lakshmivyas
reply@reply.github.com
wrote:

> Hi,
> 
> Thanks for the detailed response. Makes perfect sense.
> 
> I added the passthrough processor to circumvent a situation where Markdown's raw html pass through broke down. I will try to reproduce it with the smallest possible html when I get a chance.
> 
> Thanks Again
> Lakshmi
> 
> ## 
> 
> Reply to this email directly or view it on GitHub:
> https://github.com/waylan/Python-Markdown/pull/8#issuecomment-1088054

## 

---

\X/ /-\ `/ |_ /-\ ||
Waylan Limberg

PR #35, opened 2011-07-25
A tiny edit that allows "+" and "|" in the table separator row

mfiers:
A tiny edit that allows both "+" and of "|" in the separator row of a table. 

This is useful (to me) because it allows me to use org-mode in emacs to edit tables.

Uncertain if this break compliance with markdown - I could not find an official definition for table syntax.


waylan:
We only support the table syntax implemented and  defined by PHP Markdown Extra [here](http://michelf.com/projects/php-markdown/extra/#table). For any other syntax you can create you're own extension (which shouldn't be to hard for you seeing you've already provided a patch). This is closed wontfix. Sorry.

PR #45, opened 2011-10-05
Adding an additional configuration option to TOC for numbered lists

Wilfred:
I often find myself wanting numbering in lists, so here's new configuration option to TOC (should be backwards-compatible by defaulting to False).


waylan:
We are currently in alpha state. That means no new features. Perhaps after 2.1.0-final is released I'll think about it, but I'm inclined to reject this as feature creep - especially given that it is easy for anyone to write their own extension which adds whatever feature they want. Adding the feature here just means one more thing I have to maintain.

Wilfred:
No problem. Thanks for the prompt response.

PR #50, opened 2011-11-02
Allow specifying the full module name for extensions.

nud:
Before this patch, it was possible to pass a list of extensions and
parameters to use when instantiating a new Markdown instance, but it was
not possible to give a full module name; the extension had to be a
submodule in markdown.extensions.

Now we allow giving a full module path, to make it easy to use custom
extensions which are not bundled within the markdown package. The
previous behaviour has been retained when there is no '.' in the
extension name.

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -184,20 +184,21 @@ class Markdown:
             pairs = [x.split("=") for x in ext_args.split(",")]
             configs.update([(x.strip(), y.strip()) for (x, y) in pairs])
 
-        # Setup the module names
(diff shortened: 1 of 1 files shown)


nud:
As for the comment left on gitorious, I think this patch is superior to supporting "old" mdx_\* extensions because you can use fully-qualified module names, and so you can use modules contained in your own top-level module rather than polluting the python root. Better for self-containment.

It's also closer to what you would do if you were to import extensions manually in python as the module name is basically the same.

Issue #53, opened 2011-11-16
Support fenced code block within lists and blockquotes


waylan:
As per [this discussion](http://six.pairlist.net/pipermail/markdown-discuss/2011-November/002389.html) on the markdown list, we should support fenced code blocks inside lists and blockquotes. Currently, they only work at the document root.

While we're at it, we might add support for github's syntax as an alternative??

waylan:
Just a reminder to myself: Lots of test cases in [these](https://github.com/michelf/php-markdown/issues/9) [bug](https://github.com/michelf/php-markdown/issues/10) [reports](https://github.com/michelf/php-markdown/issues/11).

rafalp:
Anything new on this issue? I'm using this lib in my project and allowing users to quote messages with fenced code blocks is pretty important for me. Maybe directions how can this be handled by custom extension?

cgabard:
Hello,

I have the same issue in a personal project and solved it with a dirty hook : change the BlockParser.parseChunk() method with an extension. This hook allow some preprocessor extensions, in this case only fenced_code, to process text blocks before there are split arount blank line. 

A cleaner way will be to have a new kind of extension, or add an configuration option to preprocessor extension, which allow their to be processed on each texts block.

I can submit you a PR if I know how you want I implemented this.

waylan:
@cgabard thanks for the input. Although I have a plan. Some time ago (around the time I filed this issue) I made a small change to the BlockParser (b3e0359) which will allow fenced code blocks to be implemented as a BlockProcessor.

While BlockProcessors generally only work on one block at a time, they have access to all remaining (unprocessed) blocks in the document. Given the above mentioned change, a BlockProcessor's 'run' method could step through all remaining blocks until the end of the fenced code block is found. If there is no end to the fenced code block, the `run` method would return `False` and the BlockParser would move on just as is the `test` method had returned `False`. 

The tricky part is when the fenced code is nested in a blockquote, list, etc. If we reach the end of the nested level, but haven't found the end of the fenced code block, then no fenced code block. That means keeping better track of nesting (how many levels deep, etc) which means the existing BlockProcessors probably need some refactoring to. And while I'm at it, those BlockProcessors might as well be refactored to use the new API mentioned above (it would greatly simplify a few of them). Its that last part that has been holding this up. Refactoring the BlockProcessors is a major task which I just don't have a big enough block of spare time to tackle.

Of course, if someone else wanted to give it a go, I'd be willing to mentor them through it.

[...]

PR #72, opened 2012-01-14
Fixing problem with links inside of < > brackets in safe mode.

mdirolf:
There's a test case that better explains what was going wrong.

Also, a fix for an issue introduced with the logic inversion in commit 35930e0928e19...


mdirolf:
Just added another commit - a2377e1 - that fixes an issue where unclosed block-level html elements would stop further processing of markdown. This seems like correct behavior when passing-through html, but not when escaping it. The change turns off html processing when safe mode is set to 'escape'. In that case I think we just want to escape the text and be done with it.

Feedback?

waylan:
Hey mdrilf, 

Thanks for the work on this. However, in the future, could you keep unrelated issues in separate pull requests? Some of the stuff here is a no-brainier, but others I need to look at closer. Everything all grouped together makes that harder to do. I guess what I'm trying to say is that less work for me gets your requests merged faster.

I'll need to play with 425fde1 and compare with other implementations. [Babelmark](http://babelmark.bobtfish.net/) makes this fairly easy.

Regarding 542324b, is that just a documentation edit? Was the actual logic changed previously, Should it not have been?

a2377e1 is an interesting solution I hadn't considered. IIRC, the current behavior is there for historical reasons. We didn't always use ElementTree (which gives us escaping for free). No reason why we can't take advantage of that now. Question though, based on your commit message, is sounds like your patch is just masking a bug in the raw html parser which will still manifest itself when we don't use `safe_mode='escape'`. If so, we need to fix that. Could you provide more info (preferably in a separate bug report).

I'll probably just merge your commits on the fenced code blocks, but those really should be part of issue #53. At least those bugs will be fixed while waiting for the refactor and the tests will help in doing the refactor as well.

mdirolf:
Hi Waylan,

Thanks for all of your hard work on Python-Markdown. We are using it
at https://fiesta.cc

Replies are inline:

> Thanks for the work on this. However, in the future, could you keep unrelated issues in separate pull requests? Some of the stuff here is a no-brainier, but others I need to look at closer. Everything all grouped together makes that harder to do. I guess what I'm trying to say is that less work for me gets your requests merged faster.

Yes I will try to - GH was yelling at me for trying to create a second
pull request but maybe I just didn't search around hard enough.

> I'll need to play with 425fde1 and compare with other implementations. [Babelmark](http://babelmark.bobtfish.net/) makes this fairly easy.

I think the test case is definitely demonstrating proper behavior
here, but not sure that my fix is correct. I was fumbling around a bit
trying to figure out the best way to accomplish it.

> Regarding 542324b, is that just a documentation edit? Was the actual logic changed previously, Should it not have been?

That is a bug fix: commit 35930e0928e19e37f81c906d5d11dfcc1087092b changed
-        if netloc != '' or scheme in locless_schemes:
-        if netloc == '' or scheme not in locless_schemes:

In order to properly reverse the logic it needs to be an _and_ I
think. There was actually a failing test on my machine (python 2.6)
when I checked out that change.

> a2377e1 is an interesting solution I hadn't considered. IIRC, the current behavior is there for historical reasons. We didn't always use ElementTree (which gives us escaping for free). No reason why we can't take advantage of that now. Question though, based on your commit message, is sounds like your patch is just masking a bug in the raw html parser which will still manifest itself when we don't use `safe_mode='escape'`. If so, we need to fix that. Could you provide more info (preferably in a separate bug report).

I'm not sure that it's masking a bug. The problem is basically that if
I have the following markdown:

```
<div>
*bold*
```

In the non-safe-mode case, I'm trying to inline html - so the _bold_
shouldn't get processed since it's inside of an unclosed <div>. In the
safe-mode case, the <div> is basically meaningless, so _bold_ should
turn into <b>bold</b>. Does that help explain the problem?

> I'll probably just merge your commits on the fenced code blocks, but those really should be part of issue #53. At least those bugs will be fixed while waiting for the refactor and the tests will help in doing the refactor as well.

Yeah sorry for the poor protocol - we needed all of those changes fast
so I took some shortcuts in reporting them back to you. Will try to be
better in the future.

waylan:
Mike,

My replied to your replies are inline.

On Wed, Jan 18, 2012 at 3:53 PM, Mike Dirolf  wrote:

> Hi Waylan,
> 
> Thanks for all of your hard work on Python-Markdown. We are using it
> at https://fiesta.cc
> 
> Replies are inline:
> 
> > Thanks for the work on this. However, in the future, could you keep unrelated issues in separate pull requests? Some of the stuff here is a no-brainier, but others I need to look at closer. Everything all grouped together makes that harder to do. I guess what I'm trying to say is that less work for me gets your requests merged faster.
> 
> Yes I will try to - GH was yelling at me for trying to create a second
> pull request but maybe I just didn't search around hard enough.

Ah right, that weird GH limit of only one pull request per user per
project. I always forget about it - probably because I don't
understand it.

> > I'll need to play with 425fde1 and compare with other implementations. [Babelmark](http://babelmark.bobtfish.net/) makes this fairly easy.
> 
> I think the test case is definitely demonstrating proper behavior
> here, but not sure that my fix is correct. I was fumbling around a bit
> trying to figure out the best way to accomplish it.

Thanks, I'll take a closer look later tonight.

> > Regarding 542324b, is that just a documentation edit? Was the actual logic changed previously, Should it not have been?
> 
> That is a bug fix: commit 35930e0928e19e37f81c906d5d11dfcc1087092b changed
> -        if netloc != '' or scheme in locless_schemes:
> -        if netloc == '' or scheme not in locless_schemes:
> 
> In order to properly reverse the logic it needs to be an _and_ I
> think. There was actually a failing test on my machine (python 2.6)
> when I checked out that change.

No, there were 2 changes to that line. Both sides of the or were
negated. It is still an _or_.  Besides, I see no code change in your
commit so  this a documentation edit - an inaccurate one unless I'm
missing something (like your failing test??). I'll need more info
before doing anything with this.

> > a2377e1 is an interesting solution I hadn't considered. IIRC, the current behavior is there for historical reasons. We didn't always use ElementTree (which gives us escaping for free). No reason why we can't take advantage of that now. Question though, based on your commit message, is sounds like your patch is just masking a bug in the raw html parser which will still manifest itself when we don't use `safe_mode='escape'`. If so, we need to fix that. Could you provide more info (preferably in a separate bug report).
> 
> I'm not sure that it's masking a bug. The problem is basically that if
> I have the following markdown:
> 
>    <div>
>    _bold_
> 
> In the non-safe-mode case, I'm trying to inline html - so the _bold_
> shouldn't get processed since it's inside of an unclosed <div>. In the
> safe-mode case, the <div> is basically meaningless, so _bold_ should
> turn into <b>bold</b>. Does that help explain the problem?

Thanks, got it now. This change will actually speed up the markdown
parser when `safe_mode='escape'` as parsing the raw html is a
performance suck. Optimizations that avoid running code unnecessarily
are always good.

> > I'll probably just merge your commits on the fenced code blocks, but those really should be part of issue #53. At least those bugs will be fixed while waiting for the refactor and the tests will help in doing the refactor as well.
> 
> Yeah sorry for the poor protocol - we needed all of those changes fast
> so I took some shortcuts in reporting them back to you. Will try to be
> better in the future.

I understand. Thanks for the feedback.

mdirolf:
Great - sounds like all of the commits are fine except one still needs
discussion, see below:

> > > Regarding 542324b, is that just a documentation edit? Was the actual logic changed previously, Should it not have been?
> > 
> > That is a bug fix: commit 35930e0928e19e37f81c906d5d11dfcc1087092b changed
> > -        if netloc != '' or scheme in locless_schemes:
> > -        if netloc == '' or scheme not in locless_schemes:
> > 
> > In order to properly reverse the logic it needs to be an _and_ I
> > think. There was actually a failing test on my machine (python 2.6)
> > when I checked out that change.
> 
> No, there were 2 changes to that line. Both sides of the or were
> negated. It is still an _or_.  Besides, I see no code change in your
> commit so  this a documentation edit - an inaccurate one unless I'm
> missing something (like your failing test??). I'll need more info
> before doing anything with this.

Yeah but this is boolean logic, right? To negate (A or B) we need (!A
and !B) - your change made it (!A or !B). The failing test case wasn't
a new one I wrote, but one of the existing tests - can you try running
them on Python 2.6?

waylan:
On Wed, Jan 18, 2012 at 4:40 PM, Mike Dirolf  wrote:

> Great - sounds like all of the commits are fine except one still needs
> discussion, see below:
> 
> > > > Regarding 542324b, is that just a documentation edit? Was the actual logic changed previously, Should it not have been?
> > > 
> > > That is a bug fix: commit 35930e0928e19e37f81c906d5d11dfcc1087092b changed
> > > -        if netloc != '' or scheme in locless_schemes:
> > > -        if netloc == '' or scheme not in locless_schemes:
> > > 
> > > In order to properly reverse the logic it needs to be an _and_ I
> > > think. There was actually a failing test on my machine (python 2.6)
> > > when I checked out that change.
> > 
> > No, there were 2 changes to that line. Both sides of the or were
> > negated. It is still an _or_.  Besides, I see no code change in your
> > commit so  this a documentation edit - an inaccurate one unless I'm
> > missing something (like your failing test??). I'll need more info
> > before doing anything with this.
> 
> Yeah but this is boolean logic, right? To negate (A or B) we need (!A
> and !B) - your change made it (!A or !B). The failing test case wasn't
> a new one I wrote, but one of the existing tests - can you try running
> them on Python 2.6?

Right. I was thinking I changed (!A or B) to (A or !B). Anyway, your
patch fixes it. Not sure how I missed that.

mdirolf:
Okay glad we got that figured out - was worried I had done something
wrong there.

On Wed, Jan 18, 2012 at 20:38, Waylan Limberg
reply@reply.github.com
wrote:

PR #74, opened 2012-01-23
Adding support for MySQL style tables to table extension

mdirolf:
Adding support for tables that look like this:

```
+------------+--------------+------+
| Field      | Type         | Null |
+------------+--------------+------+
| username   | varchar(255) | NO   |
| password   | varchar(255) | NO   |
+------------+--------------+------+
```

The differences from the currently support format are the addition of the top and bottom lines, and the use of the `+` on borders. This would be an extension of the format supported by PHP Markdown extra, but I think it's worthwhile and the implementation is simple.

### DIFF
--- a/markdown/extensions/tables.py
+++ b/markdown/extensions/tables.py
@@ -23,14 +23,24 @@ class TableProcessor(markdown.blockprocessors.BlockProcessor):
 
     def test(self, parent, block):
         rows = block.split('\n')
-        return (len(rows) > 2 and '|' in rows[0] and 
-                '|' in rows[1] and '-' in rows[1] and 
-                rows[1].strip()[0] in ['|', ':', '-'])
+        # PHP extra format
+        if (len(rows) > 2 and '|' in rows[0] and
+            '|' in rows[1] and '-' in rows[1] and
+            rows[1].strip()[0] in ['|', ':', '-']):
+            return True
+        # MySQL format
+        if (len(rows) > 4 and '+' in rows[0] and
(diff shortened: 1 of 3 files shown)


waylan:
I can appreciate your desire to have this functionality, however, this is one of many table proposals I've seen over the years. Rather than choosing one over another I have always denied every request. We have provided the API which allows anyone to maintain their own table syntax which works for them.

So why do we have the existing table extension? Only because we are providing an **exact** (or as close as possible) copy of PHP's implementation. That way we can say 'we support extra also.' 

Personally, I'm of the opinion that all tables should be in raw html. A position which is supported by the [Inline HTML](http://daringfireball.net/projects/markdown/syntax#html) section of the Syntax Documentation, which states in part:

> Markdown is not a replacement for HTML, or even close to it. Its syntax is very small, corresponding only to a very small subset of HTML tags. The idea is not to create a syntax that makes it easier to insert HTML tags. In my opinion, HTML tags are already easy to insert. The idea for Markdown is to make it easy to read, write, and edit prose. ... For any markup that is not covered by Markdown’s syntax, you simply use HTML itself. 

As I see it, tables are not "prose," and therefore outside of Markdown's scope. I realize other's disagree (as they are entitled), but, as stated above, we offer an API making any syntax possible.  That and we have to draw the line somewhere to avoid feature creep. The Syntax rules seem to support the current position of that line IMO.

mdirolf:
Sort of expected an answer like that, but thought that since [we](https://fiesta.cc) needed the code anyway I'd let you decide if you wanted to merge it in.

Cheers!

PR #88, opened 2012-03-26
Support width and height image attributes

mkomitee:
There are a few things about this extension that I'm not 100% certain I did right. It appears that in an extension, a class that inherits from `markdown.inlinepatterns.ImageReferencePattern`, `self.markdown` points at the extension, not the markdown instance, so things like `self.markdown.references` isn't available. I had to create a `dict` for references within the extension itself. That's only odd because in a class that inherits from `markdown.preprocessors.Preprocessor`, `self.markdown` does point at the markdown instance, so I had to directly address my extensions references `dict` as a class property. It's an inconsistent behavior, which I'm not certain is intended.

Additionally while I was able to leverage `self.sanitize_url` in my reference pattern class, the fact that `self.markdown` points to the extension meant when `sanitize_url` wanted to check `self.markdown.safeMode`, it couldn't unless I added a safeMode property to my extension class. That seemed inappropriate, but to be safe, I hard set it to `True`.

If I went about this all wrong, please let me know. It's got some basic docs, and doctests which pass.

### DIFF
--- a/markdown/extensions/image_attr.py
+++ b/markdown/extensions/image_attr.py
@@ -1,0 +1,175 @@
+#!usr/bin/python
+
+'''
+Image width and height attributes
+=================================
+
+This extension adds the capability to set image size attrivutes in markdown
+when using reference style images.
+
+Currently it supports single or double quoted titles, just like regular
+reference style images. Unlike standard markdown, it does not support
+parenthesized titles.
+
+Basic usage:
+
+    >>> import markdown
+    >>> text = """
+    ... ![image][]
+    ... [image]: http://example.net/image.jpg height=800 width=600
+    ... """
+    >>> html = markdown.markdown(text, ['image_attr'])
+    >>> print html
+    <p><img alt="image" height="800" src="http://example.net/image.jpg" width="600" /></p>
+
+Single quoted title:
+    >>> import markdown
+    >>> text = """
+    ... ![image][]
+    ... [image]: http://example.net/image.jpg 'title' height=800 width=600
+    ... """
+    >>> html = markdown.markdown(text, ['image_attr'])
+    >>> print html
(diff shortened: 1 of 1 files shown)


waylan:
Interesting extension. But why should we ship this with Python-Markdown? Do other markdown implementations ship with a similar ability. Does the syntax used have precedence elsewhere? Is there any reason why this can't just be a third party extension?

Please understand, I'm not actually opposed to this extension - it looks useful, but I'm trying to avoid feature creep. We already have more extensions had I'd like. Adding more only adds more maintenance. I'm more inclined to leave this as a third party extension. Perhaps down the road if it becomes wildly popular, we could revisit the question of whether to include it with Python-Markdown.

mkomitee:
It's fairly similar to the syntax used by multimarkdown, though I know that markdown and mmd are not one and the same.

Since all I'm really interested in is image sizes, if I implemented the syntax used by the extensions that ship with Discount (http://www.pell.portland.or.us/~orc/Code/markdown/), would you be more likely to accept them? 

I'd also be open to implementing the shortcut syntax from MMD "[image]: http://path.to/image "Image title" 40x400" (from http://code.google.com/p/pandoc/wiki/ImageSizes). 

That url mentions a syntax that's been thrown around on the markdown discuss mailing list "[image]: http://path.to/image "Image title" {width=40px height=400px}" which I'd be open to implementing as well.

If not, that's fine. I'd be fine keeping it as a separate extension. I just thought others may find it useful, and I doubt it would be found if it wasn't bundled. Is there a suggested way to publish extensions to python-markdown?

waylan:
I'm going to close this for the reasons stated previously. If the extension become popular in the future we can reconsider adding it then. You can add it to the [list](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) on the wiki so others can find it if you want.

PR #95, opened 2012-04-16
Attempted fix for issue 87

cataliniacob:
I'm not familiar with the code base so please review carefully. I did run the tests after adding enable_attributes=False to the [DEFAULT] section of the following files:
- tests/extensions/extra/test.cfg
- tests/extensions/test.cfg
- tests/html4/test.cfg
- tests/options/test.cfg
- tests/php/test.cfg
- tests/pl/Tests_2004/test.cfg
- tests/pl/Tests_2007/test.cfg
- tests/safe_mode/test.cfg

Without the patch I got 17 failures, with patch there are 0 failures.

### DIFF
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -304,14 +304,15 @@ class InlineProcessor(Treeprocessor):
                 if child.getchildren():
                     stack.append(child)
 
-            if self.markdown.enable_attributes:
-                for element, lst in insertQueue:
+            for element, lst in insertQueue:
+                if self.markdown.enable_attributes:
                     if element.text:
                         element.text = \
                             inlinepatterns.handleAttributes(element.text, 
                                                                     element)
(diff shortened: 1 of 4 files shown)


waylan:
The patch looks good and wow, that was a lot of testing. If you could just add one test for this, I'll merge it. Perhaps add a file  under tests/misc/ and an appropriate entry in the tests/misc/test.cfg file to turn off attributes for that file only. Of course, the test should fail without your changes and pass with them.

cataliniacob:
Thanks for your review waylan. I've now added a test. Without the patch it fails because the output is truncated

```
--- /home/catalin/python/Python-Markdown/tests/misc/no-attributes.html
+++ actual_output.html
@@ -1,2 +1,2 @@
-<p>Regression <em>test</em> for issue 87</p>
+<p>Regression </p>
```

PR #104, opened 2012-05-31
Linenumbers

ghost:
Hello,

I've added functionality to the fenced_code extension to allow it to automatically number lines (or, crudely, statements) in a block. I've also updated the relevant documentation, tests, and written a brief test/demo document of my own, which you can see here: http://firepear.net/docs/fence-count-demo.html

I'd like to work with you to get this functionality added to your upstream.

Thank you,
Shawn

### DIFF
--- a/docs/extensions/fenced_code_blocks.txt
+++ b/docs/extensions/fenced_code_blocks.txt
@@ -63,6 +63,33 @@ The above will output:
     <pre class="html">&lt;p&gt;HTML Document&lt;/p&gt;
     </pre>
 
+You can also cause the lines of the code block to be numbered
+automatically by specifying a numbering mode, which must be preceeded
+by a semicolon:
+
+    ```{.python;lines}
+    # numbered lines of python
+    ```
+
+    ```;statements
+    # numbered statements
+    ```
+
+The modes are *lines*, *statements*, *skipblanks*, and *stmtskip*. In
+lines mode, every line is numbered. In statements mode, statements are
+numbered (so lines preceeded by a line ending with a comma or
+backslash are *not* numbered). With skipblanks, blank lines are not
+numbered, and stmtskip is the combination of statements and
+skipblanks.
+
+If a mode is given, an offset may also be given, after the mode. This
+is an integer, preceeded by a semicolon.
+
+    ```{.python;lines;37}
+    # python code, numbered starting with line 37
+    ```
+
+
 Usage
 -----
 
--- a/docs/extensions/fenced_code_blocks.txt
+++ b/docs/extensions/fenced_code_blocks.txt
@@ -36,9 +36,9 @@ Fenced Code Blocks are defined using either the syntax established in
     one-line code block
     ```
 
-Fenced code blocks can have a blank line as the first  and/or last line of a 
-code block and they can also come immediately after a list item without becoming
-part of the list.
+Fenced code blocks can have a blank line as the first and/or last line
+of a code block and they can also come immediately after a list item
+without becoming part of the list.
 
 You can define the language of the code block for use by syntax
 highlighters etc. The language will be assigned as a class attribute
--- a/docs/extensions/fenced_code_blocks.txt
+++ b/docs/extensions/fenced_code_blocks.txt
@@ -18,12 +18,11 @@ This extension is included in the standard Markdown library.
 Syntax
 ------
 
-Fenced Code Blocks are defined using the syntax established in 
-[PHP Markdown Extra][php].
+Fenced Code Blocks are defined using either the syntax established in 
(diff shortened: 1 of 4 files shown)


waylan:
Wow, you did a lot of work and put together a nice patch. Unfortunately, I don't think I will be able to use your work.
Authors of various other markdown implementations and myself have previously agreed to keep the fenced code syntax essentially the same across our different implementations (each implementation being coded in a different language). Merging your changes would be too significant a change in the syntax for my comfort.  Various common extensions of syntax have been terribly fragmented in the past and I don't want to add to the fragmentation here. Fenced code blocks already are more fragmented than I would like because others have added their own variations. That said, I see no reason why your code couldn't exist as a third party extension. Feel free to add it to the [list](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions).

Regardless of the above, I do have a few comments about your implementation you may want to consider. My first thought was that the [codehilite extension](http://packages.python.org/Markdown/extensions/code_hilite.html) that ships with Python-Markdown already offers line numbering. That extension already will highlight fenced code blocks when both extensions are enabled. However, there is currently no way to indicate whether a fenced code block should be numbered or not. That might be an idea worth exploring - if we can do so without adding more syntax to fenced code blocks. 

Second, I'm not keen on all your modes. Seems like if you want to get that fancy, let JavaScript do that on the client. But that is just my personal opinion. Personally, I see no need for more than line numbering being set on or off. Which begs the question, if you are getting so fancy with line numbers, how can you not also want code highlighting? Seems like a patch to [Pygments](http://pygments.org/) (which codehilite uses under the hood) would be more appropriate for all the fancy line numbering options.

Third, I'm not particularly fond of your output. Try copying the code block and pasting it into a text editor. Now you have to manually remove all those line numbers. Now do the same with the more popular line numbering libs (like Pygments) and you will not get any line numbers pasted with the code. There are a number of ways to do this. I suggest reading my opinions on this sort of thing [here](http://achinghead.com/syntax-highlighting-web.html) (which is a little outdated today) and, more recently, [here](https://gist.github.com/2844867).

Both of those links to my opinions presume that JavaScript code highlighting libs are being used (the first explains why I think this is a better way to go). Some time ago I started a comprehensive comparison of all such JS libs out there. I never finished, but take a look [here](http://waylan.github.com/Highlighters-JS) (I prefer and use Highlight.js). I mention it because every single one of those libs offers a different way to identify language, line numbers, and whatever other features they support. This was actually part of the discussion between some of the markdown implementors. We decided to not try to support any one way over another. All you get is to define a language which gets set as a class - unaltered (which reminds me, I need to make a few adjustments to simply my implementation - it does too much as is). You can see one such discussion [here](https://github.com/michelf/php-markdown/pull/7).

Sorry to be so long winded. And given how nice your patch is (complete with tests and docs) I hate to not be able to use it, but you picked the one issue I've spent a tremendous amount of time thinking about and working on. I have some pretty solidified opinions on the matter.

ghost:
Thank you for your consideration and for your explanation. Since you asked for my reasoning on some things, I'll try to explain.

1) How can I want "fancy" line numbers but not syntax highlighting?

I didn't really think of what I was doing as "fancy". I just tried to handle all the use cases I came up with in the past week. As to why not syntax highlighting, it is because I view the two as orthogonal. Enabling syntax highlighting in an editor doesn't also enable line numbering (or vice versa), for instance. And not everything you might want to number is actually going to be code -- lists of example commands, for instance.

2) Why not use Javascript?

To be honest, this didn't even enter my mind. I didn't know there were client-side highlighting/numbering libraries. If I had been aware of them, however, I probably still wouldn't have wanted to use them. If I write an example with line numbers so that I can easily refer to specific lines in the prose text, then making that happen on the client side seems like a bad call to me -- it should be part of the document text.

I had also not been concerned with people being able to copy and paste, because that's easily handled by having the full code appear,  unnumbered, elsewhere in a document or outside it. Again, you can see my bias toward didactic usages. Here, from my point of view, the client-side solution is the fancy one :)

3) Too many modes

I don't really disagree with you here. Rather, I don't disagree that it feels suboptimal to have 4 different switches for line numbering. I genuinely believe in the utility of the modes themselves, but in a week I couldn't come up with a more elegant way to turn them on/off. I was actually hoping you might have had a better idea for the syntax :)

What I very much did not want to do is just have an on/off switch and then have the software try to DWYM. I always favor being explicit where there is the possibility of inaccuracy, but that's just my design ethos. I also decided to go ahead and push to you with what I felt was a mildly inelegant syntax instead of fretting over it.

Thanks again. Sorry to have inadvertently hit a sore spot -- I had no idea about the factious history of markdown implementations :)

PR #134, opened 2012-08-27
toc extension: Add a 'id="toc"' attribute in toc div 

BidiX:
Hi,

I need to access to the toc DIV, to hide and show the Table of Contents. I suggest to add an id attribute with a value of 'toc'. 

So the Table of Content DIV become:  

```
<div class="toc" id="toc"\>
...
</div\>
```

I've also added this change in the test sources.

Thanks,

-- BidiX
http://notewiki.bidix.info

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -27,6 +27,7 @@ class TocTreeprocessor(markdown.treeprocessors.Treeprocessor):
 
         div = etree.Element("div")
         div.attrib["class"] = "toc"
+        div.attrib["id"] = "toc"
         last_li = None
 
         # Add title to the div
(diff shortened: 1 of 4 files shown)


waylan:
IIRC, it started out with an id=toc. However,  some people place multiple instances of the toc on a page, which would generate invalid html (an id should only occur once per page). Therefore, the id was changed to a class. As the class provides the same hook for styling etc, I see no need to ađd a id. I'm closing this wontfix.

BidiX:
Thanks for your replay. I would not thought to put multiple toc on a page. Wikipedia for example has its toc with `id="toc'" class="toc"`.

But nevertheless I get my toggle function working by enumerating element with class="toc'. It's fine.

So I agree with you, with the class attribute it is possible to reference an element. So my pull request is not relevant.

Regards,

-- BidiX
http://notewiki.bidix.info

PR #135, opened 2012-08-28
Change lists from ul (unordered) to ol (ordered).

Grahack:
Minor issue here, but I would have preferred this as the default.
Anyway appearance of ul and ol can be change to the ones of ol and ul.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -74,7 +74,7 @@ class TocTreeprocessor(markdown.treeprocessors.Treeprocessor):
                         level -= 1
 
                     if tag_level > level:
-                        newlist = etree.Element("ul")
+                        newlist = etree.Element("ol")
                         if last_li:
                             last_li.append(newlist)
                         else:


waylan:
Why should we assume that everyone shares your preference? The fact is, I have not received one complaint about the current output (unordered list) in the many years this extension has existed (that is, until now). I suspect that some people might be upset it we changed it. Just by updating to the newer (less buggy) version of markdown, they how have to alter their CSS selectors?

I think that for backward compatibility reasons, we need to leave this as is. However, it might make sense as a configuration setting that we can pass in to the extension (with the default being the status quo).

Grahack:
Agreed about backward compatibility and for the configuration setting.
This issue was about discussing a bit with people interested with the use of this extension and its subtilities (maybe I should have used the mailing list, sorry).
I found it more logical to have it ordered, so I proposed the ul->ol fix, but I'll go with ul {list-style-type: decimal}.
Thanks for this complete answer.

waylan:
if you'd like to discuss the issue, yes, the list would be where to start. Instead, you did a pull request - with a specific change. That doesn't come across as "hey let's discuss possible proposals".  

In any event, if you would like to do the work, I may merge a new pull request which adds a configuration setting to the extension allowing the user to specify that a ol be generated instead. Of course, the change should include tests and documentation and the default behavior should be unchanged.

Issue #113, opened 2012-07-10
Build failure with Python 2.5


Arfrever:
`docs/release-2.2.0.txt` says that Python 2.5 is still supported, but building with Python 2.5 fails.

```
$ python2.5 setup.py build
...
running build_docs
build/lib/markdown/odict.py:159: Warning: 'as' will become a reserved keyword in Python 2.6
Traceback (most recent call last):
  File "setup.py", line 208, in <module>
    setup(**data)
  File "/usr/lib64/python2.5/distutils/core.py", line 151, in setup
    dist.run_commands()
  File "/usr/lib64/python2.5/distutils/dist.py", line 974, in run_commands
    self.run_command(cmd)
  File "/usr/lib64/python2.5/distutils/dist.py", line 994, in run_command
    cmd_obj.run()
  File "/usr/lib64/python2.5/distutils/command/build.py", line 112, in run
    self.run_command(cmd_name)
  File "/usr/lib64/python2.5/distutils/cmd.py", line 333, in run_command
    self.distribution.run_command(command)
  File "/usr/lib64/python2.5/distutils/dist.py", line 994, in run_command
    cmd_obj.run()
  File "setup.py", line 125, in run
    import markdown
  File "build/lib/markdown/__init__.py", line 41, in <module>
    from preprocessors import build_preprocessors
  File "build/lib/markdown/preprocessors.py", line 11, in <module>
    import odict
  File "build/lib/markdown/odict.py", line 159
    except Exception as e:
                      ^
SyntaxError: invalid syntax
```

Please use `except Exception, e:` or drop support for Python 2.5.

Also please delete the following code in `setup.py`:

```
if sys.version[:3] < '2.5':
    data['install_requires'] = ['elementtree']
```

waylan:
Fixed in 34cb4827da7eee2310ab285110cdadfdfbc85993

PR #140, opened 2012-09-02
Admonitions: Initial version with insufficient tests and no docs.

slig:
This is the first commit regarding the issue #113.


slig:
(Sorry for opening another issue. I thought GH would update the other one because I referred to it using its #number).

I couldn't think of a nice way to pass the complex configs this extension requires using tuples of (key, value).

For instance:

```
config = {
    'styles': {
        'note': ('note', 'Please Note'),
        'didyouknow': ('note', 'Did you know?'),
    }
}
```

Do you have any suggestions on how to do that?

After deciding that, I can create more unit tests to be sure it's working as expected.

Also, how do I test that my extension isn't breaking the others? Is there any test that runs everything together?

Thanks!

waylan:
Looks good. A question though. What purpose does the configs serve? As I understand it, you can use custom classes whether they are defined in the configs or not. If so, why do we even need the configs?

Also regarding your tests, you'll have to create your own tests to make sure you aren't breaking anything else. I suggest looking at the existing tests for examples (I know, there are not a lot of this type).

slig:
I was scratching my own itch with that config: it was possible to have different `type` mapping to the same CSS class name. I see that the vast majority of the users won't need this, so I simplified things. If there's need for a custom behavior, the user can extend the class and override `get_class_and_title`.

Now there's no config and the `title` defaults to the capitalized `type` and `type`maps directly to a CSS class.

waylan:
I like it. Now we need some docs. If you beat me to it, great. But I'm not going to merge until we have some.

slig:
Of course. I'll do the docs as well. I want to add more tests too.

slig:
I've finally written the docs. I think I followed the existing convention, but if there's anything missing or typos, please let me know.

I've also added a new small feature to allow admonitions without a title. This doesn't break any syntax we discussed before.

Since I wanted to simplify the RE, I wrote some tests and placed them inside `test_extensions.py`. I'm not sure if the way I did it is the best way to attack this.

docs/extensions/admonition.txt
```diff
@@ -0,0 +1,75 @@
...
+This extension is included in the standard Markdown library.
+
+[rST]: http://docutils.sourceforge.net/docs/ref/rst/directives.html#specific-admonitions
+
+Syntax
+------
+
+Admonitions are created using the following syntax:
+
+    !!! [type] [optional explicit title within double quotes]
```
waylan: Let's actually put quotes around the part that is supposed to be in quotes (line 24). Other than that, the docs look good.

waylan:
Hmm, not sure about the option to not have a title. How does Rest do that? Is it even possible? I don't remember and don't have time to look right now, but if not, I'm thinking we shouldn't introduce it here.

slig:
Reading the rest docs linked on my first issue, looks like it's not possible. In rest, you can use a set of specific admonitions (`attention`, `caution`, `danger`, ...) which will use the name as a title or you can use the generic admonition (`admonition`) and use a specific title. 

Since we are already allowing the user to supply a custom title whenever he wants (inside double quotes, which means it's possible to use an explicit blank string `""`), i.e. we don't have two kinds of admonitions like rest, what I really did was not render the `<p>` in case of a empty title (instead of rendering a blank p `<p class="admonition-title"></p>`.)

It doesn't seems to diverge much more from rest and it allows the user to not use a title without having to hack a `display: none;` in the CSS. I'm also obviously biased because I have this use case.

That was my reasoning. Did I convince you? If not, we can roll back and do not allow an empty title in the regex.

brunobord:
on a pet project, I've tried out this extension by extracting its code in my own source project. it worked just fine for me. Please, can this be merged into python-markdown distribution? I wouldn't have to use this workaround to it...

waylan:
@slig if you update this, I'll merge it. I think there might be some merge conflicts in the docs.

and can we have line 24 of the docs for this extension to read:

```
    !!! type "optional explicit title within double quotes"
```

I understand why you used the brackets, but I find them confusing as they don't actually get used in the markup.

slig:
I think this is it. If you spot anything else, please let me know.

PR #148, opened 2012-10-11
patched extensions/codehilite.py to allow global line number override

noah:
Waylan,

I took the time to add this functionality while leaving the existing functionality intact for backwards compatibility.  Please take a look at `output.html` and `original_output.html` (generated by `test_patch/test.py`).

If you have any questions about what I did or why I did it, let me know.

Cheers,
noah

### DIFF
--- a/css/code.css
+++ b/css/code.css
@@ -1,0 +1,62 @@
+.codehilite .hll { background-color: #ffffcc }
+.codehilite  { background: #f8f8f8; }
+.codehilite .c { color: #408080; font-style: italic } /* Comment */
+.codehilite .err { border: 1px solid #FF0000 } /* Error */
+.codehilite .k { color: #008000; font-weight: bold } /* Keyword */
+.codehilite .o { color: #666666 } /* Operator */
+.codehilite .cm { color: #408080; font-style: italic } /* Comment.Multiline */
+.codehilite .cp { color: #BC7A00 } /* Comment.Preproc */
+.codehilite .c1 { color: #408080; font-style: italic } /* Comment.Single */
+.codehilite .cs { color: #408080; font-style: italic } /* Comment.Special */
+.codehilite .gd { color: #A00000 } /* Generic.Deleted */
+.codehilite .ge { font-style: italic } /* Generic.Emph */
+.codehilite .gr { color: #FF0000 } /* Generic.Error */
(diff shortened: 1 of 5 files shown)


waylan:
This looks like a good start. However, I don't intend to keep both `linenos` and `force_linenos` keywords around forever. It is way to confusing. So `forece_linenos` should issue a depreciation warning.

Also, the documentation needs to be updated (in the docs dir) to explain the new keyword and warn that the old one is depreciated.

And finally, the tests should be using the existing [testing framework](http://packages.python.org/Markdown/test_suite.html). And while I'm at it, there is no way I'm adding the css to markdown. It serves no useful purpose.

noah:
On Thu, Oct 11, 2012 at 9:31 AM, Waylan Limberg
notifications@github.com wrote:

> This looks like a good start. However, I don't intend to keep both linenos and force_linenos keywords around forever. It is way to confusing. So forece_linenos should issue a depreciation warning.

Well, that's your codebase so I thought I'd leave that up to you.  Let
me know what you want the warning to say and I'll put it in.

> Also, the documentation needs to be updated (in the docs dir) to explain the new keyword and warn that the old one is depreciated.

Again, you hinted previously at deprecation for `force_linenos` but
there are a number of ways to accomplish this.  Give me some more
details about how you'd like this to look.

> And finally, the tests should be using the existing testing framework. And while I'm at it, there is no way I'm adding the css to markdown. It serves no useful purpose.

I didn't look at the tests.  I'm confident my code works.  If you want
tests I'll consider writing them.  The CSS is for the benefit of the
test script I included -- I was not suggesting that you would include
it in the python-markdown dist.

Let me know and I'll `rebase` as necessary.

> —
> Reply to this email directly or view it on GitHub.

## 

Noah

waylan:
Sorry, sometimes I forget which conversations included what. Let me start
over.

I work on Python-Markdown in my spare time - which is _extremely_ limited
right now -- to the point that I have no time to work on this issue. If you
want your changes to be committed, then you need to provide me something I
can commit with no additional work. Otherwise it will likely be months
before I even look at it.

To be clear, I agree this change should be made. It just isn't a high
priority and other things will get done first if I'm not provided with a
clean pull request.

Regarding the depreciation warning, we use python's logging module already,
so the warning should probably be issued through that. I don't have time to
check, but isn't DepretiationWarning (or similar) included in the Python
standard library? I really don't have the time for more that this right now.

PR #149, opened 2012-10-19
Fix tabs being converted to spaces in the resulting HTML

### LINKED ISSUE #36: My tabs are converted into spaces
the code `→→code` (where `→` is a tab) is expanded into this:

```
    code
```

why aren’t my tabs retained?

flying-sheep:
does not alter behavior in any other way than that tabs are preserved, which you can see by [viewing the changes to the tests ignoring whitespace](https://github.com/flying-sheep/Python-Markdown/commit/21a5505d46cd48f97065b65134c5288db0acdbf2?w=1).

fixes #36

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -286,8 +286,8 @@ class Markdown:
 
         source = source.replace(util.STX, "").replace(util.ETX, "")
         source = source.replace("\r\n", "\n").replace("\r", "\n") + "\n\n"
+        #TODO: retain indented blank lines in code
         source = re.sub(r'\n\s+\n', '\n\n', source)
-        source = source.expandtabs(self.tab_length)
 
         # Split into lines and run the line preprocessors.
         self.lines = source.split("\n")
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -65,6 +65,8 @@ class BlockProcessor:
         for line in lines:
             if line.startswith(' '*self.tab_length):
                 newtext.append(line[self.tab_length:])
+            elif line.startswith('\t'):
+                newtext.append(line[1:])
             elif not line.strip():
                 newtext.append('')
             else:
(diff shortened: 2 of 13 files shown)


flying-sheep:
i also fixed another bug, namely the tab_length being hardcoded to 4 spaces in many places.

for that the regex compiling in the blockprocessors had to be deferred in order to be able to use tab_length there.

in case you want to pull both separately, i’ll move them to another branch and file a separate pull request, but why not fix two bugs in one go?

waylan:
As stated in my comments to #36, we will only be making such a change if the reference implementation does.

The Python-Markdown project has two primary goals (as I see it).
1) Match the output of the reference implementation as closely as reasonably possible given the same input.
2) Provide an [extension API](http://packages.python.org/Markdown/extensions/api.html) which would allow users to override/change/add any behavior they desire.

The only reason we would veer from the first goal is if the reference implementation's behavior in a given case is considered a bug (there are a few). I do not consider this specific behavior to be a bug. Of course, that means that we will be retaining the current behavior. However, you are free to create your own extension which has whatever behavior you want. Based on this pull request, it looks like you already have most of the work done.

As I did with issue #36, I am closing this wontfix.

waylan:
Oh, just reread your last comment. If the tablength is hardcoded anywhere, then, yes that is a bug. You may open a separate pull request with that fix only if you want. If not, I'll get to it when I have some free time.

flying-sheep:
why shouldn’t we consider it a bug if the contents of code blocks are changed? as said: that creates a syntax error in a certain programming language.

note that the changes in output are only whitespace. look at the tests: i changet not a single non-whitespace byte, so the output is matched perfectly if looked at like the original implementation looks at input code.

PS: a own extension is not feasible: the regex in every block processor has to be changed

---

OK, let’s make a deal: i’ll fix all the extensions and add a kwoption to `Markdowner`/`markdown` called `preserve_tabs`, which defaults to False. it simply reintroduces the expandtabs call, upon which the behavior will match the reference implementation’s output.

``` python
if not preserve_tabs:
    source = source.expandtabs(self.tab_length)
```

then i’ll do a pull request again and you accept it, if it has no bugs, is that acceptable?

i’d really enjoy not to ship my own version of python markdown with my projects…

waylan:
I guess I'm not familiar with this programming language you speak of.  In
fact, in all the text editors I use (my editor of choice being vim), I have
them configured to insert tablength (usually set to 4) number of spaces
when I type a tab. In other words, my code and text never contain any tabs.

For me to be convinced this is a bug, I'm going to need to see something
for specific. Also, a consensus among a majority of the other
implementation authors on the [markdown discussion list](http://six.pairlist.net/mailman/listinfo/markdown-discuss) wouldn't help -
although that is not so easy to obtain these days.

flying-sheep:
What about the part below the line?

PR #150, opened 2012-10-21
Add option to preserve tabs and fix hardcoded `tab_length` in blockprocessor regexes

flying-sheep:
since apparently changing the default behavior isn’t acceptable, but since it is important for some applications to preserve tabs in the output (as mentioned in the other bug/pull request), i added this as an option.

as you can see, behavior doesn’t change in the slightest unless `preserve_tabs=True` is passed as keyword argument: the regexes are only changed to accept tabs additionally to spaces, which doesn’t have an effect in a tab-expanded document.

i also added a test for that option.

---

please don’t outright refuse and close it again: if there are qualitative problems i can work them out, if there are other problems, we can discuss them – but i really want to be able to tell markdown to preserve tabs in a way that doesn’t involve monkeypatching all the changes you see from within a extension.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -65,6 +65,8 @@ class BlockProcessor:
         for line in lines:
             if line.startswith(' '*self.tab_length):
                 newtext.append(line[self.tab_length:])
+            elif line.startswith('\t'):
+                newtext.append(line[1:])
             elif not line.strip():
                 newtext.append('')
             else:
@@ -75,8 +77,15 @@ class BlockProcessor:
         """ Remove a tab from front of lines but allowing dedented lines. """
         lines = text.split('\n')
         for i in range(len(lines)):
-            if lines[i].startswith(' '*self.tab_length*level):
-                lines[i] = lines[i][self.tab_length*level:]
+            levels = level
+            while levels:
+                levels -= 1
+                if lines[i].startswith(' '*self.tab_length):
+                    lines[i] = lines[i][self.tab_length:]
+                elif lines[i].startswith('\t'):
+                    lines[i] = lines[i][1:]
+                else:
+                    break
         return '\n'.join(lines)
 
     def test(self, parent, block):
@@ -140,10 +149,10 @@ class ListIndentProcessor(BlockProcessor):
 
     def __init__(self, *args):
         BlockProcessor.__init__(self, *args)
-        self.INDENT_RE = re.compile(r'^(([ ]{%s})+)'% self.tab_length)
+        self.INDENT_RE = re.compile(r'^(([ ]{%s}|\t)+)'% self.tab_length)
 
     def test(self, parent, block):
-        return block.startswith(' '*self.tab_length) and \
+        return block.startswith((' '*self.tab_length, '\t')) and \
                 not self.parser.state.isstate('detabbed') and  \
                 (parent.tag in self.ITEM_TYPES or \
                     (len(parent) and parent[-1] and \
@@ -197,7 +206,7 @@ class ListIndentProcessor(BlockProcessor):
         # Get indent level
         m = self.INDENT_RE.match(block)
         if m:
-            indent_level = len(m.group(1))/self.tab_length
+            indent_level = m.group(1).count(' ')/self.tab_length + m.group(1).count('\t')
         else:
             indent_level = 0
         if self.parser.state.isstate('list'):
@@ -224,7 +233,7 @@ class CodeBlockProcessor(BlockProcessor):
     """ Process code blocks. """
 
     def test(self, parent, block):
-        return block.startswith(' '*self.tab_length)
+        return block.startswith((' '*self.tab_length, '\t'))
     
     def run(self, parent, blocks):
         sibling = self.lastChild(parent)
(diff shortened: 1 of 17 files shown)


waylan:
Please explain to me why we need to preserve tabs. Why does it matter? I would like some hard evidence, specific use cases - not just an assertion that it must be supported. As stated previously, I see **no value** in supporting this non-standard feature. If I don't see the value, I'm just going to continue to see this as feature creep.

flying-sheep:
as said, in [genie](http://en.wikipedia.org/wiki/Genie_%28programming_language%29), you need to declare beforehand, what one indentation level is. expanding tabs without replacing that specification makes the file not compile anymore.

what i want to say: it’s important that code in code blocks isn’t changed in any way.

could you please explain to me what “feature creep” is for you? in my understanding it’s restricted to code freezes, and afaik this project isn’t in a pre-release code freeze.

waylan:
> could you please explain to me what “feature creep” is for you?

I think [wikipedia](https://en.wikipedia.org/wiki/Feature_creep)
explains it well. Don't know what that has to do with code freezes.
But it is definitively something I try to avoid.

Why do I think this is feature creep? Because it unnecessarily
complicates the code. If we normalize all whitespace, we don't have to
check for both tabs and/or spaces later.

Thanks for pointing me to genie. Can't say I was aware of that.
However, given such a restrictive syntax, is it reasonable to expect
the various markup languages to get it right? Are there any other
[markdown implementations](https://en.wikipedia.org/wiki/List_of_Markdown_implementations)
which preserve tabs in code blocks? Are there any [markdown
competitors](https://en.wikipedia.org/wiki/Lightweight_markup_language)
that preserve tabs?

markdown/blockprocessors.py
```diff
@@ -75,8 +77,15 @@ def looseDetab(self, text, level=1):
...
-                lines[i] = lines[i][self.tab_length*level:]
+            levels = level
+            while levels:
+                levels -= 1
+                if lines[i].startswith(' '*self.tab_length):
+                    lines[i] = lines[i][self.tab_length:]
+                elif lines[i].startswith('\t'):
+                    lines[i] = lines[i][1:]
+                else:
+                    break
```
flying-sheep, line 88: i think the above part is the only one that got more complex, but i could make it easier or comment it, if you wish. what do you say?

flying-sheep:
generally, i’d say it’s the principle of least surprise that they are preserved – there is no reason to needlessly expand them. i understand that you want the output to match the reference implementation’s, however, thus the implementation as option.

i don’t think the code gets noticably more complex: note that other than implementing the tab preservation option, i also fixed the hardcoded `tab_width` in the regexes (which did make the code more complex, but is necessary). imho the tab-preservation -related changes are all straightforward: replacing e.g. `[ ]{%s}` with `(?:[ ]{%s}|\t)` doesn’t seem complex to me. the only part that i think gained complexity is [that one](https://github.com/waylan/Python-Markdown/pull/150/files#r1899248).

for other markup with that option: docutils’ reStructuredText (via setting negative tab_width), and pandoc (via `-p` option) for a start. i didn’t look at others, as those two seemed to me the most important/relevant. pandoc can also do markdown, so that’s a markdown implementation able to do that.

waylan:
To be clear, I understand your patch just fine. It, in itself is not
terribly complex. But it is one more thing to maintain for years to come.
And, when you add it to all the other non-standard feature requests I get,
why should this one get accepted when the others didn't. If you add all
those together, then we have something that it too complex.

Generally, I would suggest that this be implemented as a third party
extension. The only reason I haven't closed this yet is because I realize
you can't currently disable the whitespace normalization through the
extension API. Perhaps that is how this should happen?? Make whitespace
normalization a preprocessor. I'm still not sure. I'll have to think that
over a little longer.

Regarding the other examples you found. Rest is interesting to me -
especially because it is **not** the default. Also, I consider Pandoc to be
the wildcard among markdown implementations. While Pandoc's "strict mode"
is actually one of the best, Pandoc's default behavior is one of the
farthest from typical markdown behavior in my opinion. The fact that Pandoc
didn't see fit to make this a default behavior makes me think this is even
less of a concern.

Finally, in the years I've been involved in the project, this is the first
I've ever heard this request. I guess that is what is giving me pause. Why
should this very non-standard feature be accepted when many more popular
(and more common) feature requests have been denied in the past?

flying-sheep:
I wasn’t aware that there were other feature requests that came with a full implementation like mine and were rejected.

And while the behavior isn’t default in the other implementations, it’s still an option there, which is what i’d like to see in this one, too.

I’d be glad to implement this as an extension, but it’s just not feasible without changing all regexes that involve the expanded tabs, like `[ ]{%s}`. (note that the tab width makes code difficult here ;))

I have an idea: we give blockprocessor a bunch of string attributes containing those parts, after which a regex can be constructed like `re.compile(self.indent_re + r"[-+*]\s+")` instead of repeating the complex parts like `re.compile(r"[ ]{%s}[-+*]\s+" % self.tab_width)`. that would make the code easier, and then we can either find a way to replace those extracted regex parts from within an extension, or just use the one including the tab, since the code got easier in the first place :)

what i don’t have a good argument for is why you should accept this request if others have been denied, since i neither know those other requests, nor think that you denying them is a argument against accepting mine at all: feature requests should be denied if they are bad ideas, or not worth the effort. but all my changes except the mentioned one (the while loop) are trivial imho, the tests still pass, and i’ve already done all the work, so i see no reason not to include it. it’d make me happy after all :) (which is in my eyes as valid an argument as “i denied those other requests” is one)

flying-sheep:
since i haven’t heard of you for two weeks i thought you maybe overlooked the notification.

so: what do you say to my above comment?

PR #163, opened 2012-12-04
Input file not existing error reporting.

allanderek:
I have been using Python-Markdown recently and occasionally type in the incorrect input file name, resulting in:

```
  File "markdown/__main__.py", line 87, in <module>
    run()
  File "markdown/__main__.py", line 81, in run
     markdown.markdownFromFile(**options)
  File "/afs/inf.ed.ac.uk/user/a/aclark6/Python/Python-Markdown/markdown/__init__.py", line 452, in       markdownFromFile
    kwargs.get('encoding', None))
  File "/afs/inf.ed.ac.uk/user/a/aclark6/Python/Python-Markdown/markdown/__init__.py", line 357, in convertFile
    input_file = codecs.open(input, mode="r", encoding=encoding)
  File "/usr/lib64/python2.6/codecs.py", line 881, in open
    file = __builtin__.open(filename, mode, buffering)
IOError: [Errno 2] No such file or directory: 'laskdjf'
```

This is a very simple fix which will result in the slightly friendlier:

```
Input file: laskdjf does not exist.
Cannot continue, exiting.
```

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -36,6 +36,7 @@ version_info = (2,2,1, "final")
 import re
 import codecs
 import sys
+import os
 import logging
 import util
 from preprocessors import build_preprocessors
@@ -349,6 +350,10 @@ class Markdown:
         # Read the source
         if input:
             if isinstance(input, str):
+                if not os.path.exists(input) or not os.path.isfile(input):
+                  print ("Input file: " + input + " does not exist.")
+                  print ("Cannot continue, exiting.")
+                  sys.exit(1)
                 input_file = codecs.open(input, mode="r", encoding=encoding)
             else:
                 input_file = codecs.getreader(encoding)(input)


waylan:
Thanks for the suggestion and patch. However, I can't (won't) apply the patch as is. 

Python-Markdown is a library first and a commandline tool second. Many (if not most) users are using the markdown library within their web server backends - which your patch would break.

First, you used `print` which should never be used in server code (most servers don't like it when code writes to stdout). That's what logging is for. While it is not used often, there are a few calls to pythons logging framework within python-markdown's source code. Any errors you might want to display on the commandline should be written to the logging framework (which forwards it to stderror by default). 

Second, if a user of the library called `Markdown.convertFile` directly (say from their server backend) they would want the trackback, not some less detailed "pretty" error for debugging purposes (in production the server would use it's own mechanism to catch such errors). Therefore the "pretty" error should only be generated when actually called from the command line.

I suppose a try/except block in the commandline script (markdown/**main**.py) might be the way to go.?

I'm closing this pull request. Feel free to make another which addresses my concerns above.

waylan:
Oh, one more thing, never-ever call `exit(1)` from server code! From the commandline script, maybe? But it should never exist anywhere else in the code.

allanderek:
Hi, thanks for writing such a detailed explanation.

PR #171, opened 2013-01-01
fix: def_list.py raises exception for unintended : in list element.

mizuy:
Markdown source

```

- hoge
-    :    hogehoge

```

will raise exception like this:

 File "env/lib/python2.7/site-packages/markdown/extensions/def_list.py", line 51, in run
    if not terms and sibling.tag == 'p':

  'NoneType' object has no attribute 'tag'

### DIFF
--- a/markdown/extensions/def_list.py
+++ b/markdown/extensions/def_list.py
@@ -48,7 +48,9 @@ class DefListProcessor(markdown.blockprocessors.BlockProcessor):
         else:
             d = m.group(2)
         sibling = self.lastChild(parent)
-        if not terms and sibling.tag == 'p':
+        if not terms and sibling==None:
+            state = 'list'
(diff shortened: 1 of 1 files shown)


waylan:
I received the following email privately regarding this issue:

> I've spent some time tracing the problem that occurs when lists and definition list interact.
> 
> As we thought, the patch simply works around the issue which occurs upstream and, as such, is probably not a good solution. The problem is that lists are processed first and the regex does not exclude the possibility of a definition list sitting inside an item (well actually a colon being at the start). By the time the definition list processor gets hold of the block, the parser is in a state that isn't handled in the code, but it's not clear to me how that could be suitably detected and dealt with. I guess that nested a `<dl>` inside a `<ul>` or `<ol>` is valid so this should be handled somehow.

I'll have to investigate this further when I get the time.

PR #174, opened 2013-01-08
Allow for html_class to be determined dynamically based on the wikilink's label.

joshourisman:
In some cases it's desirable that the class applied to a wikilink be dynamic based on context; e.g. if you want to style wikilinks to pages that already exist differently from pages that don't yet exist. This change allows for that.

### DIFF
--- a/markdown/extensions/wikilinks.py
+++ b/markdown/extensions/wikilinks.py
@@ -125,6 +125,8 @@ class WikiLinks(markdown.inlinepatterns.Pattern):
             a.text = label 
             a.set('href', url)
             if html_class:
+                if callable(html_class):
+                    html_class = html_class(label)
                 a.set('class', html_class)
         else:
             a = ''


waylan:
I suppose that as the build_url callable has no effect on the html class, it makes sense to support this as well. However, I don't see any updates to the docs or tests in your patch. As this is pretty low on my priority list, I'm not likely to merge your request without them and have limited time to do the work myself. If this is something you care about, you might want to provide a complete patch.

joshourisman:
Oops, you're right. Wasn't really ready to submit the pull request yet...
I'll fix those up and submit a complete patch.

## 

Josh Ourisman
www.joshourisman.com
(301) 244-9674
Washington DC

waylan:
So I gave this some further thought and something doesn't add up. @joshourisman, your use case calls for changing the class based on context. Yet, your patch offers no way to pass that context to the callable. Sure, a second keyword could be added to the callable to pass in some context, but what? Seems to me the proper solution is to suggest that in such situations one should create their own extension (which could subclass  this one and override a method or two). That being the case, I'm inclined to not apply this patch.

joshourisman:
In my particular use case, I'm taking care of context by passing a class method as the callable. The class has access to all the context necessary. The other thought I had was passing it something along the lines of:

```
lambda x: 'good' if x in [...] else 'bad'
```

In either case, the needed context is inherent in the callable itself.

However, I definitely see your point. Creating my own extension was actually my first inclination, but then I saw how simple it would be to do it this way and decided to fork instead.

waylan:
Right, a class method. But how does that class obtain the url to
determine whether the page exists or not? Or is the build_url callable also
a class method from the same class which stores the url on the class
instance for later use by the html_class callable?

I have no doubt that the above would work, but that puts
more constraints on how users of Python-Markdown can organize their code.
To provide the flexibility for all use cases, I think requiring users to
create their own extension (through subclassing or otherwise) would be a
better fit. Now, if you wanted to break some of the functionality out
into separate methods to more easily override though subclassing, I may
be open to a patch that does that.

joshourisman:
I think you're probably right. The more I think about how I could make the code I wrote generic, the more I realize it probably won't work.

PR #184, opened 2013-01-29
Simpler extension mechanism

flashingpumpkin:
Hi

This is somewhat of an unsolicited pull request, but hey that's what GitHub is about. :)

I've added some code to get rid of the current ugliness of extension creation. It's backwards compatible. 

The branch registers "named" extensions, which can be placed anywhere in your code. Documentation and the tests should describe pretty well what it does. The reasoning was that it's pretty ugly to distribute root-level `mdx_*` modules. 

Anyhow. I've got another branch with a `tox.ini` file to test against different Python versions too if you're interested. 

Cheers, 
# !A

### DIFF
--- a/docs/extensions/api.txt
+++ b/docs/extensions/api.txt
@@ -594,6 +594,25 @@ For example:
 This is useful if you need to implement a large number of extensions with more
 than one residing in a module.
 
+### Named extensions
+
+If you don't want to ship your custom extension as a ``mdx_`` module,
+the third option is to name your extension class and move
+``makeExtension`` to your class as a static method:
+
+    class FootnoteExtension(markdown.Extension):
+        name = "footnotes"
+        
+        @staticmethod
+        def makeExtension(configs=None):
+            return FootnoteExtension(configs=configs)
+
+This will allow you to ship custom Markdown extensions as part of your
+package. Under the hood, :attr:`markdown.Extension` subclasses are
+registered with a meta class that keeps track of all
+`markdown.Extension` sub classes if they have a ``name`` attribute. 
+
+
 [Preprocessors]: #preprocessors
 [InlinePatterns]: #inlinepatterns
 [Treeprocessors]: #treeprocessors
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -41,7 +41,7 @@ from blockprocessors import build_block_parser
 from treeprocessors import build_treeprocessors
 from inlinepatterns import build_inlinepatterns
 from postprocessors import build_postprocessors
-from extensions import Extension
+from extensions import Extension, registry
 from serializers import to_html_string, to_xhtml_string
 
 __all__ = ['Markdown', 'markdown', 'markdownFromFile']
@@ -179,6 +179,14 @@ class Markdown:
 
         # Parse extensions config params (ignore the order)
         configs = dict(configs)
+
+        # Check if the extension was created through the meta class
+        if ext_name in registry:
+            cls = registry[ext_name]
+            assert hasattr(cls, 'makeExtension'), "'%s' requires a static method 'makeExtension'" % repr(cls)
+            return cls.makeExtension(configs.items())
+            
+
         pos = ext_name.find("(") # find the first "("
         if pos > 0:
             ext_args = ext_name[pos+1:-1]
--- a/markdown/extensions/__init__.py
+++ b/markdown/extensions/__init__.py
@@ -3,8 +3,21 @@ Extensions
 -----------------------------------------------------------------------------
 """
 
+registry = {}
+
+class ExtensionMetaClass(type):
+    def __new__(cls, name, bases, attrs):
+        Extension =  super(ExtensionMetaClass, cls).__new__(cls, name, bases, attrs)
+
+        if hasattr(Extension, 'name'):
+            registry[Extension.name] = Extension
+
+        return Extension
+
+
 class Extension:
     """ Base class for extensions to subclass. """
+    __metaclass__ = ExtensionMetaClass
     def __init__(self, configs = {}):
         """Create an instance of an Extention.
 
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -11,6 +11,39 @@ from __future__ import unicode_literals
 import unittest
 import markdown
 
+class SmilieProcessor(markdown.preprocessors.Preprocessor):
+    def run(self, lines):
+        for line in lines:
+            yield line.replace(':(', ':)')
+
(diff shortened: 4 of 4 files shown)


waylan:
Interesting solution. However, if I'm not mistaken, one would still need to import the file which contains the extension(s) to get the extension to register. True, as in your examples, defining the extension in the same file as the call to markdown eliminates the need for an import of the extension. But in either case, how is passing a string (which was registered through a meta class) better than the following (which is already possible)?

``` python
import markdown
from myextension import MyExtension

md = markdown.Markdown(extensions=[MyExtension()])
```

If anything, the above is less "[magical](http://www.b-list.org/weblog/2007/dec/03/making-magic/)". Both solutions allow multiple extensions to exist in one file (and, optionally, in the same file as the call to markdown) - so there's no lost functionality here. If anything, this removes the need for the [extension_configs](http://packages.python.org/Markdown/reference.html#extension_configs) keyword on the Markdown class. Configs can be set on each individual extension instance - which is much easier to follow six months later.

I agree that the `mdx_*` naming convention is an ugly wart (from before I joined the project). That is why I implemented support for the above method. I suppose the documentation could be updated to better reflect that as the preferred way of developing extensions. But I don't see any added benefit of adding your "magical" solution as a third way (fourth if you count the builtin extensions). In fact, from time to time I've considered updating the docs to instruct users to import the builtin extensions (`from markdown.extensions.footnotes import FootnoteExtension`) rather than use strings. That feels less like "magic" and the builtin extensions no longer appear to have preferential treatment (the current situation gives some users the impression that no support even exists for using your own third-party extensions - making such a change to the docs might help alleviate that misconception).

Regarding the tox thing; I have recently played with it myself (see my broken [tox branch](https://github.com/waylan/Python-Markdown/tree/tox)), but I haven't found a way to test in both Python 2 & Python 3 as tox does not yet support running 2to3 on the code base. Once that happens (or with a not-to-hacky workaround) I'm all for moving to tox (apparently the developer of tox [plans](https://bitbucket.org/hpk42/tox/src/f75b1cb80a59a4127401e318e0c39e487a840d0e/ISSUES.txt?at=default#cl-97)  to add support for 2to3).

flashingpumpkin:
>  However, if I'm not mistaken, one would still need to import the file which contains the extension(s) to get the extension to register. ... But in either case, how is passing a string (which was registered through a meta class) better than the following (which is already possible)?

Depends on the use case. In your example definitely not. Using `markdown` in an environment such as Django templates where one can not explicitly set up the extensions but is required to pass in extension names (modules are loaded and registered on `{% load %}`), yes:

``` django
{% load markup %}
{{ content|markup:"extension-name" }}
```

However `markdown` should not implement features because another framework or library dictates it. I've implemented the meta class extensions because shipping `mdx_*` with our Django apps and projects is ugly and `django.contrib.markup` (which is going to be removed in 1.5) gives me already a core Django way to render markdown with extensions without yet another dependency. It might be time to just write my own Markdown filters then.

> I suppose the documentation could be updated to better reflect that as the preferred way of developing extensions.  ...  In fact, from time to time I've considered updating the docs to instruct users to import the builtin extensions (from markdown.extensions.footnotes import FootnoteExtension) rather than use strings.

+1. In fact I've used strings in this implementation because the documentation was already referring to strings for extension names. It struck me as a bit weird but I just went with it.

waylan:
> Depends on the use case. In your example definitely not. Using markdown in an environment such as Django templates where one can not explicitly set up the extensions but is required to pass in extension names (modules are loaded and registered on {% load %}), yes:

Ah, right. I hadn't considered Django's requirement that all
extensions be referenced as strings. Interestingly (or not), I was
instrumental in getting the current extension support implemented in
Django. At the time, that was the only way extensions could be loaded
in markdown so Django's pattern was really just built on markdown's at
the time.

The fact is, we've kept markdown's API the same even when releasing
version 2.0 so it would continue to work with Django and the like.
When a small library like markdown is changed, it becomes a real chore
to get large frameworks like Django to follow along in a timely
manner. Fortunately, Django dropping contrib.markup will help. A
small, independently maintained reusable Django app will have a much
easier time keeping up with any changes to Markdown.

>  It might be time to just write my own Markdown filters then.

That is what I have always done. Even immediately after getting the
current support for markdown implemented in Django, I still needed to
implement my own filters for whatever specific needs I had for a given
project. I've never found a general filter to meet my specific needs.

> > I suppose the documentation could be updated to better reflect that as the preferred way of developing extensions. ... In fact, from time to time I've considered updating the docs to instruct users to import the builtin extensions (from markdown.extensions.footnotes import FootnoteExtension) rather than use strings.
> 
> +1. In fact I've used strings in this implementation because the documentation was already referring to strings for extension names. It struck me as a bit weird but I just went with it.

I agree. While you seem to have a legitimate reason for your
string-based implementation (Django's built in markdown filter), the
fact that that is reaching end-of-life makes it clear to me that this
is probably the wrong approach. I have a few ideas for a third party
reusable Django app that addresses multiple markdown use cases more
effectively - and I think that is a better way to go. Of course, if
someone else beats me to it and develops a contrib.markup replacement
first, I'm completely fine with that.

Regardless, I think the way forward is to update the docs as described
above and perhaps start to depreciate support for the old extension
loading mechanisms.

waylan:
One thing I almost forgot about is the command line. Having each extension represented as a string makes loading extensions from the command line easy. However, I think we can work around that. Python-Markdown is a library first after all. A good library API is more important that a flexible CLI IMO. 

A few possible solutions that occur to me are:

1) When string names for extensions are depreciated, rather than deleting the code, move it to the command line script. Then, the CLI will continue to work as-is. Sure, third party scripts won't be accessible (unless we keep support for the "mdx_*" naming convention - in the CLI only), but it's not that hard to write a simple python script that wraps the lib. If the user is already writing their own extensions, that shouldn't be a problem.

2) Allow the CLI to accept strings in python's dot notion (for example: `"markdown.extensions.footnotes.FootnoteExtension"`). That way, any importable extension would be accessible from the CLI. Not sure how that will work with extension specific configs though.

3) Perhaps a combination of 1 & 2. Implement 1 only for built-in extensions and 2 for third party extensions.

Personally I like 2 the best. I'm inclined to just ignore config options. From the command line you only get the defaults - or you can write your own script.

waylan:
So I just noticed that we already accept python's dot notation as a string. See the [v.2.2 release notes](https://github.com/waylan/Python-Markdown/blob/master/docs/release-2.2.0.txt#L56) for  an explanation. This is simply a documentation issue. I might have remembered this myself if it was documented properly.

PR #187, opened 2013-02-06
Add footnote content as `title` attribute of footnote link.

Cartroo:
When reading a reaosnable-sized document it's often inconvenient to jump
to the bottom of a document to read the footnote text. While this is
more or less unavoidable for large footnotes, short ones can
conveniently be displayed in a popup next to the footnote itself - this
is most easily achieved by setting the `title` attribute of the link.
This commit changes the `footnotes` extension such that the raw Markdown
source for the footnote content is added in such a capacity, truncated
to a specified width (160 characters by default) and with all whitespace
collapsed into single spaces.


waylan:
Python-Markdown's footnote extension is modeled after [PHP Markdown Extra](http://michelf.ca/projects/php-markdown/extra/#fn-output)'s implementation. We will consider altering our output when the various other footnote implementations do as well. Therefore, to maintain consistency across implementations we will not be implementing this feature at this time.

Of course, anyone is free to create and maintain their own extensions (even as forks of the included extensions) which implement whatever behavior one desires. A list of many such extensions exist on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions).

As an aside, there are browser extensions and/or bookmarklets which you can install to get a similar behavior. I've seen (although don't recall where offhand) at least one bookmarlet that can be included as a script on your site to provide such behavior across all your pages even if your users don't have the bookmarklet. I believe it even preserves lengthy, formatted footnotes. You might want to check Google and see what is available.

PR #191, opened 2013-02-17
Make TocTreeprocessor extendable and be gentle on badly structured headings

benjaoming:
In django-wiki, I want to use an extension of toc, but the current does not allow to be inherited from. This is fixed.

Also, I wrote an external utility function that sorts a list of headers to accommodate this scenario:

```
## Wrongly placed heading
# Top-level heading
```

Both are now included in the TOC, with the first heading as a top-level heading. I know this may upset people enforcing a correct structure of headings, but quite frankly, it is done all the time in the real world of busy users :)

Hope it's useful... thanks for keeping python-markdown up and running!

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -15,36 +15,126 @@ from markdown.extensions.headerid import slugify, unique, itertext
 import re
 
 
+def flatten_list(toc_list):
+    """Given an unsorted list with errors and skips, return a nested one.
+    [{'level': 1}, {'level': 2}]
+    =>
+    [{'level': 1, 'children': [{'level': 2, 'children': []}]}]
+    
+    A wrong list is also converted:
+    [{'level': 2}, {'level': 1}]
+    =>
+    [{'level': 2, 'children': []}, {'level': 1, 'children': []}]
+    """
+    
+    def build_correct(remaining_list, prev_elements=[{'level': 1000}]):
+        
+        if not remaining_list:
+            return [], []
+        
+        current = remaining_list.pop(0)
+        if not 'children' in current.keys():
+            current['children'] = []
+        
+        if not prev_elements:
+            # This happens for instance with [8, 1, 1], ie. when some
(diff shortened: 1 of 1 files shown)


waylan:
@benjaoming can you briefly describe what your changes accomplish. In other words, give me a use case you are trying to address.

Regarding the wrongly ordered headings, can you provide an example of before and after output?

benjaoming:
Sure! Actually, I also remembered wrong... unstructured headings are not being ignored in the old version, they generate invalid lists.

Before:

```
>>> markdown.markdown("##heading2\n\n#heading1\n\n[TOC]", extensions=["toc"])

<h2 id="heading2">heading2</h2>
<h1 id="heading1">heading1</h1>
<div class="toc">
  <ul>
    <li><a href="#heading2">heading2</a></li>
  </ul>
    <li><a href="#heading1">heading1</a></li>
</div>
```

After:

```
>>> markdown.markdown("##heading2\n\n#heading1\n\n[TOC]", extensions=["toc"])

<h2 id="heading2">heading2</h2>
<h1 id="heading1">heading1</h1>
<div class="toc">
  <ul>
    <li><a href="#heading2">heading2</a></li>
    <li><a href="#heading1">heading1</a></li>
  </ul>
</div>
```

waylan:
Ok, that seems like a sensible improvement (invalid -> valid). How about a use case for being extendable? What can you do now that you couldn't before?

benjaoming:
In general, when writing the run() methods for TreeProcessors, try to avoid having presentation logic (ie. creating Element stuff) directly in the run() method.. it is (I think) quite common to want to change this. RIght now, I'm using jQuery to post-manipulate the toc-output... which is a bit ugly :)

There is no way of extending TocExtension and TocTreeprocessor.. I have to recreate the whole run method and even TocExtension because it cannot be configured to use a different TreeProcessor.

This might be the case in other extensions too, but as the TOC is quite sensible to aesthetic desires, I can see it being more important here :)

Issue #52, opened 2011-11-16
Better (and documented) Support for markdown="1"


Mekk:
PHP Markdown allows one to <a href="http://michelf.com/projects/php-markdown/extra/#markdown-attr">use markdown inside HTML blocks</a> - simply by adding markdown=1 attribute to appropriate HTML block. 

It would be nice if python markdown also allowed for such a feature.

I faced the problem after migrating my blog from PHP to Python but seems <a href="http://stackoverflow.com/questions/1705895/python-markdown-markdown-inside-html-blocks">I am not the only one</a>.  It is useful in cases like

``` html
<div class="blahblah" markdown="1">
Some *normal* markdown [text][] here.
</div>
```

or

``` html
<blockquote markdown="1">
Some *markdown text*

<pre name="code" class="python">
# code block which is to stay inside blockquote
</pre>

Yet another *markdown text*
</blockquote>
```

waylan:
> PHP Markdown allows one to <a href="http://michelf.com/projects/php-markdown/extra/#markdown-attr">use markdown inside HTML blocks</a> - simply by adding markdown=1 attribute to appropriate HTML block.

Actually, PHP Markdown **Extra** supports that feature -- as does the [Extra](https://github.com/waylan/Python-Markdown/blob/master/docs/extensions/extra.md) Extension of Python-Markdown. Unfortunately, it is not documented as the feature doesn't work consistently due to limitations of our rudimentary html parser. Of course, patches are always welcome.

I should also note that unlike all other features of the Extra Extension, this feature is built right into the markdown core and is simply turned on with a flag by Extra. See the [source](https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/extra.py#L49) for [details](https://github.com/waylan/Python-Markdown/blob/master/markdown/preprocessors.py#L57).

waylan:
Upon further consideration, I'm reopening this as a "feature" to implement "someday-maybe". Note that I changed the title to: _Better and documented_ Support for markdown="1"

kaba2:
Hi,

I would like to encourage in implementing this feature. I am using Python Markdown in a documentation system which extends the capabilities of Markdown by macros. These macros construct things in Markdown and html, and embed them in html-div blocks, so that they can be styled by CSS. This works in Python Markdown 2.0 (well enough for me), but not in Python Markdown 2.2, so I am unable to upgrade.

Kalle

f0x2501:
Hi,

[...]

PR #199, opened 2013-03-07
Suggested fix for Issue #52 - supporting markdown="1" in nested blocks

christopherjwhite:
Attempt to track the depth level for the tag that was marked as markdown="1" and only perform markdown when the matching tag according to depth is found.

### DIFF
--- a/markdown/preprocessors.py
+++ b/markdown/preprocessors.py
@@ -67,6 +67,7 @@ class HtmlBlockPreprocessor(Preprocessor):
         |                                                         # OR
         \s+(?P<attr2>[^>"'/= ]+)                                  # attr
         """
+    # "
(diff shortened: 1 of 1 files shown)


waylan:
@christopherjwhite looks interesting. Do you have any tests for that?

christopherjwhite:
I have a small test file I was using -- let me see if I can get that integrated into the test framework.

christopherjwhite:
Ok, found the raw-html.txt test and using that.  Found an issue that will take a bit more time, let me poke at it some more...

christopherjwhite:
Going to close this for now.  The entire html block parsing needs more work.  Even without any blocks with markdown="1" I found some problems. 

The following example illustrates the issue:

```
<div>

Nested inside a _plain_ div.

<div>

_foo_

</div>

_bar_

</div>
```

The closing div after _foo_ but before _bar_ is detected as the end of the block that started at the top, which does not take nesting into account.

I have some ideas on how to better clean this up, but will take more testing.

PR #211, opened 2013-04-14
Test for headerid extension

benjaoming:
Thought it'd be nice to have this test, ie. there was a bug recently due to `elem.id` instead of `elem.get(id)` - this should now be caught.

Have a nice day :)

### DIFF
--- a/tests/extensions/headerid.html
+++ b/tests/extensions/headerid.html
@@ -1,0 +1,8 @@
+<h1 id="header-1">Header 1</h1>
+<h2 id="header-2">Header 2</h2>
+<h3 id="header-3">Header 3</h3>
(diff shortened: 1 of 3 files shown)


waylan:
@benjaoming the HeaderId extension is tested in the [unit tests](https://github.com/waylan/Python-Markdown/blob/master/tests/test_extensions.py#L185). What does this add that we're not already testing there? Wouldn't it make more sense to add any new tests there?

benjaoming:
Aha, that explains. Wasn't ware of the unit tests - ran into the problem on 2.1.1, found the fix, and thought it had gone unnoticed due to missing tests, since there was no tests/extensions/headerid.txt present.

PR #226, opened 2013-06-28
Fix issue when second value of captured group is None

nosuchip:
It is possible to create a regexp with named groups which evaluates some groups values to `None` like in following example:

```
import re

rx_pattern = r'([^(]|^)(http|https)://(?:www.|)vimeo\.com/(?P<vimeoid>\d+)(?:\[(?P<width>\d+|),\s*,\s*(?P<height>\d+|)\]|)\S*'
test_str = "http://vimeo.com/45678967234"

rx = re.compile(rx_pattern)
matches = rx.search(test_str)
print matches.groups()   # >>> ('', 'http', '45678967234', None, None)
```

In this case serializator failed when trying to find something inside string which is `None`

### DIFF
--- a/markdown/serializers.py
+++ b/markdown/serializers.py
@@ -150,7 +150,7 @@ def _serialize_html(write, elem, qnames, namespaces, format):
                 _serialize_html(write, e, qnames, None, format)
         else:
             write("<" + tag)
-            items = elem.items()
+            items = [item for item in elem.items() if item[1] != None]
             if items or namespaces:
                 items.sort() # lexical order
                 for k, v in items:


waylan:
Is None being passed to the serializer by any code in the markdown codebase?  Or is this only occurring with some third-party extension? If it is the former, shouldn't we fix the problem at the source? If it is only the latter, then it seems that the solution is to write a better extension which doesn't insert None into ElementTree objects. ElementTree is in the python standard library and has never worked with None values. Why should markdown's (slightly modified) serializer be any different? The only reason we don't use ElementTree's default serializer (we used to) is because it was built for xml and we needed better (valid) html output. Even so, that doesn't change "garbage in, garbage out".

My point is that this patch feels more like a bandaid that addresses the symptom rather than the problem. Why should we add another loop, reducing parsing speed, just in case an extension author doesn't properly test her extension?

I realize that's a lot of questions. But I'll need satisfactory answers to at least some of them before accepting this pull request. Maybe I'm missing something obvious, but is seems to me that a better solution is to document this in the extension API docs.

waylan:
I'm closing this "wontfix" for the reasons stated in my previous comment. An extension author needs to make sure his/her extension never inserts None values into ElementTree. This is easy enough with something like `element.text = somematch or ''`.

Issue #12, opened 2011-05-27
Add SmartyPants extension as part of Python-Markdown


david-a-wheeler:
This is a feature request.  It'd be nice if there was a built-in (batteries included) extension to implement SmartyPants quoting by turning on a simple extension.

I notice that someone is already using SmartyPants with Markdown for Python, though not as an extension:
http://byrneswoder.com/blog/one-secret-to-generating-clean-html-from-text/

waylan:
I'm not completely opposed to this, but what's the benefit of:

```
markdown.markdown(text, extensions=['smartypants'])
```

over:

```
smartypants(markdown.markdown(text))
```

The latter works today and requires less typing. Of course, I realize in some more complex situations (like using smartypants with the codehilite extension) things may not work as well. Therefore, I could see a smartypants extension which implemented all the various features as inline patterns inserted into the parser.

However, this is not high enough on my priority list to devote the time to. Of course, patches and/or merge requests are always welcome.

david-a-wheeler:
That doesn't work so nicely on the command line, or with other extensions (as you noted).

waylan:
Regarding the command line, markdown outputs to stdout, so any decent
commandline script should be able to take that in stdin with a pipe.
If smartypants doesn't do that, I'd consider that an issue for
smartypants not markdown.

Another issue with making smartypants an extension is how to make it
work on raw html. Markdown does nothing to alter raw html, but users
will expect smartypants to run on the raw html. Sure, an extension
could run as a postprocessor extension on the full text, but them we
have the same issues with other extensions as we have now (when
wrapping the markdown output with a call to smartypants). So what's
the point?

I'm not saying its not worth doing, just that it will be more work
than I have time for if it is done right. Of course, we'll accept
patches/merge requests.

## 

---

\X/ /-\ `/ |_ /-\ ||
Waylan Limberg

waylan:
Based on the reasons stated previously, I will not be implementing this. If and when someone else writes the extension, they can do a merge request and I'll reconsider. Until then, I'm closing this.

jonathaneunice:
Here is such an extension: https://bitbucket.org/jeunice/mdx_smartypants 

Anyone can use the code directly. I would be happy to put it in whatever format or license would make it most palatable to the Python-Markdown team to include. Smartypants seems like _such_ a natural fit with Markdown. That it's not part of the core is just odd.

waylan:
[...]

PR #231, opened 2013-07-26
Add Smarty extension and use it for building docs.

mitya57:
This pull request adds Smarty extension, as discussed in #12. Also, it makes use of that extension when building Python-Markdown's own docs.

The diff of the docs before and after using Smarty: https://gist.github.com/mitya57/6086532.

One cosmetic issue is that `headerid` extension creates ids like [`whatwzxhzdk1s-new-in-python-markdown-20`](https://gist.github.com/mitya57/6086532#file-gistfile1-diff-L756), but links work anyway so that shouldn't be an issue (or can you fix the `headerid` extension instead?)

### DIFF
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -1,0 +1,235 @@
+# -*- coding: utf-8 -*-
+# Smarty extension for Python-Markdown
+# Author: 2013, Dmitry Shachnev <mitya57@gmail.com>
+
+# SmartyPants license:
+#
+#   Copyright (c) 2003 John Gruber <http://daringfireball.net/>
+#   All rights reserved.
+#
+#   Redistribution and use in source and binary forms, with or without
+#   modification, are permitted provided that the following conditions are
+#   met:
+#
+#   *  Redistributions of source code must retain the above copyright
+#      notice, this list of conditions and the following disclaimer.
+#
+#   *  Redistributions in binary form must reproduce the above copyright
+#      notice, this list of conditions and the following disclaimer in
+#      the documentation and/or other materials provided with the
+#      distribution.
+#
+#   *  Neither the name "SmartyPants" nor the names of its contributors 
+#      may be used to endorse or promote products derived from this
+#      software without specific prior written permission.
+#
+#   This software is provided by the copyright holders and contributors "as
+#   is" and any express or implied warranties, including, but not limited
+#   to, the implied warranties of merchantability and fitness for a
+#   particular purpose are disclaimed. In no event shall the copyright
+#   owner or contributors be liable for any direct, indirect, incidental,
+#   special, exemplary, or consequential damages (including, but not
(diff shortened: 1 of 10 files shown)


waylan:
@mitya57 the HeaderId problem exists outside of your smarty extension (see issue #232 which I just opened -- thanks for pointing it out), so that's not a concern for you to worry about here.

Things are looking good. Although I think the tests could be a little more complete. For example I don't see any tests for escaping quotes and it seems to me that many various situations which your code accounts for are not actually being tested. Your gist of the docs provides a nice real-world example, but I'd prefer better tests.

Also, thanks for squashing your commits into the three sensible commits attached to this PR. If you could add one more commit with more extensive tests, I'll merge this. Thanks for all your work.

markdown/extensions/smarty.py
```diff
@@ -0,0 +1,235 @@
...
+
+class SmartyExtension(Extension):
+    def __init__(self, configs):
+        self.config = {
+            'smart_quotes': [True, 'Educate quotes'],
+            'smart_dashes': [True, 'Educate dashes'],
+            'smart_ellipses': [True, 'Educate ellipses']
+        }
+        for key, value in configs:
+            if value.lower() in ('true', 1):
```
waylan: I wonder if we should do `if value.lower() in (True, 'true', 't', 1, '1', 'yes', 'y')`. Some people might create an instance in Python and pass it into Markdown. In that case, we definitely need to account for the bool value `True`. And doesn't the command line pass a string? If so, we need to accept the string `'1'`. The values `'t'`, `'yes'` and `'y'` are probably not necessary, but nice to have (especially with command line users - not so much with pythonists).

markdown/extensions/smarty.py
```diff
@@ -0,0 +1,235 @@
...
+                value = True
+            elif value.lower() in ('false', 0):
```
waylan: Same here: `if value.lower() in (False, 'false', 'f', 0, '0', 'no', 'n')`

waylan:
I just realized why you aren't testing escaping of quotes. Its not implemented. That needs to work before I'll accept this.

Fortunately, it should be easy - although I don't believe it's documented. The escapable chars are all in a list on the Markdown instance (see [Line75](https://github.com/waylan/Python-Markdown/blob/master/markdown/__init__.py#L75) of `__init__.py`), which already includes  the '-' and '.'. So just append the quotes to the list in your `extendMarkdown` method (`md.ESCAPED_CHARS.extend(['"', "'"])`). Then update the tests and that should do it.

I've completed a careful review of your code and it looks good. Address the few issues I raised above and I'll merge the PR. Thanks for all the hard work on this @mitya57.

markdown/extensions/smarty.py
mitya57: I've done this in a bit different way: when the value is already `bool` or `int`, we can't call `value.lower()`.

PR #235, opened 2013-07-31
Enhanced allowed schemes for usage of data-URIs.

Dekue:
It is not possible to use data-URIs to render an image as long as safe_mode isn't false.

### DIFF
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -356,7 +356,7 @@ class LinkPattern(Pattern):
             return ''
         
         locless_schemes = ['', 'mailto', 'news']
-        allowed_schemes = locless_schemes + ['http', 'https', 'ftp', 'ftps']
+        allowed_schemes = locless_schemes + ['http', 'https', 'ftp', 'ftps', 'data']
         if scheme not in allowed_schemes:
             # Not a known (allowed) scheme. Not safe.
             return ''


waylan:
Absolutely not happening! safe_mode is a security feature. Most anything can be embedded in a data-URI. The [spec](http://tools.ietf.org/html/rfc2397) itself states in part: 

> Sites which ... disallow the retrieval of certain media types (such as application script languages or types with known security problems) will find it difficult to screen against the inclusion of such types using the "data" URL scheme.

For example, consider the following:

```
data:text/html;base64,PGh0bWw+PGhlYWQ+PHRpdGxlPlRlc3Q8L3RpdGxlPjwvaGVhZD4NCjxib2R5PjxwPjxhIGhyZWY9ImphdmFzY3JpcHQ6YWxlcnQoJ2V2aWwhJykiPmNsaWNrIG1lITwvYT48L2JvZHk+PC9odG1sPg==
```

Copy and paste that into your browser's address bar. Its an entire web page with javascript which executes upon clicking a link (tested in Chrome). While, it harmlessly only calls `alert('evil!')`, the point is clear; evil things could be done. Sure, you could try to filter by MIME-type, but I'd rather not go there.

Data-URI's are not the sort of thing you want to allow untrusted users to be including in documents on your site. If you really want this, then I suggest turning off safe_mode and using a different method (perhaps like [Bleach](https://github.com/jsocol/bleach)) to secure your user contributed content.

phihag:
@waylan Could you outline the problem with filtering by mime type? Anything that starts with `data:image/png;` should be an image, shouldn't it?

Also note that you've skipped the critical part of that quote:

> Sites which **use firewall proxies** to disallow the retrieval of certain
> media types (such as application script languages or types with known
> security problems) will find it difficult to screen against the
>  inclusion of such types using the "data" URL scheme.  However, they
> should be aware of the threat and take whatever precautions are
> considered necessary within their domain.

So what the RFC says is that it's no longer sufficient to only look into the `Content-Type` HTTP headers, and that is only relevant to proxies or "intelligent" firewalls.

waylan:
"safe_mode" is intentionally restrictive. So I can live with it restricting the use of things that others might find useful. Especially given that there are already other ways available to achieve your desired result. In fact, for the last year or so I've been debating internally about whether to rip "safe_mode" out all together and leave that function to other libraries. I think I would be more inclined to do that than add more features to "safe_mode" right now.

Aside from turning "safe_mode" off and using a third party lib (like Bleach liked above), you can also use the [Extension API](http://pythonhosted.org/Markdown/extensions/api.html) and override the existing behavior. That's what the API is there for. Then list your extension on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) and if it becomes wildly popular, I'll reconsider.

Issue #245, opened 2013-08-29
What's your opinion about porting Sphinx html_add_permalinks feature to markdown?


talha131:
Documentation generated via Sphinx shows a permalink to the every heading when you hover on it.
![screen shot 2013-08-29 at 5 08 53 pm](https://f.cloud.github.com/assets/1094724/1049382/e5eedc9c-10a3-11e3-85e4-67843b11c6b7.png)

Sphinx names [this features `html_add_permalinks`](http://sphinx-doc.org/config.html?highlight=permalink#confval-html_add_permalinks).

To the best of my knowledge, [headerid extension](http://pythonhosted.org/Markdown/extensions/header_id.html) does define id attribute of every heading. But it is not visible in the rendered output.

So 

``` text
#header 
```

will appear as

``` html
<h1 id="header">Header</h1>
```

rather than

``` html
<h1 id="header">Header <a href="#header" title="Permalink to header" class="permalink">¶</a></h1>
```

The [only way](http://stackoverflow.com/q/5636191/177116) I could find  to make it work with markdown is

``` markdown
# A header [¶](#a-header) {#a-header}
```

But this is not automatic.

I am curious has there been any attempt to port this feature to markdown? Was it shot down? Or is there an obvious way that I have missed? Basically I want to know the community's opinion about this feature.

Thanks for your time.

waylan:
I seem to recall this discussion taking place back when the [TOC](http://pythonhosted.org/Markdown/extensions/toc.html) extension was added (which imports much some of its functionality from the HeaderId extension). The TOC extension currently has an option (anchorlink) which converts the entire header into a link. IIRC, that behavior was preferred to displaying a pilcrow as the link. I would suggest searching the [list archive](http://sourceforge.net/mailarchive/forum.php?forum_name=python-markdown-discuss) for details.

Regardless of how the link gets added, I suppose it would make sense to break that functionality out into the HeaderId extension like the other functionality shared by both extensions. Pull requests are certainly welcome. :-)

talha131:
Thanks for your response. I sure do want to work on this feature plus "breaking down functionality" that you mentioned.

talha131:
Thanks to @mitya57, this feature has been implemented.

PR #252, opened 2013-09-22
Enable anchorlinks in Python-Markdown documentation

mitya57:
This will enable users to more easily add links to Python-Markdown documentation parts, and also make them know that such an option exists (and prevent issues like #245).

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -13,7 +13,7 @@ from __future__ import absolute_import
 from __future__ import unicode_literals
 from . import Extension
 from ..treeprocessors import Treeprocessor
-from ..util import etree
+from ..util import etree, parseBoolValue
 from .headerid import slugify, unique, itertext, stashedHTML2text
 import re
 
@@ -127,7 +127,7 @@ class TocTreeprocessor(Treeprocessor):
         div.attrib["class"] = "toc"
         header_rgx = re.compile("[Hh][123456]")
         
-        self.use_anchors = self.config["anchorlink"] in [1, '1', True, 'True', 'true']
+        self.use_anchors = parseBoolValue(self.config["anchorlink"])
         
         # Get a list of id attributes
         used_ids = set()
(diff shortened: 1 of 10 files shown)


waylan:
I think the intent is good here, but wouldn't be better if the output worked with Sphinx's stylesheets? I'm more inclined to accept a patch that adds a "permalinks" option to TOC which outputs html that works with Sphinx's stylesheets out of the box. I think we can leave the existing "anchorlinks" option as is and use a new permalinks option in the docs.

A couple suggestions about the implementation: The default should be no permalink (in other words if I'm using the extension now and do nothing - when I upgrade markdown, my output does not change). Probably use the keyword `permalinks=False` as a default to accomplish that. If `permalinks=True` (the boolean value) then default to a pilcrow as the anchor text for the permalink. If `permalinks="some string"` then use that string as the anchor text. This matches Sphinx's behavior - which I think is a sensible approach.

mitya57:
Done (though I named the config option `permalink` (without s) for consistency with existing `anchorlink` option).

markdown/util.py
```diff
@@ -84,6 +84,19 @@ def isBlockLevel(tag):
     # Some ElementTree tags are not strings, so return False.
     return False
 
+def parseBoolValue(value, fail_on_errors=True):
```
waylan, line 87: I've been intending to do this for some time but never got around to it. Great job. Thanks.

markdown/extensions/toc.py
```diff
@@ -89,16 +89,24 @@ def iterparent(self, root):
...
+        anchor.attrib["class"] = "toclink"
+        c.text = ""
+        for elem in c.getchildren():
+            anchor.append(elem)
+            c.remove(elem)
+        c.append(anchor)
+
+    def add_permalink(self, c, elem_id):
+        permalink = etree.Element("a")
+        permalink.text = ("\xb6" if self.use_permalinks is True
```
waylan: Should we perhaps be outputting an html entity (`&#182;` or `&para;`) here? What does Sphinx do?

waylan:
Looks good. Update the docs for TOC and I'll merge this.

markdown/extensions/toc.py
mitya57: Sphinx inserts raw unicode. I've switched it to inserting `&para;` in the last commit, feel free to choose whatever you prefer.

mitya57:
Docs updated now.

PR #258, opened 2013-10-05
markdown="1" | "Default to Span Mode"

ryneeverett:
Implements a 'default span mode' in those blocks with the `markdown="1"` attribute that should never have block elements within them. Nothing here fixes the deficiencies mentioned in issue #52, but it does fall under the category of 'better support for markdown="1"'.

The regex in `contain_span_tags` comes directly from php-markdown:

```
# Tags where markdown="1" default to span mode:
protected $contain_span_tags_re = 'p|h[1-6]|li|dd|dt|td|th|legend|address';
```

Spanmode tests:
Input (with 'extra' enabled):

```
<p markdown="1">
    This is not a code block since Markdown parses paragraph
    content as span.
</p>

<p markdown="1">_Hello_ _world_</p>
```

Output Before:

```
<p>

<pre><code>This is not a code block since Markdown parses paragraph
content as span.
</code></pre>
</p>

<p>

<p><em>Hello</em> <em>world</em></p>
</p>
```

Output After:

```
<p>
    This is not a code block since Markdown parses paragraph
    content as span.
</p>
<p><em>Hello</em> <em>world</em></p>
```

### DIFF
--- a/tests/extensions/extra/test.cfg
+++ b/tests/extensions/extra/test.cfg
@@ -15,3 +15,6 @@ extensions=footnotes
 
 [tables]
 extensions=tables
+
+[raw-html]
+normalize=1
--- a/markdown/extensions/extra.py
+++ b/markdown/extensions/extra.py
@@ -31,6 +31,7 @@ from __future__ import unicode_literals
 from . import Extension
 from ..blockprocessors import BlockProcessor
 from .. import util
+import re
 
 extensions = ['smart_strong',
               'fenced_code',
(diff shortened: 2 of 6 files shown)


tests/extensions/extra/test.cfg
```diff
@@ -15,3 +15,6 @@ extensions=footnotes
 
 [tables]
 extensions=tables
+
+[raw-html]
+normalize=1
```
waylan, line 20: This should not be necessary. "normalize=1" should only be used on tests that we can't edit, like the tests from markdown.pl. If you are having trouble reproducing the correct output by hand, use the test framework's test generator. It will take the txt file as input and write the html file for you. The following commands should do the trick:

``` bash
touch tests/extensions/extra/raw-html.txt
make update-tests
```

That will cycle through all tests and update any in which the last-modified-time is later for the txt file that the html file. If you want to only generate a specific file you can do that as well:

``` bash
make update-tests tests/extensions/extra/raw-html.txt
```

ryneeverett:
Shoot, didn't realize I was pushing to the same branch.

Regardless, this issue can be closed for now. I hadn't intended to tackle the rest of issue 52 when I made the original pull request, but I started working on it and think I'm pretty close now, so might as well wait a day or two so I can work out the remaining bugs and make a fresh, appropriately titled pull request.

PR #266, opened 2013-11-13
miscellaneous issue52 improvements

ryneeverett:
Just cleans up some of my previously merged code.

### DIFF
--- a/docs/extensions/extra.txt
+++ b/docs/extensions/extra.txt
@@ -47,11 +47,9 @@ Markdown Inside HTML Blocks
 
 Unlike the other Extra features, this feature is build into the markdown core and is turned on when `extra` is enabled.
 
-The content of any block-level element can be Markdown-formatted simply by adding a `markdown` attribute to the opening tag. The markdown attribute will be stripped from the output, but all other attributes will be preserved.
+The content of any raw html block element can be Markdown-formatted simply by adding a `markdown` attribute to the opening tag. The markdown attribute will be stripped from the output, but all other attributes will be preserved.
 
 If the markdown value is set to `1` (recommended) or any value other than `span` or `block`, the default behavior will be executed: `p`,`h[1-6]`,`li`,`dd`,`dt`,`td`,`th`,`legend`, and `address` elements skip block parsing while others do not. If the default is overrident by a value of `span`, *block parsing will be skipped* regardless of tag. If the default is overriden by a value of `block`, *block parsing will occur* regardless of tag.
-
-*An opening tag with the markdown attribute must start immediately on a line following a blank line.*
 
 #### Simple Example:
 ```
@@ -70,10 +68,10 @@ This is *true* markdown text.
 ```
 
 ### Nested Markdown Inside HTML BLocks
-Nested elements are more sensitive and must be used cautiously. Violation of the following will lead to unexpected behavior or unhandled exceptions.
-	* Only block mode elements may have further elements nested within them.
-	* The closing tag of inner elements must be followed by a blank line.
-	* More than one level of nesting is not supported (i.e., elements nested within elements nested within elements). This feature is not an alternative to templating. 
+Nested elements are more sensitive and must be used cautiously. To avoid unexpected results:
+	* Only nest elements within block mode elements.
+	* Follow the closing tag of inner elements with a blank line.
+	* Only have one level of nesting.
 
 #### Complex Example:
 ```
--- a/markdown/extensions/extra.py
+++ b/markdown/extensions/extra.py
@@ -75,8 +75,8 @@ class MarkdownInHtmlProcessor(BlockProcessor):
         # Build list of indexes of each nest within the parent element.
         nest_index = []  # a list of tuples: (left index, right index)
         i = self.parser.blockprocessors.tag_counter + 1
-        while len(self.parser.markdown.htmlStash.tag_data) > i and self.\
-                parser.markdown.htmlStash.tag_data[i]['left_index']:
+        is_nest = self.parser.markdown.htmlStash.tag_data[i]['left_index']
+        while len(self.parser.markdown.htmlStash.tag_data) > i and is_nest:
             left_child_index = \
                 self.parser.markdown.htmlStash.tag_data[i]['left_index']
             right_child_index = \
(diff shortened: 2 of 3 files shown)


docs/extensions/extra.txt
```diff
@@ -52,6 +52,7 @@ The content of any block-level element can be Markdown-formatted simply by addin
 If the markdown value is set to `1` (recommended) or any value other than `span` or `block`, the default behavior will be executed: `p`,`h[1-6]`,`li`,`dd`,`dt`,`td`,`th`,`legend`, and `address` elements skip block parsing while others do not. If the default is overrident by a value of `span`, *block parsing will be skipped* regardless of tag. If the default is overriden by a value of `block`, *block parsing will occur* regardless of tag.
 
 *An opening tag with the markdown attribute must start immediately on a line following a blank line.*
```
waylan: Hmm, I missed this before. Is this always the case? Because it is only the case some of the time in PHP Markdown Extra - which this should be emulating. For example, in PHP’s implementation a heading (h1-6) before the raw HTML block does require a blank line, while a paragraph does. I realize that the difference in the implementation has some effect on these edge cases, but I try to go with what feels less buggy. So, for example, if the document author leaves out the blank line after a paragraph, that feels more like a bug- unless, of course, the docs specifically say otherwise. Unfortunately, PHP’s docs make no mention of this requirement.

So my question to you is, did you change the behavior, or has Python-markdown always had this behavior? If this has previously Been the case, then let's leave it for now, but if not, then I'm inclined to consider this a regression and back out the changes.

docs/extensions/extra.txt
```diff
@@ -52,6 +52,7 @@ The content of any block-level element can be Markdown-formatted simply by addin
...
+*The closing tag must follow a blank line if there are multiple block elements nested within the block.*
```
waylan: This is definitely a regression and is not a requirement of Php's implementation. This is a no-go. I haven't checked whether this was implemented I your previous PR or in this one, but it's not staying in.

docs/extensions/extra.txt
ryneeverett: This was a requirement before my initial PR. To me, this seems like an unsurprising limitation of the blockparser, which depends on line breaks for pattern matching. Here's what I get with the official package:

No linebreak before block:

```
import markdown


example = """\
before
<div markdown="1" name="Example">
The text of the `Example` element.
</div>

after
"""

output = markdown.markdown(example, extensions=['extra'])
print ("""-----------------------------------------------
%s
-------------------------------------------------""") % output
```

Result:

```
-----------------------------------------------
<p>before
<div markdown="1" name="Example">
The text of the <code>Example</code> element.
</div></p>
<p>after</p>
-------------------------------------------------
```

Linebreak before block:

```
import markdown


example = """\
before

<div markdown="1" name="Example">
The text of the `Example` element.
</div>

after
"""

output = markdown.markdown(example, extensions=['extra'])
print ("""-----------------------------------------------
%s
-------------------------------------------------""") % output
```

Result:

```
-----------------------------------------------
<p>before</p>
<div name="Example">

<p>The text of the <code>Example</code> element.</p>
</div>

<p>after</p>
-------------------------------------------------
```

docs/extensions/extra.txt
ryneeverett: This was indeed a problem I noticed with my original PR, and I have not gotten around to trying to resolve it.

docs/extensions/extra.txt
waylan: Right, which is the exact same behavior as PHP's implementation. However, if the input is:

```
#before
<div markdown="1" name="Example">
The text of the `Example` element.
</div>

after
```

Then the output should be (because this is what PHP does):

```
<h1>before</h1>

<p><div markdown="1" name="Example">
The text of the <code>Example</code> element.
</div></p>

<p>after</p>
```

But it is actually this:

```
<h1>before</h1>
<p><div markdown="1" name="Example">
The text of the <code>Example</code> element.
</div></p>
<p>after</p>
```

So that is a bug in this implementation. My point is that it should not be a hard and fast rule that a blank line **must** come between a raw block and the preceding block. So the docs shouldn't document it as a rule. Of course, it could be documented as a recommendation. Perhaps something like "To avoid unexpected results, it is generally best to include a blank line before and after any raw html blocks." This doesn't make it a hard and fast rule. It still stands even if the above bug is later fixed. And it is generally good advice anyway.

docs/extensions/extra.txt
ryneeverett: Not a big deal to fix though. Another commit forthcoming.

waylan: Per my comment on the line above, we don't want to document a bug as if it is a feature. I think my suggested change to the docs covers this as well.

docs/extensions/extra.txt
ryneeverett: Ok, I think I understand the distinction you're making: Markdown's rules are completely external to this implementation. Anything else is a bug or recommendation.

waylan: I'm sorry. I've been leading you the wrong way. I forgot that markdown has the following [rule](http://daringfireball.net/projects/markdown/syntax#html):

> The only restrictions are that block-level HTML elements — e.g. `<div>`, `<table>`, `<pre>`, `<p>`, etc. — must be separated from surrounding content by blank lines, and the start and end tags of the block should not be indented with tabs or spaces.

The 'markdown=1" feature is supposed to be a copy of PHP's implementation and I was checking against that. But apparently PHP's implementation goes outside markdown's syntax rules and is a little overzealous in some cases.

So yes, you were initially correct, a blank line before and after is required. However, as that is the case for all raw html blocks, whether 'markdown=1' or not, I'm not sure why we need to document it for the extension specifically. Just stating that 'any block level raw html block as defined in the syntax rules with markdown=1 will work' should be sufficient.

Again, sorry for the confusion. That's what I get for trying to respond from my mobile device.

docs/extensions/extra.txt
waylan: Ok, not that we've cleared up the other item. this is still a bug. If you fix this, I'll merge your commit.

ryneeverett:
Content issues aside, I hope I correctly pushed that commit. I'm still a git noob.

docs/extensions/extra.txt
ryneeverett: I agree.

ryneeverett:
I believe I've addressed all concerns.

ryneeverett:
I finally noticed how massively I screwed up this branch the other night. I fixed it up and removed the offending lines.

PR #267, opened 2013-11-13
tables and attr_list compatibility

ryneeverett:
Treats `<td>` as an inline element even though it's a block element. As far as I know, there's no way to have multi-line cells with the tables extension, and simply using an html `<td>` gets wrapped in an additional `<td>`, so I think there's a case for an exception. [This deficiency has practical consequences.](http://stackoverflow.com/questions/18901540/how-do-i-add-attributes-to-tables-in-python-markdown/19615075#19615075)

Example:

```
import markdown


example = """\
First Header                                           | Second Header
------------------------------------------------------ | -------------
Content Cell{: class="foo bar" title="Some title!" }   | Content Cell
Content Cell                                           | Content Cell{: class="foo bar" title="Some title!" }
"""

output = markdown.markdown(example, extensions=['extra'])
print """-----------------------------------------------
EXAMPLE OUTPUT:
    %s
-------------------------------------------------""" % output
```

Output Before:

```
-----------------------------------------------
EXAMPLE OUTPUT:
    <table>
<thead>
<tr>
<th>First Header</th>
<th>Second Header</th>
</tr>
</thead>
<tbody>
<tr>
<td>Content Cell{: class="foo bar" title="Some title!" }</td>
<td>Content Cell</td>
</tr>
<tr>
<td>Content Cell</td>
<td>Content Cell{: class="foo bar" title="Some title!" }</td>
</tr>
</tbody>
</table>
-------------------------------------------------
```

Output After:

```
-----------------------------------------------
EXAMPLE OUTPUT:
    <table>
<thead>
<tr>
<th>First Header</th>
<th>Second Header</th>
</tr>
</thead>
<tbody>
<tr>
<td class="foo bar" title="Some title!">Content Cell/n</td>
<td>Content Cell</td>
</tr>
<tr>
<td>Content Cell</td>
<td class="foo bar" title="Some title!">Content Cell</td>
</tr>
</tbody>
</table>
-------------------------------------------------
```

### DIFF
--- a/markdown/extensions/attr_list.py
+++ b/markdown/extensions/attr_list.py
@@ -121,6 +121,8 @@ class AttrListTreeprocessor(Treeprocessor):
                 elif elem.text:
                     # no children. Get from text.
                     m = RE.search(elem.text)
+                    if not m and elem.tag == 'td':
+                        m = re.search(self.BASE_RE, elem.text)
                     if m:
                         self.assign_attrs(elem, m.group(1))
                         elem.text = elem.text[:m.start()]
--- a/tests/extensions/extra/tables_and_attr_list.html
+++ b/tests/extensions/extra/tables_and_attr_list.html
@@ -1,0 +1,18 @@
+<table>
+<thead>
+<tr>
+<th>First Header</th>
+<th>Second Header</th>
+</tr>
+</thead>
+<tbody>
+<tr>
+<td class="foo bar" title="Some title!">Content Cell</td>
+<td>Content Cell</td>
+</tr>
+<tr>
+<td>Content Cell</td>
+<td class="foo bar" title="Some title!">Content Cell</td>
+</tr>
+</tbody>
+</table>
--- a/tests/extensions/extra/tables_and_attr_list.txt
+++ b/tests/extensions/extra/tables_and_attr_list.txt
@@ -1,0 +1,4 @@
+First Header                                           | Second Header
+------------------------------------------------------ | -------------
+Content Cell{: class="foo bar" title="Some title!" }   | Content Cell
+Content Cell                                           | Content Cell{: class="foo bar" title="Some title!" }
--- a/tests/extensions/extra/test.cfg
+++ b/tests/extensions/extra/test.cfg
@@ -15,3 +15,6 @@ extensions=footnotes
 
 [tables]
 extensions=tables
+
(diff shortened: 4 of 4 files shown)


waylan:
I guess this seems reasonable. I'm generally not keen on adding improvements to tables (see my explanation in issue #74). But this is actually a tweek on the attr_list extension which would even be required if a different table extension was being used. Although, a test would be good. Add a test and I'll merge this.

Also, apparently I need to be following python markdown related questions on stack overflow. Nobody seems to use the official list any more - the only place I expect support questions to be asked (github is for bugs not support). If only I could follow and respond to stack overflow like an email list. Call me old school, but email list serves are still the best format IMO.

ryneeverett:
Actually, in response that stack question I demonstrated that table extensions which allow multi-line cells _are_ compatible with the current attr_list extension.

The reason I did not add a test is that the existent tests only apply individual extensions, not the entire extra extension. It seemed silly to add a syntax test just for those two lines. It seems like there ought to be an extension-compatibility test which applies all of them, or am I missing it?

I'm annoyed that there doesn't seem to be any way to subscribe to multi-tag combinations on SO (like searching for "[python][markdown]", but your'e an expert of markdown generally so it's probably just as well if you patrol them all. I'm a 90's baby so I find mailing lists intimidating.

me-and:
@waylan, I didn't realise there was a mailing list until now (I've just signed up). Stack Overflow lets folk ask questions without needing to find the correct forum for those questions, and the points system means the answers tend to be a bit more friendly than the brutal treatment I've seen people getting on a few support lists. 

You can't sign up for emails from Stack Overflow AFAIK, but you can add RSS feeds for specific tags to an RSS reader. I'm sure there are services that'll convert RSS to email, too.

waylan:
@ryneeverett I'm still waiting for a test for this. A test is never silly. If you want to add the test for only two extensions you can totally do that. As [documented](http://pythonhosted.org/Markdown/test_suite.html#syntax-test-config-settings), in the [test.cfg](https://github.com/waylan/Python-Markdown/blob/master/tests/extensions/test.cfg#L2) file, the `extensions` option accepts a comma delimited list of extension names. Just be sure to not include any whitespace in the list. So something like:

```
[tables_and_attr_list]
extensions=tables,attr_list
```

PR #276, opened 2014-01-06
allow unicode text in metadata

yoniLavi:
The metadata keys and values should explicitly allow unicode text

### DIFF
--- a/markdown/extensions/meta.py
+++ b/markdown/extensions/meta.py
@@ -46,8 +46,8 @@ from ..preprocessors import Preprocessor
 import re
 
 # Global Vars
-META_RE = re.compile(r'^[ ]{0,3}(?P<key>[A-Za-z0-9_-]+):\s*(?P<value>.*)')
-META_MORE_RE = re.compile(r'^[ ]{4,}(?P<value>.*)')
+META_RE = re.compile(r'^[ ]{0,3}(?P<key>[-\w]+):\s*(?P<value>.*)', re.U)
+META_MORE_RE = re.compile(r'^[ ]{4,}(?P<value>.*)', re.U)
 
 class MetaExtension (Extension):
     """ Meta-Data extension for Python-Markdown. """


waylan:
I agree that the values should allow anything. In fact, doesn't the existing implementation already allow that. I realize your change explicitly ensure it does, but, as I understand it is not nessecary. Were you getting a failure with any specific input?

As for the keys, the existing limits were a conscious decision. I realize they are being used as python dict keys which could theoretically contain anything, but it seemed to me that there is nothing unusual about limiting keys. Especially if and when those keys are being passed into other libraries/CMSs/frameworks. Is there any specific use case were the existing implementation is too restrictive?

yoniLavi:
I apologize for not elaborating.
In my case, one of the first metadata keys contained non-English characters, and this caused that line and all lines below to become part of the html. My motivation was to allow for fully internationalized metadata in our site's markdown blog posts.

According to the [docs](http://docs.python.org/2/library/re.html) (and my experimentation), adding the "\w" metacharacter together with the re.U flag causes the expression to accept everything classified as alphanumeric in unicode. I don't know about any specific libraries that may be offended by my change, but it is my understanding that the future of python is for full unicode support wherever strings are encountered.

waylan:
Regarding keys, consider this example:

``` python
md = markdown.Markdown()
html = md.convert(some_text)
someframework.somemethod(body=html, **md.Meta)
```

If the keys are not valid python identifiers, that will fail miserably. It seems to me that that restriction is the responsibility of the code, not a document author to enforce. Actually, if any change were to be made, we should probably remove support for the dash (`-`) in keys as that is not a valid python identifier.

yoniLavi:
I agree with your point - if allowing the kwargs syntax in all cases is a requirement, than the dash should go. But given that the dash was there since the file's creation, this added freedom has probably been ok with the community.

Anyway, I do not have very much familiarity of the various use cases for the meta extension and would not want to push my needs to the detriment of others. I just felt it's my duty to make this pull request, since the change helped me in my use case.

yoniLavi:
btw, I just saw that python 3 allows for unicode characters in identifiers - http://docs.python.org/3.4/reference/lexical_analysis.html#identifiers ([PEP-3131](http://www.python.org/dev/peps/pep-3131/))

waylan:
Python 3 may allow unicode characters in identifiers, but we still support python 2 so I won't be accepting that change. And as stated above, the change to the values has no actual effect (`re.U` effects use of `\w` but not `.` in the regex). Therefore I'm closing this won'tfix. 

Of course, you are free to do whatever you want in your own extensions. Feel free to fork and hack the extension to meet your own needs (within the limits of the licence, of course). That's why we have the extension API. We can keep the default implementation simple, but you are free to do as you see fit for your specific use case.

PR #278, opened 2014-01-13
Add range support for HL in codehilite

cgabard:
Because a recent PR add feature for emphasizing some lines in a code block, I propose to add range support too : 

``````
```python hl_lines="2 4-6"
def Test():
    # I am highlighted
    print "not me"
    print """I am highlighted
               me too
               and me"""
    # Not me !
```
``````

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -35,16 +35,32 @@ except ImportError:
 def parse_hl_lines(expr):
     """Support our syntax for emphasizing certain lines of code.
 
-    expr should be like '1 2' to emphasize lines 1 and 2 of a code block.
+    expr should be like '1 2' to emphasize lines 1 and 2 of a code block
+    or contains lines ranges like '1 3-5' to emplasize lines 1 and 3 to
+    5 included.
     Returns a list of ints, the line numbers to emphasize.
     """
     if not expr:
         return []
 
-    try:
-        return map(int, expr.split())
-    except ValueError:
-        return []
+    listsHL = []
+    for exp in expr.split():
+        lex = exp.split("-")
+        if len(lex) == 1:
+            try:
+                val = int(lex[0])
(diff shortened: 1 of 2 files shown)


waylan:
This is a reasonable request, but it seems to me that it should be made to the pygments project. If they implement it, then we shouldn't need to do anything - it will just work.

My suggestion would be to try there first. If they reject it, then I'll reconsider.

cgabard:
Thanks for the reply. I will try this but we always need to do something. The `hl_lines` parameter of the html formater must be for now a list of int (each lines that would be highlighted). Adding range support without break compatibility will require to change this parameter type by, for example, "a list of int or other list of length = 2 (stard and end of each range". We always must add specific support in  codehilite according to this modification and parse the input string into the correct parameter type.

By the way, I will need to add support of pygment's linenostart option : The line number for the first line. For now, it is very difficult to make an extension allowing then in both codehilite and fenced_code because we must change the RE in both. I think it would be nice if we can handle this.

cgabard:
Hum, the previous hl_lines parse the string and create a list of integer. But pygment can handle the string directly. I could use this to handle range with string. But my other point is always opened. We can't easily add support of other pygment options or add this with an extension.

waylan:
Just going through and reviewing open PRs. Sorry, it looks like I forgot about this one. 

@cgabard, as per your last comment, if Pygments can accept the raw string, then that is exactly what CodeHilite should pass it. If Pygments permits ranges, then we would also by default (making no changes to our code). If Pygments does not, then we don't either. Plain and simple. No need to document our own behavior. We just point to Pygments' docs and say we accept whatever Pygments accepts for that keyword.

I'm closing this PR. If you want to make the above change (pass in the string unchanged), feel free to open a new PR. If you have other concerns (like general support for all options), please address them in a separate Issue or PR.

PR #279, opened 2014-01-16
Support org-table syntax in tables extension

duelafn:
Emacs org-table uses a `+` instead of a `|` when drawing header dashes. It would be nice if the tables parser would support this syntax.

```
| Column A | Column B | Column C |
|----------+----------+----------|
| Cell A1  | Cell B1  | Cell C1  |
| Cell A2  | Cell B2  | Cell C2  |
| Cell A3  | Cell B3  | Cell C3  |
```

org-tables will always include `|` on left and right side of table and only the internal `|`'s are replaced with `+`.

Currently the tables extension will detect the table but throw out all columns but the first. This patch will cause simple org-tables to be parsed correctly.

### DIFF
--- a/markdown/extensions/tables.py
+++ b/markdown/extensions/tables.py
@@ -19,6 +19,7 @@ from __future__ import unicode_literals
 from . import Extension
 from ..blockprocessors import BlockProcessor
 from ..util import etree
+import re
 
 class TableProcessor(BlockProcessor):
     """ Process Tables. """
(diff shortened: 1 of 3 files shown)


waylan:
Thanks for the suggestion however I will not be merging this pull request. Please see my response to issue #74 for a full explanation. Interestingly, at a glance it would appear this is requesting the same feature addition, albeit with a different label attached to it. My suggestion would be to release it as a third party extension and list it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) for others to find.

PR #285, opened 2014-02-12
Fix some tests failures

mitya57:
Currently, I see some tests failures (at least with Python 3):

```
ERROR: testHighlightLinesWithColon (tests.test_extensions.TestCodeHilite)
ERROR: Test Fenced Code with Highlighted Lines.
MarkdownSyntaxError: TestSyntax: "extensions/extra/tables_and_attr_list"
```

The first two happen because Pygments' `get_list_opt` requires a list, but gets a generator instead. The last one happens because the HTML file has a trailing newline, while it shouldn't.


mitya57:
Thanks for quick merging!

By the way, do you have any plans for the next release?

waylan:
The next release is basically waiting for the release notes to be written up.  I also had a little cleanup, but your PR took care of it. Thank you. I wanted to resolve a few more open issues, but those look like they are going to require a significant amount of work, so I think it makes sense to do a release now to get the existing changes out there for people to use. If anyone wanted to take a crack at the release notes (use previous versions as a model), it will move things along.

mitya57:
I have now [started](https://github.com/mitya57/Python-Markdown/commit/c041da79088f2ba8bd6672a04aeacff0592fae72) updating the release notes. Please let me know if I missed something important.

For some reason smarty extension converts "What's new" to "What&lsquo;s new" (with `lsquo` instead of `rsquo`), I should look at that.

PR #298, opened 2014-03-20
Split convert method in two steps

ghost:
Hi Waylan

I am currently writing a project that will use markdown input files (more precisely, a spritz-like reader, that would be aware of text structure like paragraphs, titles etc.), and I need a markdown parser that would return me structure I can iterate on. I tried to use your implementation to do so. I browsed the documentation about using and extending your module but I failed at finding an entry point that would give me what I need. Most of the documentation cover writing extension plug-ins, which I can't use since I don't intend to output HTML.

Ultimately, I hacked an entry point myself. I've split Markdown.convert method in two steps, parse and serialize, so I can extract the tree from the parse method. I also aliased convert to be serialize(parse(text)) to keep API compatibility.

I submit you this pull request in case you would be interested in my modification. Developers could re-use this entry point to plug Markdown language to other forms of output (like, for example, wiki markup). Please let me know if there was any cleaner workaround, or if you consider this change to be worth including in your project.

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -248,16 +248,13 @@ class Markdown(object):
             e.args = (message,) + e.args[1:]
             raise
         return self
-
-    def convert(self, source):
+    def parse(self, source):
         """
-        Convert markdown to serialized XHTML or HTML.
-
-        Keyword arguments:
+        Keyword argument: 
 
         * source: Source text as a Unicode string.
 
-        Markdown processing takes place in five steps:
+        Markdown parses the document in three steps:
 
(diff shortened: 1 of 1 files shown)


waylan:
I'm not sure why this is nessecary. Why not just swap out the serializer?

```
md.markdown.Markdown()
md.serializer = your_replacement_searializer
```

ghost:
OK, that would work. I need to return a dummy string to post-processors, and ignore `convert()` result. It's hackier on my side of the code, but it's good enough.

Thank you for support and responding so quickly.

Issue #264, opened 2013-11-05
Escaped quotes test failing


waylan:
When running the included PHP tests I'm getting the following failure:

```
======================================================================
MarkdownSyntaxError: TestSyntax: "php/PHP-Specific Bugs"
----------------------------------------------------------------------
MarkdownSyntaxError: Output from "/home/pi/md/tests/php/PHP-Specific Bugs.text" failed to match expected output.

--- /home/pi/md/tests/php/PHP-Specific Bugs.xhtml
+++ actual_output.html
@@ -5,13 +5,13 @@
   Headers below should appear exactly as they are typed (no backslash added or removed).
 </p>
 <h1>
-  Header "quoted\" again \""
+  Header "quoted" again \""
 </h1>
 <h2>
-  Header "quoted\" again \""
+  Header "quoted" again \""
 </h2>
 <h3>
-  Header "quoted\" again \""
+  Header "quoted" again \""
 </h3>
 <p>
   Test with tabs for <code>_Detab</code>:


----------------------------------------------------------------------
Ran 292 tests in 54.644s

FAILED (MarkdownSyntaxError=1, SKIP=24)
```

I haven't had a chance to debug it yet,but didn't want to forget about it.

waylan:
Can't seem to replicate this on another system. As this is for a PHP specific bug anyway, I'm closing this and ignoring it for now.

waylan:
This has reappeared. After inplementing Travis CI, this is failing [there](https://travis-ci.org/waylan/Python-Markdown/jobs/21473956). In a comment on #299 @mitya57 stated:

> Re the failing test, I don't see why “Headers below should appear exactly as they are typed (no backslash added or removed)”. I.e., if I put a single backslash, I want to escape the next character, and if I want a real backslash, I would put a double backslash.
> 
> Also, can you reproduce it locally (and if yes, on what OS)? I can't.

I believed I reproduced this on a RaspberryPI, but not anywhere else. Still, the fact that it is giving different results in some situations is concerning. That said, I haven't really worked out what is being tested here (this is a test from PHP's test suite). It may be appropriate to simply skip this test.

I'm reopening thIs issue until a definitive solution can be reached.

waylan:
Fixed this in 85be15973

PR #299, opened 2014-03-24
Add .travis.yml for continuous integration

mitya57:
I have set up an account for myself, and it works: https://travis-ci.org/mitya57/Python-Markdown. Maybe you'll want to do the same (Travis is really a cool thing, and it can even check all incoming pull requests).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,5 @@
+language: python
+python:
+  - "2.7"
+  - "3.3"
+script: nosetests


waylan:
@mitya57 thanks for the extra push to finally do this. As you can see I implemented it a little different for the reasons stated in my commit message (also had to fix it in 94808ec37). However, I'm glad I did, as there is a [failing test](https://travis-ci.org/waylan/Python-Markdown/builds/21469913) only when running the extra tests that require tidylib (I have no idea what is causing the bug and find it curious that is doesn't exist on Python 3.3).

mitya57:
Nice, I tried to `pip install pytidylib`, but that didn't work without libtydy itself, and I didn't realize you can `apt-get install` stuff.

mitya57:
Re the failing test, I don't see why &ldquo;Headers below should appear exactly as they are typed (no backslash added or removed)&rdquo;. I.e., if I put a single backslash, I want to escape the next character, and if I want a real backslash, I would put a double backslash.

Also, can you reproduce it locally (and if yes, on what OS)? I can't.

waylan:
@mitya57 I have responded to this in the discussion on issue #264. Please take any further discussion there. Thanks.

PR #315, opened 2014-06-01
Add failing unit test for smarty

lkesteloot:
Added this failing test:

```
"Ellipsis within quotes..."
```

The second double quotes should be rdquo, but smarty makes ldquo. I tried to figure out where the bug might be, but I couldn't narrow down the bug because all of these cases succeed:

```
"One period."
"Two periods.."
"Semicolon;"
```

There must be something about the order of regexp evaluation I'm not understanding.


waylan:
> There must be something about the order of regexp evaluation I'm not understanding.

That would be my guess. Remember, Smarty handles the ellipsis also, but not the period or semicolon. Either the order of regexes should be switched, or one of them needs to be a little more sophisticated.

Thanks for the report @lkesteloot. I'll take a closer look when I find the time ... Unless someone else beats me to it.

mitya57:
That was because `endOfWordClass` contains `.` and `:`, but not `…`. I have now changed the patterns order (in my branch), and things are working again.

PR #317, opened 2014-06-08
Python 3.4 and fixes for doctests

mitya57:
This replaces #314.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,7 @@ env:
   - TOXENV=py27
   - TOXENV=py32
   - TOXENV=py33
+  - TOXENV=py34
 before_install:
   - sudo apt-get update -qq
   - sudo apt-get install libtidy-0.99-0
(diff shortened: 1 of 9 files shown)


waylan:
We no longer use doctests as the tests are duplicated with unit tests (see commit  1ff396d). That being the case, I don't see much value in this -- it's just more to maintain. The documentation is nice, but I'd rather that be in the docs.

My inclination is to not commit this, but remove the doctests and perhaps improve the documentation for those extensions. And if any of the doctests are not yet duplicated  in unit tests, then we should do that also.

mitya57:
We (@wRAR and me) mostly need this because Debian's new pybuild system runs `nose --with-doctest` when building the package.

So, removing doctests is fine for us, but if you keep them, I don't see why they can't be converted to new print() syntax.

mitya57:
@waylan I have updated this PR to **not** include the &ldquo;run doctests during nose run&rdquo; part.

PR #319, opened 2014-06-18
smarty: add support for angled quotes

mitya57:
Such quotes are used in French, Russian/Ukrainian/Belorussian, German and many other languages. See http://en.wikipedia.org/wiki/Guillemet for details.

Now smartypants will convert `<<` to `«` and `>>` to `»` if the corresponding option is set (disabled by default).

### DIFF
--- a/tests/extensions/test.cfg
+++ b/tests/extensions/test.cfg
@@ -40,4 +40,4 @@ extensions=nl2br,attr_list
 extensions=admonition
 
 [smarty]
-extensions=smarty
+extensions=smarty(smart_angled_quotes=1)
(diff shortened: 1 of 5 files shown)


tests/extensions/test.cfg
```diff
@@ -40,4 +40,4 @@ extensions=nl2br,attr_list
 extensions=admonition
 
 [smarty]
-extensions=smarty
+extensions=smarty(smart_angle_quotes=1)
```
waylan: Shouldn't that be `smart_angled_quotes` (you're missing the 'd')? I believe that is why the Travis build failed. Fix it and I'll merge. I think this is agreat addition.

mitya57:
@waylan Fixed, sorry.

PR #320, opened 2014-06-26
Add new tab extension.

ghost:
Haven't added documentation, but I just coded something quick to modify the link patterns to open all links in a new tab for people that might want it.

### DIFF
--- a/markdown/extensions/newtab.py
+++ b/markdown/extensions/newtab.py
@@ -1,0 +1,51 @@
+"""
+New Tab Extension for Python-Markdown
+=====================================
+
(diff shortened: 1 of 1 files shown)


waylan:
Interesting, although I'm not sure I see the value in shipping this with Markdown (the value it adds verses the additional maintenance). Seems like a good candidate for a third party extension. I would suggest packaging it up and and listing it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions).

ghost:
Alright, then. That's what I was considering, although I figured that I'd offer a pull request first.

PR #323, opened 2014-07-08
Orgtbl compat

DEADB17:
This branch adds additional syntax to the table extension such that the `+` character can be used as a cross-section character in header separators. This is particularly convenient for emacs org-mode users as it has tools based on this syntax. e.g.:

```
| First Header | Second Header |
|--------------+---------------|
| Content Cell | Content Cell  |
| Content Cell | Content Cell  |
```

This addition is backwards compatible and does not affect existing functionality.

Tests (all passing) and docs updates are included.

### DIFF
--- a/docs/extensions/tables.txt
+++ b/docs/extensions/tables.txt
@@ -49,6 +49,14 @@ will be rendered as:
       </tbody>
     </table>
 
+Alternatively, as a convenience for [org-mode](http://orgmode.org/) users the
+following text will yield the same html:
+
+    | First Header | Second Header |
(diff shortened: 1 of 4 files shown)


waylan:
Thanks for the suggestion however I will not be merging this pull request. Please see my response to issue #74 for a full explanation.  My suggestion would be to release it as a third party extension and list it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) for others to find. For that matter, an existing extension listed on the wiki may already offer support for the syntax you desire.

PR #328, opened 2014-07-27
Fix including MANIFEST.in (fixes tox from sdist)

koobs:
Currently (without MANIFEST.in in the sdist), running tox results in the following error:

```
running build_docs
error: [Errno 2] No such file or directory: 'docs/_template.html'
```

Fetching the raw MANIFEST.in from the github repository correctly allows the relevant files (recursive-include docs *, in this case) to be included in the virtualenv that is created by tox.

I'm assuming that the addition of "MANIFEST" in MANIFEST.in" was a typo, and that MANIFEST.in was intended instead. If this is not the case, feel free to reject this PR and resolve in a different manner

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,4 +10,4 @@ include makefile
 include LICENSE.md
 include README.md
 include INSTALL.md
-include MANIFEST
+include MANIFEST.in


waylan:
Things are the way they are intentionally. When you run setup.py it looks at the MANIFEST.in file and builds the MANIFEST file - which it then includes in the build.  If new files have been added since the last build, you want those files included in the MANIFEST file. Therefore only the MANIFEST.in file should be in source control (but not MANIFEST) - this is basic python stuff.

If you are using an existing build (from PyPI) then the MANIFEST file already exists in that build (and as no changes are expected for a given build) there is no reason to have a MANIFEST.in file to rebuild the MANIFEST file.

Given the above, I don't understand what problem you are trying to solve. I suspect perhaps you have a bad tox config but I can't say without more info.

waylan:
Ok, I think I see the problem. I just tested this and apparently when the MANIFEST file is generated (from a MANIFEST.in file) on a run of "setup.py sdist" is it not included in the build. This seems like a bug in Python to me, but in any event it is easy enough to fix with a few quick additions to the makefile.

koobs:
Thanks @waylan :)

PR #329, opened 2014-07-29
Do not fail on None value

noirbizarre:
This PR handle `None` value as an empty string avoiding failure with the following code:

``` python
import markdown

source = None
markdown.markdown(source)
```

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -272,7 +272,7 @@ class Markdown(object):
         """
 
         # Fixup the source text
-        if not source.strip():
+        if not source or not source.strip():
             return ''  # a blank unicode string
 
         try:
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -27,6 +27,10 @@ class TestMarkdownBasics(unittest.TestCase):
         """ Test blank input. """
         self.assertEqual(self.md.convert(''), '')
 
+    def testNoneInput(self):
+        """ Test None input. """
+        self.assertEqual(self.md.convert(None), '')
+
     def testWhitespaceOnly(self):
         """ Test input of only whitespace. """
         self.assertEqual(self.md.convert(' '), '')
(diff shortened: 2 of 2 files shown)


waylan:
Two things:

1) If we are to implement this, then there should be a test for it. Probably [here](https://github.com/waylan/Python-Markdown/blob/master/tests/test_apis.py#L19).

2) In the many years I've been involved with this project I've not seen a request for this before now. So why do we need it? The docs clearly state that only a Unicode string is accepted. While I can appreciate the minor shortcut this may provide developers, I'm concerned that it could become a way to hide user errors.

Therefore, before I accept this, can you give me a reasonable use case where this is genuinely useful? For example, is there some existing API's In the Python Standard Library or in popular frameworks/ORMs where a 'string or None' is returned?

Even then, as worst is would only be one more line of code to type. At best a few more characters on the same line. What's wrong with this:

``` python
# Adding only one more line:
src = mightreturnnone() or ''
html = markdown.markdown(src)

# or adding six characters to an existing line
html = markdown.markdown(mightreturnnone() or '')
```

and how is anything that's wrong with that not also a problem with your change? Both would hide the same bugs right? As least in my example the bug is clearly in the users code. They can easily refactor and catch it. With your patch, that becomes more difficult doesn't it?

I guess the point I'm trying to make it that perhaps the user forgets that `mightreturnnone` might actually return None and thinks that markdown is the problem when actually the problem exists before it every gets to markdown. This sounds like a support headache to me. So what benefits do we get that justifies the extra support -- especially when it seems to not be a highly requested feature?

noirbizarre:
Hi!

Two response :)

1) There is a test provided in the second commit.

2) I see your point and it was a proposal. It's up to you to accept or refuse this change.
To be honest, I use it with flask-markdown as a Jinja filter where some value can be `None` and as you said it's only a shortcut to avoid writting `{{ obj.attribute or ''|markdon }}` everywhere.

By the way to answer your sample bug,  `mightreturnnone` should be tested separately. :)
In my case, `obj.attribute` can be `None`, it's the expected behavior.

I thought it was a good idea to not fail on `None` value but maybe it's more logical to you that I provide this patch to flask-markdown instead so only the filter can accept `None` value as input.
I'll do it.

PR #332, opened 2014-08-09
Added options to the Smarty extension that configure the text that is us...

ghost:
...ed to replace quotes. This makes it possible to use the correct quotes in languages other than English.

Use e.g. the following config for correct German quotes:

```
extensionConfigs = {
    'smarty': [('smart_lsquo', '&sbquo;'), # sb is not a typo!
               ('smart_rsquo', '&lsquo;'),
               ('smart_ldquo', '&bdquo;'),
               ('smart_rdquo', '&ldquo;')]
}
```


mitya57:
Please:
- Use different names for configuration variables. They should not follow the same naming scheme as the existing ones, as it would lead to confusion. I suggest using `[lr][sd]quo_replacement` names.
- Use consistent indentation in configuration variables description.
- Add some tests.

waylan:
I think @mitya57 has a point. The `smart_*` config keys are enabling/disabling features, whereas yours are changing the behavior of a feature. Using a different naming scheme might make sense ... Or, perhaps a dict could be passed in to one key. Something like `chars={'lsqou': '&sbquo;'}`. I'm unsure which would be better.

Also this definitely needs tests before consideration. Perhaps the example you added to the docs could be converted into a test. And while you're at it, if you could keep the docs hard wrapped at 80 chars, that would be preferred.

ghost:
Ok, I replaced the smart_lsquo etc. keys by a single one: smart_substitutions and I updated the docs accordingly. However, I don't understand how I can specify a dict as configuration option in the test.cfg file, so currently I cannot write a test.

waylan:
You're correct that INI test config files cannot contain python dicts. There are a few other extensions which need more complex tests that the INI file can provide and their tests are contained in [`tests/test_extensions.py`](https://github.com/waylan/Python-Markdown/blob/master/tests/test_extensions.py). You should include some tests there.

ghost:
tests/test_extensions.py now has a test case for the new feature.

waylan:
@mitya57 I just gave you commit access. Smarty was your work. I'll let you merge this, and/or clean it up as you see fit.

mitya57:
Thanks, I will get to this in a couple of days.

And, actually, I still prefer that you do `s/smart_substitutions/substitutions/` and `s/Overwrite/overwrite/`.

PR #339, opened 2014-09-04
dont use homegrown OrderedDict, use pythons standard

sinistersnare:
I was looking through the source for this library and I saw that you guys use Django's OrderedDict implementation which is now deprecated for pythons standard `collections.OrderedDict`. I took some time to changed all instances of `markdown.odict.OrderedDict` to the proper `collections.OrderedDict`.
1. Is this a good idea?
2. should I remove usage of OrderedDict from `api.txt` also?


sinistersnare:
:/ nevermind i guess, you cant arbitrarily add elements to an OrderedDict, I forgot to run the tests before i pushed this up, and as I did i realized this... sorry :)

waylan:
Yeah, I looked into this a while back and was disappointed we can't arbitrarily add elements to an OrderedDict. Until/unless that changes we will continue to use what we have.

Issue #253, opened 2013-09-24
***italic bold** italic*


tobiasberling:
I think the output for `***italic bold** italic*` should be `<p><strong><em>italic bold</strong> italic</em></p>` (generated by Dingus).
The output I get from Python-Markdown is `<p><strong>*italic bold</strong> italic*</p>`. 
I checked this for version 2.2 and the latest master commit (e92d7cf65e)

waylan:
Well, our output is certainly better than [some](http://johnmacfarlane.net/babelmark2/?normalize=1&text=***italic+bold**+italic*). This is a common problem for markdown parsers. The way the Python-Markdown parser works we either get something that works with your sample, or this: `***italic bold* bold**` but not both. It comes down to what order the various pieces of the parser are run in. At the time, very few if any implementations got it right (but they did get the [alternative](http://johnmacfarlane.net/babelmark2/?normalize=1&text=***italic+bold*+bold**) right) so we followed suite. I see that about half get I correct now. Guess we should to. Thanks for the report.

waylan:
So I reread the original post and to clear any confusion up, @tobiasberling's suggested solution (`<p><strong><em>italic bold</strong> italic</em></p>` is **_not**_ the correct solution. That is a known bug in markdown.pl. Look at the nesting of the 'strong' and 'em' tags. That is not valid html by any standard. The correct output should be: 

``` html
<p><em><strong>italic bold</strong> italic</em></p>
```

Oh and I forget to mention in my previous comment, an easy workaround is to mix '*' and '_' in your input:

```
_**italic bold** italic_
```

There is absolutely no ambiguity there. The parser will always do what the author intended ([See what I mean](http://johnmacfarlane.net/babelmark2/?normalize=1&text=_**italic+bold**+italic_)). Well, that is unless the author wants [invalid nesting](http://johnmacfarlane.net/babelmark2/?normalize=1&text=_**italic+bold_+italic**). At least the output of most implementations in that instance inform the author that s/he did something wrong.

JordanReiter:
In fairness if you look at the results page, one set of parsers does get it right (peg-markdown 0.4.14 et al). 

In fact, this set of parsers gets it right [_both ways_](http://johnmacfarlane.net/babelmark2/?text=***italic+bold**+italic*%0A%0A**bold+*italic+bold***)!

[...]

PR #342, opened 2014-09-20
Fix InlineProcessor and add better italic and bold support

facelessuser:
### Changes
- Fixes issues with tails in InlineProcessor
- Adds better italic and bold support
- New tests for changes and improved code coverage
### Tests

I cannot install Python 3.1 on my OSX Mavericks

```
  py27: commands succeeded
ERROR:   py31: InterpreterNotFound: python3.1
  py32: commands succeeded
  py33: commands succeeded
  py34: commands succeeded
```

```
Name                               Stmts   Miss  Cover   Missing
----------------------------------------------------------------
markdown/__init__                    193     42    78%   141, 330-333, 391-434, 479-493
markdown/__main__                     40      0   100%
markdown/blockparser                  30      0   100%
markdown/blockprocessors             273      7    97%   189, 194-195, 204, 253, 546, 552
markdown/extensions/__init__          34      4    88%   28-29, 36-37
markdown/extensions/abbr              38      0   100%
markdown/extensions/admonition        45      0   100%
markdown/extensions/attr_list         96      0   100%
markdown/extensions/codehilite        99     19    81%   25-27, 43-44, 105-120, 177-178, 181
markdown/extensions/def_list          59      2    97%   95-96
markdown/extensions/extra             54      0   100%
markdown/extensions/fenced_code       48      0   100%
markdown/extensions/footnotes        176      8    95%   91-92, 105, 111, 118, 243, 288-289
markdown/extensions/headerid          84      4    95%   72-73, 75, 103
markdown/extensions/meta              35      0   100%
markdown/extensions/nl2br             11      0   100%
markdown/extensions/sane_lists        17      0   100%
markdown/extensions/smart_strong      14      0   100%
markdown/extensions/smarty            87      0   100%
markdown/extensions/tables            56      0   100%
markdown/extensions/toc              134     18    87%   50-52, 96-104, 118-120, 151, 182, 191
markdown/extensions/wikilinks         49      0   100%
markdown/inlinepatterns              247      1    99%   225
markdown/odict                       113     37    67%   25-32, 35, 42, 54, 57, 60-66, 69-71, 104-105, 108-110, 119-122, 129, 136, 139-140, 160, 185-189
markdown/postprocessors               49      0   100%
markdown/preprocessors               207     17    92%   87, 92, 116, 135, 171, 199, 273, 292-304
markdown/serializers                 153     48    69%   82-83, 106-117, 143, 147-150, 158, 160, 165, 169-174, 181, 203, 218, 224-235, 238, 254, 259, 262, 266, 269
markdown/treeprocessors              187      4    98%   80, 203-205
markdown/util                         59      0   100%
----------------------------------------------------------------
TOTAL                               2687    211    92%
```

The only modification that had to be made to existing tests were for these two issues (which I view as improvements):

``` diff
--- /Users/facelessuser/Desktop/Python-Markdown/tests/misc/para-with-hr.html
+++ actual_output.html
@@ -2,5 +2,5 @@
 <hr />
 <p>Followed by another paragraph.</p>
 <p>Here is another paragraph, followed by:
-*** not an HR.
+<em>*</em> not an HR.
 Followed by more of the same paragraph.</p>
```

``` diff
--- /Users/facelessuser/Desktop/Python-Markdown/tests/misc/em_strong.html
+++ actual_output.html
@@ -4,7 +4,7 @@
 <p>With spaces: * *</p>
 <p>Two underscores __</p>
 <p>with spaces: _ _</p>
-<p>three asterisks: ***</p>
+<p>three asterisks: <em>*</em></p>
 <p>with spaces: * * *</p>
 <p>three underscores: ___</p>
 <p>with spaces: _ _ _</p>
```

Let me know what you think.

### DIFF
--- a/markdown/extensions/smart_strong.py
+++ b/markdown/extensions/smart_strong.py
@@ -20,16 +20,18 @@ from __future__ import unicode_literals
 from . import Extension
 from ..inlinepatterns import SimpleTagPattern
 
-SMART_STRONG_RE = r'(?<!\w)(_{2})(?!_)(.+?)(?<!_)\2(?!\w)'
-STRONG_RE = r'(\*{2})(.+?)\2'
+SMART_STRONG_RE = r'(?<!\w)(_{2})(?![_\s])(.+?_*?)(?<!\s)\2(?!\w)'
+SMART_STRONG_EM_RE = r'(?<!\w)(_{3})(?![_\s])(.+?_*?)(?<!\s)\2(?!\w)'
 
 class SmartEmphasisExtension(Extension):
     """ Add smart_emphasis extension to Markdown class."""
 
     def extendMarkdown(self, md, md_globals):
         """ Modify inline patterns. """
-        md.inlinePatterns['strong'] = SimpleTagPattern(STRONG_RE, 'strong')
-        md.inlinePatterns.add('strong2', SimpleTagPattern(SMART_STRONG_RE, 'strong'), '>emphasis2')
+        md.inlinePatterns['strong2'] = SimpleTagPattern(SMART_STRONG_RE, 'strong')
+        md.inlinePatterns['strong_em2'] = SimpleTagPattern(SMART_STRONG_EM_RE, 'strong')
+        del md.inlinePatterns['strong_em4']
+        del md.inlinePatterns['em_strong2']
 
 def makeExtension(*args, **kwargs):
     return SmartEmphasisExtension(*args, **kwargs)
--- a/markdown/extensions/smart_strong.py
+++ b/markdown/extensions/smart_strong.py
@@ -26,15 +26,15 @@ SMART_STRONG_2_RE = r'(?<!\w)(_{2})(?![\s_])%s(?<!\s)\2(?!\w)' % SMART_CONTENT
 # Smart rules for when "smart emphasis" is enabled
 # ___strong,em_strong__
 SMART_STRONG_EM_4_RE = \
-r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
+r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_(?!\w)%s(?<!\s)_{2}(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
 # ___em,strong__em_
 SMART_EM_STRONG_2_RE = \
-r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_{2}(?![\s_])%s(?<!\s)_(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
+r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_{2}(?!\w)%s(?<!\s)_(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
 
 # Smart rules for when "smart emphasis" is disabled
 # ___strong,em_strong__
 STRONG_EM_4_RE = \
-r'(?<!\w)(_{3})(?!\s)%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (UNDER_CONTENT, SMART_CONTENT)
+r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (UNDER_CONTENT, SMART_CONTENT)
 # ___em,strong__em_
 EM_STRONG_2_RE = \
 r'(?<!\w)(_{3})(?![\s_])%s(?<!\s)_{2}(?!\w)%s(?<!\s)_' % (SMART_CONTENT, UNDER_CONTENT)
--- a/markdown/extensions/smart_strong.py
+++ b/markdown/extensions/smart_strong.py
@@ -4,34 +4,57 @@ Smart_Strong Extension for Python-Markdown
 
 This extention adds smarter handling of double underscores within words.
 
-See <https://pythonhosted.org/Markdown/extensions/smart_strong.html> 
+See <https://pythonhosted.org/Markdown/extensions/smart_strong.html>
 for documentation.
 
 Original code Copyright 2011 [Waylan Limberg](http://achinghead.com)
 
 All changes Copyright 2011-2014 The Python Markdown Project
 
-License: [BSD](http://www.opensource.org/licenses/bsd-license.php) 
+License: [BSD](http://www.opensource.org/licenses/bsd-license.php)
 
 '''
 
 from __future__ import absolute_import
 from __future__ import unicode_literals
 from . import Extension
-from ..inlinepatterns import SimpleTagPattern
+from ..inlinepatterns import SimpleTagPattern, DoubleTagPattern
+from ..inlinepatterns import UNDER_CONTENT, SMART_CONTENT, SMART_STRONG_EM_RE
 
-SMART_STRONG_RE = r'(?<!\w)(_{2})(?![_\s])(.+?_*?)(?<!\s)\2(?!\w)'
-SMART_STRONG_EM_RE = r'(?<!\w)(_{3})(?![_\s])(.+?_*?)(?<!\s)\2(?!\w)'
+SMART_STRONG_2_RE = r'(?<!\w)(_{2})(?![\s_])%s(?<!\s)\2(?!\w)' % SMART_CONTENT
+
+# Smart rules for when "smart emphasis" is enabled
+# ___strong,em_strong__
+SMART_STRONG_EM_4_RE = \
+r'(?<!\w)(_{3})(?!\s)(?![\s_])%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
+# ___em,strong__em_
+SMART_EM_STRONG_2_RE = \
+r'(?<!\w)(_{3})(?!\s)(?![\s_])%s(?<!\s)_{2}(?![\s_])%s(?<!\s)_(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
+
+# Smart rules for when "smart emphasis" is disabled
+# ___strong,em_strong__
+STRONG_EM_4_RE = \
+r'(?<!\w)(_{3})(?!\s)%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (UNDER_CONTENT, SMART_CONTENT)
+# ___em,strong__em_
+EM_STRONG_2_RE = \
+r'(?<!\w)(_{3})(?!\s)(?![\s_])%s(?<!\s)_{2}(?!\w)%s(?<!\s)_' % (SMART_CONTENT, UNDER_CONTENT)
+
 
 class SmartEmphasisExtension(Extension):
     """ Add smart_emphasis extension to Markdown class."""
 
     def extendMarkdown(self, md, md_globals):
         """ Modify inline patterns. """
-        md.inlinePatterns['strong2'] = SimpleTagPattern(SMART_STRONG_RE, 'strong')
-        md.inlinePatterns['strong_em2'] = SimpleTagPattern(SMART_STRONG_EM_RE, 'strong')
-        del md.inlinePatterns['strong_em4']
-        del md.inlinePatterns['em_strong2']
+
+        md.inlinePatterns['strong2'] = SimpleTagPattern(SMART_STRONG_2_RE, 'strong')
+        md.inlinePatterns['strong_em2'] = DoubleTagPattern(SMART_STRONG_EM_RE, 'strong,em')
+
+        if not md.smart_emphasis:
+            md.inlinePatterns['em_strong2'] = DoubleTagPattern(EM_STRONG_2_RE, 'em,strong')
+            md.inlinePatterns['strong_em4'] = DoubleTagPattern(STRONG_EM_4_RE, 'strong,em')
+        else:
+            md.inlinePatterns['em_strong2'] = DoubleTagPattern(SMART_EM_STRONG_2_RE, 'em,strong')
+            md.inlinePatterns['strong_em4'] = DoubleTagPattern(SMART_STRONG_EM_4_RE, 'strong,em')
 
 def makeExtension(*args, **kwargs):
     return SmartEmphasisExtension(*args, **kwargs)
--- a/markdown/extensions/smart_strong.py
+++ b/markdown/extensions/smart_strong.py
@@ -26,10 +26,10 @@ SMART_STRONG_2_RE = r'(?<!\w)(_{2})(?![\s_])%s(?<!\s)\2(?!\w)' % SMART_CONTENT
 # Smart rules for when "smart emphasis" is enabled
 # ___strong,em_strong__
 SMART_STRONG_EM_4_RE = \
-r'(?<!\w)(_{3})(?!\s)(?![\s_])%s(?<!\s)_%s(?<!\s)_{2}(?!\w)' % (SMART_CONTENT, SMART_CONTENT)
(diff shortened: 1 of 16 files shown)


waylan:
This looks pretty good. Two concerns though:

1) is it really nessecary to add so many more regexes? Seems to me like modifying one or two of the existing regexes would have been enough. But then it appears you tested things pretty thoroughly.

2) I'm unsure about the one changed test. I think  the previous behavior was more correct. So do most implementations: http://johnmacfarlane.net/babelmark2/?normalize=1&text=Foo+*** That said, your change does follow markdown.pl.  And an author can always escape one of the asterisks if she really wants three literal asterisks: http://johnmacfarlane.net/babelmark2/?normalize=1&text=Foo+%5C*** But existing docs may render different now and that may be an unwelcome change.

facelessuser:
I will start by saying, ultimately the decision is yours, but I will at least layout my thinking as to why approached it this way. 

### Why So Many Regex?

Yes some can probably be combined, and maybe ultimately will.  I didn't combine some so that things like `smart_strong` can more easily be implemented etc.  Maybe someone would like a `smart_strong_em` for `*` like github does.

Originally I was approaching the issue of improving bold and italic.  Depending on how a scenario is interpreted skews what people thing the result should be (markdown is really a very loose standard open to interpretation).  Take this for example `***Test** me*`:

Dingus says `<strong><em>Test</strong> me</me>`  I kind of think that is silly.  Gruber seems okay with it though.

Python Markdown just evaluates it from left to right. `<strong>*Test</strong> me*`.  Perfectly valid interpretation to me, but I think it misses the spirit of what is being attempted (this is up to interpretation though).

My approach was to mimic more of what the procedural parsers do and trigger off the starting triple token and evaluate the content in a more intuitive way like Kramdown, RDiscount, PHP Markdown etc. http://johnmacfarlane.net/babelmark2/?normalize=1&text=***italic+bold**+italic*.  This would give us `<em><strong>Test</strong> me</em>`.  Am I or they any more right?  Maybe not.  (note: In writing this I realized I reversed the logic of my implementation (`***Test** me**` and `***Test* me**`) which flip flops the strong and em which matches dingus in rendering but with valid html code, but I think that is wrong. so if we went this route, I would need to fix this)  To me this makes more sense...but that doesn't mean it is right.  Anyways, to approach these very different scnarios: `***Test me***`, `***Test** me**` and `***Test* me**`, I had to handle them separately.

I kind approached triple tokens with the same thinking.  So when dealing with `***smart*****triple***` rules,  as long as the tokens don't terminate the triple symbols in the inline block, it should continue.  Again, feel free to strike this down.  I was modeling behavior from what I have seen => github (_**smart_____triple**_) and PHP Markdown http://johnmacfarlane.net/babelmark2/?text=***smart*****triple***%0A.  Again, this doesn't mean I'm right.

To account for these very _different_ scenarios, one regex could not be used.  But if this is a direction you don't want to go, then I can pull it out.

### What about `***`?

Now I spent the first point talking about the spirit of what is intended, so if you don't like that, I would at least think you would be okay with this :smile:.

When I constructed this behavior I was just following the rules.  A starting token cannot be followed by a space, it must contain at least one character in between the block, and it must close with the same token.  The ending token cannot be proceeded by a space.  Currently Python Markdown is inconsistent with how it approaches `___` and `***` see below:

```
EMPHASIS_RE = r'(\*)([^\*]+)\2'                    # *emphasis*
EMPHASIS_2_RE = r'(_)(.+?)\2'                 # _emphasis_ 
```

Which way should it be?  I kind of think we should pick one and stick with it.  As you also see, Python Markdown doesn't care if the start token is followed by a space or not.  Nor does it care if the closing token is preceded by a space, but I think this is wrong.  So do a number of other parsers http://johnmacfarlane.net/babelmark2/?text=_+test+_%0A.

I also think smart rules should gobble up similar tokens until it properly terminates like github does: `___Test______` => **_Test___**_.  Python Markdown currently terminates after the first ending three.

### Conclusion

If I had one change I would want to fight for, it would be this one:  I think we should have consistent rules, and starting tokens should not be followed by whitespace and ending tokens should not be preceded by white spaces either.  Yes this changes the behavior of `***`, but I think it makes sense.

In addition, I am strongly found of the requiring that smart rules continue to gobble up tokens until it properly terminates with a non word character like github does: `___Test______` => **_Test___**_.

All the other stuff I am less adamant on.  But regardless of how adamant I am about any of this, it is up to you.

I understand the concern about changing the document output.  The issue I was trying to address was improving bold and italic.  By improving the consistency or relevance of the results it will be very hard to not change the output.  The nature of what is being done here _is_ to change the output.  Especially when it comes to rules not being consistent, I feel it can't be helped to change one rule one way or the other rule the other way.

Sorry to go on and on.  But at least you can understand my thinking and make a decision based on that.

So if you keep everything, I would need to fix the flip flopped logic I mentioned earlier.  And if you strike down some of my suggestions, I will have to revert those parts.  So either way I have more stuff to do :smile: .  I will await your response before proceeding further.

waylan:
Alright, you've convinced me. Fix your flip flopped logic and I think this will be good to go.

facelessuser:
Ha!  That went over much easier than I expected :smile: .  Will do.  Thanks for hearing me out!

facelessuser:
I am going to spend a little more time and write some more complex tests for smart scenarios.  I want to make sure smart rules for underscores make sense in all scenarios since they can be mixed with dumb strong or smart strong.  I need to cover this thoroughly.  I want to make sure I've dotted my 'i's and crossed my 't's.  Though generally they look good in basic test that are here, I don't think I have yet spent enough time proving out that part.  I suspect I may find some issues.

I will let you know when I am done with that as well.

facelessuser:
After diving into the smart rules deeper and characterizing the behavior, I think I almost have the dumb and smart underscore playing nice together in a more expected way during more complex rules.  I also improved some of the triple behavior (I will go into this when I wrap this up in the next day or so).  What's more, all the other tests remain unchanged and passing.  I think after I get this done, I will be completely done.

facelessuser:
Okay, once I finish verifying the test data by hand (hopefully tonight).  I should be done and will commit.  I am glad I took the time to really test all of this good.  I really wanted to make sure this was tested as good as I possibly can to catch any quirks.

The new changes will simply be:
- triple token rules will properly terminate in scenarios where they should:
  - smart: should terminate when a nested token is followed by a non-word char like github does `___Test _Test___` => **_Test _Test**_  vs `___Test_ Test___` => **_Test_ Test_** (again this is just making triple chars act like they are truly nested em strongs).
  - non-smart: should properly deffer to nested rules when italic and bold are nested within
- Scenarios such as `___Test_Test__` and `___Test__Test_` will have the appropriate nested section use smart or non-smart rules and boundaries depending on when `smart_emphasis` or `smart_strong` are enabled.
- Tests for underscore emphasis and strong when: only smart_emphasis is enabled, only smart_strong is enabled, when both are enabled, or when both are disabled.  They should now give outputs that make sense in all these scenarios.

There will be no additional rules added for this than I already have in the first commits, All I am doing is swapping out rules depending on the scenario.  There is no impact to any of the built in tests as these scenarios only pop up in the 4 syntax tests I have added.

I am sure there will always be people comparing outputs to various  other parsers, so I am sure there is no way to please everyone.  But based on the internal architecture of this parser, the output should hopefully be very consistent and sane.  Triple tokens should feel like they are truly nested bold and italic rules.

facelessuser:
@waylan, I had a chance to look at my results over lunch today, and I want to ask your opinion.

In the test file https://github.com/waylan/Python-Markdown/blob/master/tests/misc/underscores.txt, there is the following test:

 `THIS___SHOULD___STAY?` => `<p>THIS<strong><em>SHOULD</em></strong>STAY?</p>`

When this test was written, I am not sure if this was actually meant as a question or was just meant to be some random text, but this is the only thing that disagrees with my final set of changes.

When this test is run, `smart_emphasis` == `True` and `smart_strong` is **not** enabled.

Since this set is a smart em inside a strong, I am requiring the triple block to respect smart boundaries meaning this would **not** be  `<p>THIS<strong><em>SHOULD</em></strong>STAY?</p>` but instead would be  `<p>THIS_<strong>SHOULD</strong>_STAY?</p>` since strong is not a smart rule and em is and would require non-word characters on either side.  If smart_enabled were turned off, this would be a different story.

When **only** `smart_emphasis` is enabled, I am currently treating boundaries on triple blocks like so (`^` will represent a smart boundary):

``` html
^<strong><em>Test</em></strong>^

^<strong><em>Test</em>^ Test</strong>

^<em><strong>Test</strong> Test</em>^
```

What is your opinion on this test?  If you are okay with this change, I will commit tonight.  If not, I will need to consider refactoring some how and determine another way to qualify the mixing of smart and non-smart rules.

facelessuser:
I went ahead and just committed it so that way you can evaluate all of it.

facelessuser:
Sorry, I am a bit of a perfectionist.  I have been over these changes multiple times now, and the rules should now be consistent.  I am going to stop touching this.

facelessuser:
In regards to the underscore rule this breaks.  I was thinking I could revise this to honor smart boundaries only when the outside tag is a smart tag.  For instance:

So when `smart_emphasis` == `True` and `smart_strong` == `False`:

``` html
<strong><em>Test</em></strong>

<strong><em>Test</em>^ Test</strong>

^<em><strong>Test</strong> Test</em>^
```

When `smart_emphasis` == `False`  and `smart_strong` == `True`:

``` html
^<strong><em>Test</em></strong>^

^<strong><em>Test</em> Test</strong>^

<em><strong>Test</strong>^ Test</em>
```

Maybe this is what you intended all along?  This seems like it would make sense as well.  It would allow me to get the original `underscore.txt` test to pass the way it was before.

facelessuser:
Don't worry, I do realize you are busy; so no rush.

Just wanted to let you know that once I have direction from you in regards to the nested smart and non-smart rules, I will make those minor tweaks (if needed).

After that I will also clean up the tests with more pointed cases to test out very specific scenarios (instead of just having a bunch of crazy syntax like it is now, but they were very helpful when writing the regex).  I just didn't want to spend any more time streamlining tests since the specifics might change on me.  I have absolutely no other last minute changes.  At this point I just need to know how to handle smart boundaries in nested scenarios and I can wrap this up.

waylan:
@facelessuser I appreciate that you have put a lot of work into this. Unfortunately, I have been sick this past week and today is the first day that my head feels clear enough to actually take a serious look at this. However, as I do this in my spare time, I doubt I'll have time to really break down your test cases until some time next week. I believe you have the bug with tails fixed. But I want to make sure I agree with your interpretation of the various combinations of em and strong tags (which I haven't looked at yet). From just a quick read of your comments, I wonder if perhaps you are overthinking things. However, I don't have the time to take a closer look right now. Give me a week and I'll try to have some feedback for you (or I'll merge your PR if I like what I see).

facelessuser:
Fair enough.  I  realized the proposal was a bit bold when I made it; but hey, you never know until you try :wink:.  I can always rebase without all the bold/italic changes.  And it really is a personal preference thing.  But hey, it at least motivated me to fix the tail bug.  And I don't mind working off a fork for personal stuff if I turn out to be extremely partial to it.

I had always found this parser to treat triple tokens as a whole new kind of token set (a smart, non-smart hybrid).  It was actually when I was trying to emulate more of the github smart feel that I realized it wasn't going to be possible with the current constraints.  But seeing the tail bug, I realized I could get closer to that feel by fixing it.

Either way, thanks for considering.  And thanks for a great package.  I will await your feedback.  And don't feel bad if you don't like it :smile:.  When you make huge sweeping changes like I did, it is expected that maybe you won't get it in.

facelessuser:
Looks like you are grabbing what you want from here on a test branch.  So feel free to ignore any activity here, or close if you feel necessary.

facelessuser:
New test cases cover all my cases now and helped shake out the remaining bugs I think.  I think I can call this done.  I suspect that for the official package things are moving in a different direction (which is perfectly fine), so I have compiled all of the changes in an extension (tail fix required) for anyone who wishes to play with it: https://gist.github.com/facelessuser/4a220bfa3671ee4d61e9.  This extension doesn't mix smart and dumb rules for a given token `*` or `_`.  So smart is enabled for either all asterisk rules or all underscore rules to keep things simple.  There is no smart_em but not smart_strong for a token.

Summary of what was being addressed:
1. `** test **` should not be valid.  Tokens should not have a space directly after opening and directly before closing.
2. `***test* test***` triple tokens are not a special token unto themselves, but are actually tokens for em nested in strong, and as such, if one of the tokens are terminating like is occurring above, it should not be recognized as a strong, em.
3. Triple tokens, if the single and double are smart, should act like a smart variant such that things like `***test********test***` should be (`<strong><em>test********test</em></strong>`).  So should double and smart by themselves:  `**test********test**` and `*test********test*` => `<strong>test********test</strong>` and `<em>test********test</em>`.
4. `***test*test***` when smart rules is off, should be handled by case 5 below.
5. `***test* test**` should be `<strong><em>test</em> test</strong>` while `***test** test*` should be `<em><strong>test</strong> test</em>`. For sane results, this form will not allow the this `*****test**` => `<strong><em>*</em>test</strong>` and will differ to normal strong  and em `<em>*</em><strong>test</strong>` (output if smart rules are disabled).
6. For the sake of keeping the extension simple, there will be a restriction for all or nothing smart rules.  So if `*` is smart, so should `**` and same for `_` and `__`.  It would greatly reduce the complexity of the rules.  It will make it easy to handle smart boundaries etc.  There is no reason why you couldn't enable smart completely for `_` and not on `*` and be get all the complexity you should need.  If you are really going crazy with nesting, at that point maybe directly using strong and em tags is what is needed.  So in short, when smart is enabled for `*` this would work `text ***test*** text` this would not `text***test***text`.  But if `_` where dumb, you could do `text_**test**_text` and get what you need.

waylan:
@facelessuser I want to thank you for all the work you did on this. As you observed, I am going a slightly different direction. In my opinion most of the tests you had are non-obvious in how they should be parsed -- even by the most experienced markdown users. I certainly have my opinions, but I can't force my exact interpretation on others. I even removed some of the passing tests, because I am not willing to commit to that specific interpretation. Yes, the implementation I'm going with may offer some results that are a little surprising in various edge cases, but in those situations, that should signal to the document author that they need to take a different approach. In fact most of the edge cases can be avoided if underscores and asterisks are mixed (eg: underscores for strong, asterisks for em). The common (obvious) cases work and I'm good with that.

And @facelessuser, I realize you are taking your work to a third party extension (please add it to the [wiki page](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions)) so it is not a waste of time, but it also helped me be more sure that what I'm doing is the right thing. Without all those tests you created, I wouldn't be so confident in my fix. So thank you again. And thanks for the fix to tails. Without that none of this would have happened.

I am closing this PR now. If anyone is interested in @facelessuser's work, they can follow up with him directly at the link he provided above. If anyone is interested in the actual fix, see Issue #253.

facelessuser:
@waylan no worries, I completely understand.  Originally I had only planned to offer the tail fix, then I figured, I'd throw out the other stuff anyways; the worst that could happen was a "no thanks" (I was already working on it as an extension for personal use anyways).

>  Yes, the implementation I'm going with may offer some results that are a little surprising in various edge cases, but in those situations, that should signal to the document author that they need to take a different approach

I agree, and that is true with, I think, any implementation.  If you are mixing a crazy amount of emphasis, you should probably just use html tags directly to get what you want.

>  I realize you are taking your work to a third party extension (please add it to the wiki page) so it is not a waste of time.

I will.  I am experimenting with a bit of stuff right now in regards to markdown https://github.com/facelessuser/mdown/tree/master/mdownx.  So far it is just me exploring what I can do, and what I actually want to do, so a lot of my ideas are in flux (some good, some bad). I haven't settled on what I want to officially offer/use or even how to package them proper for public use.  But when I get to that point, I will probably add it and possibly a couple other's to the wiki.

I think the next thing I plan on looking into is blockparsers (I am really hoping to get fenced blocks embedded proper in other elements).  I think that is actually on the top of my list right now.  But we will see how motivated I can get myself though :smile:.

PR #343, opened 2014-09-21
Allow tab indentation for multiline metadata.

AlexJF:
Since the original Markdown reference does not enforce space-based indentation for Markdown sources (at least as far as I can see), the meta extension should also support tab indentation for continuation lines.


waylan:
Thanks for the report.. However, with your suggested fix, what happens when someone changes [tab_length](https://pythonhosted.org/Markdown/reference.html#tab_length)? 

I believe the bug was introduced in commit 3a1806b3b77d. I think the correct way to fix this should be obvious with a look at that (hint: the same fix is applied to the footnotes extension in that commit).

And a test would be appropriate so this bug doesn't get reintroduced in any later changes.

PR #344, opened 2014-09-22
Get python 2.6 working again.

EnTeQuAk:
- Add importlib to dependencies if Python 2.6
- Ignore warning-capturing if on Python 2.6

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -50,7 +50,12 @@ from .serializers import to_html_string, to_xhtml_string
 __all__ = ['Markdown', 'markdown', 'markdownFromFile']
 
 logger = logging.getLogger('MARKDOWN')
-logging.captureWarnings(True)
+
+try:
+    logging.captureWarnings(True)
+except AttributeError:
+    # Python 2.6 does not contain `captureWarnings`, ignore for now.
+    pass
 
 
 class Markdown(object):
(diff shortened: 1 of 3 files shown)


waylan:
This was an intentional calculated change. I will not be testing in Python 2.6 and therefore will not be supporting Python 2.6. It appears that perhaps there are other issues with running on 2.6 as well (see #345), so, simply installing importlib isn't enough. 

As for the reason: we are running a single codebase on both Python 2 ans 3. The recommended minimum Python version in that scenario is Python 2.7. And as more Python 3 versions get released, it becomes more difficult to continue to support both the new version of Python 3 and Python 2.6. Now that we have added 3.4, it no longer makes sense to also jump through hoops to support Python 2.6.

EnTeQuAk:
Hey @waylan - thanks for getting back regarding this. I know that there might be more issues on 2.6, those were the first and biggest ones I found so far (at least tests and day-to-day business run as usual).

I can totally understand your point and as you make this an official decision I'll try to do my best to support 2.6 for the time being in a fork if that is alright with you. Unfortunately I need to support 2.6 in some projects for now.

waylan:
@EnTeQuAk this is open-source software. You are free to do anything permitted by the [license](https://github.com/waylan/Python-Markdown/blob/master/LICENSE.md).

PR #356, opened 2014-10-17
Fix out of order tail issue

facelessuser:
While experimenting, I stumbled on a issue where tails get inserted out of order if put under the right stress.

Test (the case below is absurd, but I had to recreate the issue without the extension I was testing):

```
This should remain in order:

***I am ___italic_ and__ bold* I am `just` bold**
```

Expected compared against actual:

``` diff
--- right.html  Fri Oct 17 09:28:35 2014
+++ wrong.html  Fri Oct 17 09:28:26 2014
@@ -1,2 +1,2 @@
 <p>This should remain in order:</p>
-<p><strong><em>I am <strong><em>italic</em> and</strong> bold</em> I am <code>just</code> bold</strong></p>
+<p><strong><code>just</code> bold<em>I am <strong><em>italic</em> and</strong> bold</em> I am </strong></p>
```

The fix was very straight forward.  Tails of subnodes, when inserted back into the parent, should be inserted after the subnode.

### DIFF
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -132,7 +132,7 @@ class InlineProcessor(Treeprocessor):
         childResult = self.__processPlaceholders(text, subnode, isText)
 
         if not isText and node is not subnode:
-            pos = list(node).index(subnode)
+            pos = list(node).index(subnode) + 1
         else:
             pos = 0
 
--- a/tests/extensions/nl2br_tail_stress.html
+++ b/tests/extensions/nl2br_tail_stress.html
@@ -1,0 +1,2 @@
+<p>This should remain in order:</p>
+<p><strong><em>I am <strong><em>italic</em> and</strong> bold</em> I am <code>just</code> bold</strong></p>
--- a/tests/extensions/nl2br_tail_stress.txt
+++ b/tests/extensions/nl2br_tail_stress.txt
@@ -1,0 +1,3 @@
+This should remain in order:
+
+***I am ___italic_ and__ bold* I am `just` bold**


waylan:
Thanks for this @facelessuser, but I'm curious why the test is located in the tests for an extension that is not related to this fix. We already have a [file](https://github.com/waylan/Python-Markdown/blob/master/tests/misc/nested-patterns.txt) for testing nested inline elements. Why not add the test to that file?

Also, I notice that you never have more than one line in any of your commit messages. Whereas your pull requests have a nice explanation of both the problem and the fix. Seems like that should be part of the commit message to me (perhaps everything except the example in this instance). When I'm reviewing the commit history to see why something was changed in the past, I look at the commit messages, not the issue/pull request on GitHub - especially if the commit message makes no mention of any GitHub issue to look at. See [this document](https://wiki.openstack.org/wiki/GitCommitMessages#Information_in_commit_messages) for a good summary of best practices. And GitHub provides some good [documentation](https://help.github.com/articles/closing-issues-via-commit-messages/) explaining how to reference issues from within commit messages. In fact, they document [all sorts](https://help.github.com/) of good stuff.

Granted, I don't have a perfect track record with this stuff, but I do try. If you intend to continue to contribute, i would ask that you do the same. I certainly appreciate your contributions, but any future contributions are more likely to be accepted if I have less work to do. For example, if you add a second commit to this PR which moves the test, then I will have to pull your change locally, squash the two commits into one, merge the change, and then push back to GH. However, if you take care of squashing (which would be an excellent opportunity for you to also provide a better commit message), I only need to click one button on GitHub to merge. It may take weeks for me to get to the former, but only hours for the later.

facelessuser:
> Thanks for this @facelessuser, but I'm curious why the test is located in the tests for an extension that is not related to this fix. We already have a file for testing nested inline elements. Why not add the test to that file?

So you do... :smile:.  The issue was exposed when using nl2br, but not without (though I am sure the issue could be forced without).  At the time, it made sense, but less so now that I uncovered the root of the issue.  I will look into moving it.

> Also, I notice that you never have more than one line in any of your commit messages.

I don't know, I guess when pull requesting, I always explain things in the request because I can format the response in Markdown all pretty like.  I am aware of how to link issues and such, if there is an issue, I _usually_ try and link to that (but yeah, not always).   I guess you just get into habits.  I have usually found the issue tracker at my place of employment to contain the actual meat and the commits reference the actual issue as there may be multiple commits, so each commit is more telling you about what happened in the commit opposed to a sweeping explanation of the problem.

I get what you are saying though, it really is all based on the workflow of the development team, and you make some good points.  I don't mind squashing the commits.  I will move the test and re-base and re-pull when I get some time today or tomorrow (I will also explain the issue in the commit more thoroughly).

PR #360, opened 2014-10-24
Changed logging level when loading extensions

topless:
So it suppresses messages for loaded extensions when its used as third party.


lipis:
Yes, please! :+1:

waylan:
Hmm, IIRC, I had set the level to 'info' during development with the intention of changing it to 'debug' before commuting. Guess I forgot about it. Thanks for catching this.

PR #372, opened 2014-11-24
Have Meta extension support YAML

kernc:
The Meta extension currenly only supports extended RFC822-style headers.

Another ubiquitous meta-data format, even prevalent perhaps when it comes to *.md files, is YAML. [Pandoc](http://johnmacfarlane.net/pandoc/) [supports YAML](http://johnmacfarlane.net/pandoc/demo/example9/pandocs-markdown.html#extension-yaml_metadata_block) metadata, [Jekyll uses it](https://help.github.com/articles/using-jekyll-with-pages/#frontmatter-is-required).

It'd be nice if this markdown implementation also supported YAML meta data. Making a plugin is an option, but [the only existing plugin](https://github.com/teoric/python-markdown-yaml-meta-data) seems to not work at all _and_ break the established API.

It'd [make some sense](https://www.google.com/search?q=markdown+metadata) to have YAML meta-data supported out-of-the-box. Please let me know what you think. With heads up, I'll be happy to further write tests and document the addition.

### DIFF
--- a/docs/extensions/meta_data.txt
+++ b/docs/extensions/meta_data.txt
@@ -44,9 +44,16 @@ If a line is indented by 4 or more spaces, that line is assumed to be an
 additional line of the value for the previous keyword. A keyword may have as
 many lines as desired. 
 
-The first blank line ends all meta-data for the document. Therefore, the first 
-line of a document must not be blank. All meta-data is stripped from the 
-document prior to any further processing by Markdown.
+The first blank line ends all meta-data for the document. Therefore, the first
+line of a document must not be blank.
+
+Alternatively, if the first line in the document is `---`, a YAML document
+separator, then the meta-data is searched for between it and the next `---`
+(or `...`) line. Even though YAML delimitors are supported, meta-data is
+not parsed as YAML unless the `yaml` option is set (see below).
+
+All meta-data is stripped from the document prior to any further processing
+by Markdown.
 
 Usage
 -----
@@ -54,7 +61,15 @@ Usage
 See [Extensions](index.html) for general extension usage, specify `markdown.extensions.meta`
 as the name of the extension.
 
-This extension does not accept any special configuration options.
+The following options are provided to configure the output:
+
+* **`yaml`**: Support meta-data specified in YAML format.
+
+    Default: `False`
+
+    If `yaml` is set to `True`, the lines between `---` separators are parsed
+    as a full YAML object. PyYAML is required for this, and a warning is
+    issued if PyYAML (or equivalent) isn't available.
 
 Accessing the Meta-Data
 -----------------------
@@ -85,8 +100,13 @@ line breaks if desired. Or the items could be joined where appropriate. No
 assumptions are made regarding the data. It is simply passed as found to the 
 `Meta` attribute.
 
-Perhaps the meta-data could be passed into a template system, or used by 
-various Markdown extensions. The possibilities are left to the imagination of 
+Note, if `yaml` option is set, the resulting `Meta` attribute is the object as
+returned by `yaml.load()` and may deviate significantly from the above
+description (e.g. may be a list of dicts, with value objects other than
+strings, ...).
+
+Perhaps the meta-data could be passed into a template system, or used by
+various Markdown extensions. The possibilities are left to the imagination of
 the developer.
 
 Compatible Extensions
--- a/markdown/extensions/meta.py
+++ b/markdown/extensions/meta.py
@@ -20,36 +20,68 @@ from __future__ import unicode_literals
 from . import Extension
 from ..preprocessors import Preprocessor
 import re
+import logging
+
+try:
+    import yaml
+    try:
+        from yaml import CSafeLoader as SafeLoader
+    except ImportError:
+        from yaml import SafeLoader
+except ImportError:
+    yaml = None
+
+log = logging.getLogger('MARKDOWN')
 
 # Global Vars
 META_RE = re.compile(r'^[ ]{0,3}(?P<key>[A-Za-z0-9_-]+):\s*(?P<value>.*)')
 META_MORE_RE = re.compile(r'^[ ]{4,}(?P<value>.*)')
+YAML_BEGIN_RE = re.compile(r'^-{3}(\s.*)?')
+YAML_END_RE = re.compile(r'^(-{3}|\.{3})(\s.*)?')
 
 
 class MetaExtension (Extension):
     """ Meta-Data extension for Python-Markdown. """
+    def __init__(self, *args, **kwargs):
+        self.config = {
+            'yaml': [False, "Parse meta data specified as a "
+                            "'---' delimited YAML front matter"],
+            }
+        super(MetaExtension, self).__init__(*args, **kwargs)
 
     def extendMarkdown(self, md, md_globals):
         """ Add MetaPreprocessor to Markdown instance. """
-
-        md.preprocessors.add(
-            "meta", MetaPreprocessor(md), ">normalize_whitespace"
-        )
+        md.preprocessors.add("meta",
+                             MetaPreprocessor(md, self.getConfigs()),
+                             ">normalize_whitespace")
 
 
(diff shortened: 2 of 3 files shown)


waylan:
I have a number of concerns about this.

First, the existing tests are failing. That needs to be fixed. And where's the tests and documentation of the few features?

But before we get to that, why can't this be a seperate extension? And why would I use YAML to define structured data and then throw all that away? The existing meta-data extension has no idea of data types and that's why it returns everything as a list of strings. But with YAML I get type inference that works for free. If I'm using YAML, I want to keep that. Seems to me a seperate extension with a different API makes more sense.

That said, I realize that various existing projects (Pelican, MkDocs, etc.) use the existing extension and (type ignorant) API. I suppose one might be interested in using YAML with such a framework without the need to hack the framework, which this would accomplish. That said, if after carefully crafting my YAML meta-data with a give type, that type is thrown out and a different type is assigned by the framework, I would find that extremely annoying.

I could go either way on this. In fact, as I have recently been playing with some of these frameworks that use the existing extension, I was thInking that it might be useful to provide a way to assign a given type to a specific name. Perhaps when loading the extension, a set of key-value pairs could be passed in as configs that define a type for given key names in the meta data. Something like `{title: string, date: datetime, summary: multiline}`. That way, the returned data would already be of the give type. I may reach out to the developers of some of these frameworks for some feedback. Not sure how adding YAML support fits into that. I'll have to chew on that.

waylan:
After thinking on this, I believe the best approach is to follow [MultiMarkdown](http://fletcherpenney.net/multimarkdown/)'s [lead](http://fletcher.github.io/MultiMarkdown-4/syntax#metadata) and optionally allow the YAML deliminators (`---` & `...`), but not actually use YAML to parse the content. That way, YAML would still be able to parse the content (in implementations that support YAML), but our implementation/API would not change any (and would avoid a third-party dependency).

I suppose we could also add a config option to turn on YAML parsing which should return the data as YAML parses it. But the default would be to have that feature turned off (for backward compatibility) even when the YAML deliminators are present.

That way, existing frameworks will continue to work as-is, and their users could even add the YAML deliminators (which cause the metadata to get rendered as a table by GitHub for example). And, if any framework actually wants to have real YAML parsing, they can turn on the YAML parsing and get structured data. In fact, I like that better that providing an API to define types for the current implementation (as I suggested in my previous comment). if you want structured data, turn on YAML and if you want raw data (the current API), turn off YAML---either way we should still support the YAML deliminators.

kernc:
Ok, I agree to what you are saying about retaining structure. Took me 40 minutes yesterday to figure out it's by design I can't force-`load` YAML as pure strings. :smiley: 

So, is something like this (PR updated) what you had in mind? The dependency on PyYAML is a soft one, required only if YAML is enabled.

waylan:
Not exactly. You added support for YAML, but you did not add support for the YAML deliminators when YAML is not turned on---which should just be a couple more lines of code.

Also, let's not import yaml within the method. I'd prefer to import at the module level and if it fails assign `False` to it. Something like:

``` python
try:
    import yaml
except ImportError:
    yaml = False

# then later in the method:

        if self.config['yaml'] and yaml ...
```

Also, lets call the config key 'yaml', no reason to make it so lengthy.

kernc:
So if YAML is not turned on, what happens, the YAML header is just skipped?

And regarding

``` python
        if self.config['yaml'] and yaml
```

If YAML is requested but `yaml` is not available, the extension shouldn't pass that silently, correct?

kernc:
Oh, I get it. You want to enable just skipping `---` regardless of what mode is on. Then, if YAML is disabled, just parse the contents between `---` as regular meta data. On it!

**Update**: But what if a YAML header, say a multi-level or a list example of it, is unparsable as simple meta-data (when parsing YAML is disabled)?

waylan:
The contents of a document should never cause the parser to raise an error (we can't be crashing web servers---the libraries most common use case). Your implementation would raise an error only when a document contained YAML but the yaml lib was not available. However, the error would only occur when trying to parse the document (not on class initialization). Using my suggested method, the parser doesn't raise an error---the results are just not what was expected. Perhaps not as easy to debug---but good documentation should address that.

kernc:
Ok, how about now? Not crashing anywhere, only a warning is logged iff YAML parsing is enabled and a YAML-header'd document is encountered and `yaml` is unavailable.

markdown/extensions/meta.py
```diff
@@ -20,21 +20,40 @@
...
+        super(MetaExtension, self).__init__(*args, **kwargs)
 
     def extendMarkdown(self, md, md_globals):
         """ Add MetaPreprocessor to Markdown instance. """
-
-        md.preprocessors.add(
-            "meta", MetaPreprocessor(md), ">normalize_whitespace"
-        )
+        preprocessor = MetaPreprocessor(md)
+        preprocessor.config = self.getConfigs()
```
waylan: Shouldn't we be passing the configs into the Class on initiation? Like [this](https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/wikilinks.py#L49).

kernc: Dunno. Inspecting all the built-in extensions, I've seen examples of both, and more. This way, we avoid one thin `__init__()`. Since this isn't functional programming and since `MetaPreprocessor` and `MetaExtension` are tangled one way or another, doesn't really matter IMHO. Unless you have a preference. ¯\_(ツ)_/¯

waylan: Yes, this is my preference. It makes for easier to read code later and it is the way the API is designed to work. If other extensions don't follow this pattern, they should be updated--unless there is a good reason not to of course.

kernc: Done. FTR, the given wikilinks example sets `md` attribute exactly the same way:

``` python
        wikilinkPattern.md = md
```

kernc:
Whaddya think? :smiley:

tests/test_extensions.py
```diff
@@ -501,17 +502,19 @@ def testBasicMetaData(self):
 Blank_Data:
 
 The body. This is paragraph one.'''
-        self.assertEqual(
```
waylan, line 504: Let's leave this test as-is and add a seperate test with the YAML deliminators. That way we have a test for each possible scenario.

docs/extensions/meta_data.txt
```diff
@@ -44,9 +44,15 @@ If a line is indented by 4 or more spaces, that line is assumed to be an
...
 
-The first blank line ends all meta-data for the document. Therefore, the first 
-line of a document must not be blank. All meta-data is stripped from the 
-document prior to any further processing by Markdown.
+The first blank line ends all meta-data for the document. Therefore, the first
+line of a document must not be blank.
+
+Alternatively, if the first line in the document is `---`, a YAML document
+separator, then the meta-data is searched for between it and the next `---`
+(or `...`) line.
```
waylan: Let's add a warning here that even though YAML  deliminators are supported,  the metadata is not parsed as YAML unless the YAML option is set.

kernc: Done.

tests/test_extensions.py
```diff
@@ -530,6 +553,25 @@ def testMetaDataWithoutNewline(self):
         self.assertEqual(self.md.convert(text), '')
         self.assertEqual(self.md.Meta, {'title': ['No newline']})
 
+    def testYamlObjectMetaData(self):
+        """ Test metadata specified as a complex YAML object. """
+        md = markdown.Markdown(extensions=['markdown.extensions.meta(yaml=True)'])
```
waylan: Sorry, I missed this before. This way of passing config options into extensions is pending deprecation. I realize some existing tests still use this, but let's not add any more. We'll just have to change it before the next release anyway.

kernc: Oh, no problem. Hope now is fine?

waylan:
@kernc thanks for the work on this. I just merged your work with a couple minor changes.

kernc:
Yay, thanks! :smiley:

PR #381, opened 2014-12-31
Fix a small typo

svetlyak40wt:
By the way, why do you use such strange indenting inside this section?


waylan:
Not sure why that is strange indenting. We are nested in a list so that requires at least 4 spaces and we are in a code block which requires another 4 spaces for 8 spaces of indent. Anyway, thanks for the patch.

PR #383, opened 2015-01-30
Add support for Python 2.6

msabramo:


waylan:
I see you closed this before I had a chance to respond. Perhaps you saw the tests were failing? In any event, for future reference, please see my comment on #344. My position has not changed,

msabramo:
Ah ok I guess I need to do better searching. Thanks for clarifying! I closed it because I saw it was going to be more difficult than I thought and not worth the effort. So I agree with your position.

PR #385, opened 2015-02-05
Improve robustness of extension configs.

nekoya:
I found the problem when Python-Markdown working with py-gfm (Github-Flavored Markdown extension).

https://github.com/dart-lang/py-gfm

```
  File "/path/to/venv/local/lib/python2.7/site-packages/markdown/extensions/__init__.py", line 36, in __init__
    self.setConfigs(kwargs.pop('configs', {}))
  File "/path/to/venv/local/lib/python2.7/site-packages/markdown/extensions/__init__.py", line 75, in setConfigs
    for key, value in items:
TypeError: 'NoneType' object is not iterable
```

It is because the extension initialize configs as None.

Simply it is the extension's matter, but I think that more robust code of Python-Markdown is the better way.

### DIFF
--- a/markdown/extensions/__init__.py
+++ b/markdown/extensions/__init__.py
@@ -76,8 +76,9 @@ class Extension(object):
         if hasattr(items, 'items'):
             # it's a dict
             items = items.items()
-        for key, value in items:
-            self.setConfig(key, value)
+        if items:
+            for key, value in items:
+                self.setConfig(key, value)
 
     def extendMarkdown(self, md, md_globals):
         """


waylan:
Actually, you need to change your extension to not pass in `None` to the 'config' kwarg. If you notice, that is being [deprecated](https://github.com/waylan/Python-Markdown/blob/master/markdown/extensions/__init__.py#L40). In short you should change this:

``` python
def makeExtension(configs=None):
    return GithubFlavoredMarkdownExtension(configs=configs)
```

to this:

``` python
def makeExtension(*args, **kwargs):
    return GithubFlavoredMarkdownExtension(*args, **kwargs)
```

That way your extension won't pass in `None` by default.

That said, It appears that I missed upgrading this from a `PendingDeprecationWarning` to a `DeprecationWarning`, so thanks for the reminder. And I should add a note to the release notes as well. There will be a number of subtle changes to the Extension API over the next release or two, so all extension authors should be reading the release notes carefully.

waylan:
I've taken an alternate approach in 75855cd.

nekoya:
Thank you for the great advice.

I try to update the extension.

PR #395, opened 2015-03-09
if border is true, set border attrib on table to 1

neurobashing:


waylan:
Thanks for the suggestion. However I will not be merging this pull request. Please see my response to issue #74 for a full explanation of why I will not consider any additions to the table extension.

Even if that wasn't the case, this extension is intended to be a clone of PHP Markdown Extra's behavior. Therefore, I would not accept the new behavior for that reason either.

For that matter, the feature you are asking for seems more like something that should be implemented through CSS, not HTML, which is outside the cope of Markdown. 

If you would still like to have this feature, then my suggestion would be to release it as a third party extension and list it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) for others to find.

PR #400, opened 2015-04-04
support empty table

pieterprovoost:


waylan:
@pieterprovoost thank you for your submission. However, before I accept your PR, please review my response to #74. Does the change you made more closely match the behavior of tables in PHP Markdown Extra? If so, could you provide an example table (preferably as a test that fails without your change and passes with it). If not, then may I suggest you fork the existing extension and make it do whatever you want -- or perhaps a [third party extension](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) already exists that meets your needs.

pieterprovoost:
@waylan PHP Markdown supports zero rows, so this change does make Python Markdown more like PHP Markdown. This test fails with the current version:

```
import markdown
html = markdown.markdown("|a|b|\n|--|--|\n", extensions=['tables'])
assert("<table>" in html) 
```

waylan:
Thanks. If you add a test [here](https://github.com/waylan/Python-Markdown/blob/master/tests/extensions/extra/tables.txt), I'll accept it.

pieterprovoost:
@waylan I added the test to my branch, but it's not picked up? https://github.com/pieterprovoost/Python-Markdown/commit/47737e4a515e63e21bbfd2d2b35404fe44432890

waylan:
Well, you didn't also update the corresponding HTML file which provides the expected output for the test to compare against. If you run `run-tests.py update` the updated test should be found and the HTML file automaticaly regenerated. 

Although now that I take a second look, I see that the tests are passing, which is odd. Does the `update` command detect the change? Or do I have a bug in the testing framework?

Issue #401, opened 2015-04-05
Footnotes don't get reset after Postprocess


TintypeMolly:
## Problem

Once markdown instance converted text containing footnotes, those footnotes' contents keep appearing at every single result converted by the markdown instance.
## How to reproduce

```
import markdown

md = markdown.Markdown(extensions=["markdown.extensions.footnotes"])
md.convert("Foo[^1] Bar\n[^1]: Baz")
print md.convert("Foo")
print md.convert("Bar")
```
## Desired Output

```
<p>Foo</p>
<p>Bar</p>
```
## Current Output

```
<p>Foo</p>
<div class="footnote">
<hr />
<ol>
<li id="fn:1">
<p>Baz&#160;<a class="footnote-backref" href="#fnref:1" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
</ol>
</div>
<p>Bar</p>
<div class="footnote">
<hr />
<ol>
<li id="fn:1">
<p>Baz&#160;<a class="footnote-backref" href="#fnref:1" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

waylan:
As [documented](https://pythonhosted.org/Markdown/reference.html#convert), this is a feature, not a bug. You need to call the `reset` method on the Markdown class between each call to convert. There may be valid reasons why a user may not want to reset between documents. For example, on the front/home page one could collect all the footnotes for multiple posts and display them al together at the bottom. As an added bonus, all of the links to the footnotes would then be unique with no duplicates.

TintypeMolly:
I can't grasp your explanation. Can you give me a concrete example?

TintypeMolly:
I can't grasp your explanation. Can you give me a concrete example? I know some kind of things(e.g. Meta-Data) have some advantages on keep their state but not with footnotes.

waylan:
The one and only way to reset state should be to call `Markdown.reset()`. No other extension gets reset called without it explicitly being requested by the user, therefore we should not expect anything different for footnotes.

Regarding my stated use case, I had forgotten that when the `UNIQUE_IDS` config option is set to `True` it works across calls to reset. That said, the expectation is that there would be calls to reset - which is clearly documented.

PR #402, opened 2015-04-05
Reset footnotes after postprocessing

TintypeMolly:
This is fix pull request for #401


waylan:
As the user should be calling `Markdown.reset()` between calls to `Markdown.convert()` this is unnecessary and breaks the expected (and documented) behavior.

PR #403, opened 2015-04-06
Use zip() for positional argument compatibility

iKevinY:
This PR refactors `__init__.py` to use `zip()` for positional argument compatibility, and also fixes some miscellaneous capitalization/spelling errors.
- The list returned by `zip()` is automatically truncated to the length of the shortest argument list, which greatly simplifies the code that loops through the list of positional arguments.
- The expression `if len(args):` relies on the truthiness of `len(args)`; using `if args:` does not change the behaviour of the conditional, but is easier to understand at a glance (since what's really being checked is whether or not there are any arguments).

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -114,13 +114,11 @@ class Markdown(object):
 
         # For backward compatibility, loop through old positional args
         pos = ['extensions', 'extension_configs', 'safe_mode', 'output_format']
-        for c, arg in enumerate(args):
-            if pos[c] not in kwargs:
-                kwargs[pos[c]] = arg
-            if c+1 == len(pos):  # pragma: no cover
-                # ignore any additional args
-                break
-        if len(args):
+        for arg, pos_arg in zip(args, pos):
+            if pos_arg not in kwargs:
+                kwargs[pos_arg] = arg
+
+        if args:
             warnings.warn('Positional arguments are deprecated in Markdown. '
                           'Use keyword arguments only.',
                           DeprecationWarning)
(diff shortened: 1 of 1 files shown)


mitya57:
+1 from me.

waylan:
Sorry but I'm having a little trouble understanding why this pull request was made. I'm sure the change is a good one -- if we were keeping that code, but the change even affected the code which raises the `DeprecationWarning` which indicates that this code will all be removed in the next release. If the code will all be deleted in the next release, then why make this change at all?

In fact, while working on the next release, I have already [deleted this code](https://github.com/Python-Markdown/markdown/commit/344f0f91697a3a65ca54aac3fc3da9ffa5f8ced3). So it is unclear what value making this change would serve.

iKevinY:
Ah, it didn't occur to me that this code was being removed in the next release; I agree that this change is unnecessary. Would it be alright to repurpose this PR into one that just contains iKevinY@40c968e?

waylan:
Yes, that would be great. Thanks.

iKevinY:
I found it irritating that the branch name was still going to be `minor-refactoring` despite not containing any refactoring, so I cherry-picked the commit from a new branch and opened #404.

mitya57:
> If the code will all be deleted in the next release, then why make this change at all?

@waylan, do you mean that there will be no more releases from _this_ repository? If yes, maybe it will make sense to stop accepting pull requests here and mention in the README that the code has been moved?

waylan:
This repo is in "maintenance mode" and may receive future minor bug-fix releases. So far all accepted changes have fallen under that category. However you make a point, I probably should make a note of that somewhere.

PR #405, opened 2015-04-19
Change sanitize_url method of the LinkPattern class so it can handle Twitter pic URL

ap-Codkelden:
When parsing links like Twitter pics with `size` attribute (see Twitter [docs](https://dev.twitter.com/overview/api/entities-in-twitter-objects)), e. g. [this](https://pbs.twimg.com/media/CCOzV7vUgAE79pq.jpg:large), Python-Markdown return HTML tag `<a>` with empty `href` attirbute, because `sanitize_url` make of colon  in "path", "parameters", "query" or "fragment" is suspect.

### DIFF
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -410,9 +410,12 @@ class LinkPattern(Pattern):
 
         for part in url[2:]:
             if ":" in part:
-                # A colon in "path", "parameters", "query"
-                # or "fragment" is suspect.
-                return ''
+                # Check if this is just Twitter media link with 'size'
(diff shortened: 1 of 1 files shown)


waylan:
Just don't use 'safe_mode'. In fact safe_mode is deprecated will be removed in the next release. Therefore I see no reason to make any changes to it. If you don't like its current behavior then you should turn if off and use a different solution (like Bleach, as suggested in the docs). See the [release notes](https://pythonhosted.org/Markdown/release-2.6.html#safe_mode-deprecated) for more details.

ap-Codkelden:
A, ok. Thank you for explanation.

PR #411, opened 2015-05-20
Fixed #410

### LINKED ISSUE #410: Strange behaviour of returning symbol
According to the source, the return-back arrow is only inlined into last `p` element of footnoted text.

This leads to a strange cases when footnote ends with `blockquote` or `header` elements.

2015-05-17 14-07-03 new tab - mozilla firefox

This probably could be fixed by some weird CSS like

```
div.footnote p ~ blockquote:last-of-type
{
    display: inline-block;
    width: 90%;
}

div.footnote blockquote + p
{
    display: inline;
}

div.footnote p ~ h3:last-of-type
{
    display: inline;
}

div.footnote h3 + p
{
    display: inline;
}
```

But I suggest just removing the workaround from [here](https://github.com/georgthegreat/Python-Markdown/blob/master/markdown/extensions/footnotes.py#L158)

georgthegreat:
Link is correctly appended to the bottommost element of the markup.

### DIFF
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -153,14 +153,17 @@ class FootnoteExtension(Extension):
             )
             backlink.text = FN_BACKLINK_TEXT
 
-            if li.getchildren():
-                node = li[-1]
-                if node.tag == "p":
-                    node.text = node.text + NBSP_PLACEHOLDER
-                    node.append(backlink)
-                else:
-                    p = etree.SubElement(li, "p")
-                    p.append(backlink)
+            #appending backlink to the lowermost element 
+            #of last tag in footnote
+            elem = li
+            while True:
+                children = elem.getchildren();
+                if len(children) == 0:
+                    break;
+                elem = children[-1]
+        
+            elem.text = node.text + NBSP_PLACEHOLDER
+            elem.append(backlink)
         return div
 
 
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -162,7 +162,7 @@ class FootnoteExtension(Extension):
                     break;
                 elem = children[-1]
         
-            elem.text = node.text + NBSP_PLACEHOLDER
+            elem.text += NBSP_PLACEHOLDER
             elem.append(backlink)
         return div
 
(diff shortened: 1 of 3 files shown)


mitya57:
With the following source:

``` markdown
footnote[^1]

[^1]: some text

        code block
```

I get this HTML:

``` html
<p>footnote<sup id="fnref:1"><a class="footnote-ref" href="#fn:1" rel="footnote">1</a></sup></p>
<div class="footnote">
<hr />
<ol>
<li id="fn:1">
<p>some text</p>
<pre><code>code block
&#160;
<a class="footnote-backref" href="#fnref:1" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></code></pre>
</li>
</ol>
</div>
```

A link _inside_ a code block is wrong.

Actually, I think we should keep the current behaviour and not introduce hacks.

georgthegreat:
I've posted rationale in #410.
Styling backreference symbol in current markup is a pain.
Styling bakcreference after the change seems to be quite easy (single CSS rule to make it block element).

Though this leads to a breaking change in markup, I'm not sure if anyone (except for myself) used such complicated markups at all.

georgthegreat:
BTW, I don't see anything wrong with the markup from above.
Backref Is rendered on another line, which seems to be wrong.

waylan:
Absolutely not. A link does not go **inside** a code block. The only way to add a backref link when a footnote ends in a codeblock is to add a new paragraph. I will not consider any other alternative.

>  I'm not sure if anyone (except for myself) used such complicated markups at all.

I think that may be the case. In all the years I've been involved, I've never seen anyone make this complaint before. And most documents I read have simple footnotes that contain nothing more than a paragraph or two. As I stated in #410 I don't think a solution can be found which gives you what you want and accounts for all of the edge cases properly. Of course, if you prove me wrong, I will reconsider. However, this PR only proves me right so far.

That said, you are free to fork the extension and maintain your own which behaves in whatever way to want. Just be sure you do so within the limits of the [license](https://github.com/waylan/Python-Markdown/blob/master/LICENSE.md).

georgthegreat:
I will not consider forking as an alternative.
If you think that this PR isn't doing anything good, you may close it, I can live without it (arrows on a separate line are annoying, but not _that annoying_).

BTW, tests are continuing to fail, I know about the problem :)

PR #415, opened 2015-06-01
Fixed #414 parser ignoring value of `tab_length`

### LINKED ISSUE #414: `tab_length` parameter ignored
Input file:

``` markdown
* Test 1
  * Test A
  * Test B
* Test 2
```

run with `markdown.markdownFromFile('input.md', tab_length=2)` produces output file:

``` html
<ul>
<li>Test 1</li>
<li>Test A</li>
<li>Test B</li>
<li>Test 2</li>
</ul>
```

when it should produce a nested list.

I think I've found the issue. In `markdown/blockprocessors.py` line 304: 

``` python
RE = re.compile(r'^[ ]{0,3}\d+\.[ ]+(.*)')
```

`3` should be `self.tab_length - 1`. Similarly for `CHILD_RE` and `INDENT_RE` in `OListProcessor`, `RE` in `UListProcessor`, and `CHILD_RE` in both `SaneOListProcessor` and `SaneUListProcessor`.

When I make these changes locally, I get the expected output for my test case, and all other tests pass as well. See [this commit](https://github.com/karepker/Python-Markdown/commit/6cbc66b4e63de0ddba7413b653591274fc6587cf) in my fork of your repository.

I can send you a pull request with these changes if you'd like.

karepker:
Changed hard-coded numbers assuming tab length to be 4 to use `self.tab_length`.

This change applies to he default parsers in `mardown/blockprocessors.py` and those in `markdown/extensions/sane_lists.py`.

All current tests should pass.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -311,18 +311,13 @@ class OListProcessor(BlockProcessor):
     def __init__(self, parser):
         BlockProcessor.__init__(self, parser)
         # Detect an item (``1. item``). ``group(1)`` contains contents of item.
-        self.RE = re.compile(
-                ''.join([r'^[ ]{0,', str(self.tab_length - 1), 
-                         r'}\d+\.[ ]+(.*)']))
+        self.RE = re.compile(r'^[ ]{0,%d}\d+\.[ ]+(.*)' % (self.tab_length - 1))
         # Detect items on secondary lines. they can be of either list type.
-        self.CHILD_RE = re.compile(
-                ''.join([r'^[ ]{0,', str(self.tab_length - 1), 
-                         r'}((\d+\.)|[*+-])[ ]+(.*)']))
+        self.CHILD_RE = re.compile(r'^[ ]{0,%d}((\d+\.)|[*+-])[ ]+(.*)' %
+                                   (self.tab_length - 1))
         # Detect indented (nested) items of either type
-        self.INDENT_RE = re.compile(
-                ''.join([r'^[ ]{', str(self.tab_length), ',', 
-                         str(self.tab_length * 2 - 1), 
-                         r'}((\d+\.)|[*+-])[ ]+.*']))
+        self.INDENT_RE = re.compile(r'^[ ]{%d,%d}((\d+\.)|[*+-])[ ]+.*' %
+                                    (self.tab_length, self.tab_length * 2 - 1))
 
     def test(self, parent, block):
         return bool(self.RE.match(block))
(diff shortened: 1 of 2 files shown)


markdown/blockprocessors.py
```diff
@@ -314,6 +308,22 @@ class OListProcessor(BlockProcessor):
     # List of allowed sibling tags.
     SIBLING_TAGS = ['ol', 'ul']
 
+    def __init__(self, parser):
+        BlockProcessor.__init__(self, parser)
```
waylan: Shouldn't `super` be called here (and in all other subclassesd below)?

markdown/blockprocessors.py
```diff
@@ -314,6 +308,22 @@ class OListProcessor(BlockProcessor):
...
+        # Detect an item (``1. item``). ``group(1)`` contains contents of item.
+        self.RE = re.compile(
+                ''.join([r'^[ ]{0,', str(self.tab_length - 1),
```
waylan: Wouldn't it make sense to use string formatting here?

``` python
r'^[ ]{{0,{0:d}}}\d+\.[ ]+(.*)'.format(self.tab_length - 1)
```

Granted, needing to escape the curly brackets in the regex is slightly annoying. But the old `%`-substitution would work as well:

``` python
r'^[ ]{0,%d}\d+\.[ ]+(.*)' % (self.tab_length - 1)
```

In both cases, the type is defined as an integer (`d`) and converted for us. No need to call `str` explicitly.

It also makes it much easier to reed the regex and the latter is actually the method used elsewhere in the code base.

waylan:
@karepker if you address my two comments and fix the white space issues raised by the [flake8 test](https://travis-ci.org/waylan/Python-Markdown/jobs/64950954) I'll be happy to merge this.

markdown/blockprocessors.py
karepker: Are you targeting compatibility with Python 2? If so, I'm not sure `super()` would work in this context, because using it [requires new style classes](https://docs.python.org/2/library/functions.html#super).

Also, a call to `super()` seems like it would be inconsistent in the code base. In `ListIndentProcessor.__init__(...)` method (line 144 in `markdown/blockprocessors.py`), there is a call to `BlockProcessor.__init__(...)`.

karepker:
See my questions regarding the use of `super()`. I'll make the other changes.

markdown/blockprocessors.py
waylan: Looks like I missed updating a class to new style classes in 794bffb4f1. As the [release notes](https://pythonhosted.org/Markdown/release-2.3.html#backwards-incompatible-changes) stated (for 2.3 -- that was a while ago):

> All classes are now “new-style” classes. In other words, all classes subclass from ‘object’.

However, I can clearly see that [Blockprocessor](https://github.com/waylan/Python-Markdown/blob/e594213ab689847ee7d9654817b8fa80fafb0fb7/markdown/blockprocessors.py#L41) does not. Looks like this whole file needs to be updated to be consistent with the rest of the library. If want to do that, great. If not, I'll take care of it myself.

karepker: Sure, I'm happy to make this change. I'll update the pull request shortly.

waylan:
Awesome. Thanks.

PR #417, opened 2015-06-15
Fix processing angle quotes in smarty extension

mitya57:
I have got the following bug report privately:

``` pyconsole
>>> import markdown
>>> from markdown.extensions.smarty import SmartyExtension
>>> md = markdown.Markdown(extensions=[SmartyExtension(smart_angled_quotes=True)])
>>> md.convert("<<hello>>")
'<p>&lt;<hello>&gt;</p>'
```

(Expected result will be `'<p>&laquo;hello&raquo;</p>'`).

This pull request fixes it. @waylan: I have not pushed it directly because you may know better ways to fix it, than separating the processors. If I just move everything to before inline, then I get the Python-Markdown internals leaking into the resulting HTML.

### DIFF
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -211,10 +211,10 @@ class SmartyExtension(Extension):
         rightAngledQuotePattern = SubstituteTextPattern(
             r'\>\>', (self.substitutions['right-angle-quote'],), md
         )
-        self.inlinePatterns.add(
+        self.angledQuotesPatterns.add(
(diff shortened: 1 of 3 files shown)


waylan:
Looks fine to me. Thanks.

mitya57:
Oops, now that regex seems to match even text inside code blocks :-(

I will try to look at that in the coming days, in the mean time suggestions on how to fix this are appreciated.

The issue is not _that_ severe because `smart_angle_quotes` is disabled by default.

PR #419, opened 2015-06-19
Use a different approach to fix the smarty bug

mitya57:
Just update `HTML_RE` so that it doesn't act on `<<...>>`.

### DIFF
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -83,7 +83,7 @@ smartypants.py license:
 
 from __future__ import unicode_literals
 from . import Extension
-from ..inlinepatterns import HtmlPattern
+from ..inlinepatterns import HtmlPattern, HTML_RE
 from ..odict import OrderedDict
 from ..treeprocessors import InlineProcessor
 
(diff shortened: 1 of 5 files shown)


markdown/inlinepatterns.py
```diff
@@ -151,7 +151,7 @@ def build_inlinepatterns(md_instance, **kwargs):
 AUTOMAIL_RE = r'<([^> \!]*@[^> ]*)>'
 
 # <...>
-HTML_RE = r'(\<([a-zA-Z/][^\>]*?|\!--.*?--)\>)'
+HTML_RE = r'(\<([a-zA-Z/][^\>]*?|\!--.*?--)\>)(?!\>)'
```
waylan: I think this is a sensible approach when the smarty extension is in use, but the default behavior should not be changed when smarty is not enabled. I would expect that the smarty extension would replace the HTML pattern with a new one if angle quotes are enabled. Otherwise, leave it unchanged.

And we should have a test for each state (enabled and not enabled). The not enabled state should return the same results as we would get without smarty before this patch was applied.

mitya57: Is `<foo>>` really a valid HTML tag?

waylan: No, of course not. But the parser is not concerned with whether raw HTML is "valid". Without smarty's angle quotes enabled, I expect the same default behavuor we have now: `<foo>>` becomes `<foo>&gt;`.

mitya57:
@waylan I have now updated the pull request as you suggested.

PR #423, opened 2015-08-05
Inline CSS class specification

cyisfor:
For instance the Rust documentation (reference.md) uses markdown that specifies CSS classes for fenced code, like ```{.classa .classb} which produces unbalanced fenced code currently. This should fix it so those classes are parsed and added to the resulting HTML, and the fenced code isn't made unbalanced.

Also removed that LANG_TAG substitution, since assuming the name of the class attribute in HTML to be "class" is pretty safe to make, plus not all CSS classes are only for specifying language highlighting.

No syntax errors in python2.7, python3.4 or pypy3

### DIFF
--- a/markdown/extensions/fenced_code.py
+++ b/markdown/extensions/fenced_code.py
@@ -40,11 +40,12 @@ class FencedBlockPreprocessor(Preprocessor):
 (\{?\.?(?P<lang>[a-zA-Z0-9_+-]*))?[ ]*  # Optional {, and lang
 # Optional highlight lines, single- or double-quote-delimited
 (hl_lines=(?P<quot>"|')(?P<hl_lines>.*?)(?P=quot))?[ ]*
+# Optional CSS classes
+(?P<classes>[a-zA-Z0.9_\\.+ -]+)?[ ]*
 }?[ ]*\n                                # Optional closing }
 (?P<code>.*?)(?<=\n)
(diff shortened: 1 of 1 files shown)


cyisfor:
and of course another file in the Rust doc also decides to use some weird ```lang,class,class syntax, but I give up. -_-

waylan:
@cyisfor that you for your contribution. However, before I accept your patch, I have a few concerns I would like to address: 

What do you mean by unbalanced? I don't see any invalid HTML so you must be referring to something else.

Also, could you include a test to demonstrate that your change works. Specifically, a test that fails before the change is applied and passes after.

Additionally, is appears that you [broke](https://travis-ci.org/waylan/Python-Markdown/jobs/74296545) an existing test. There are a large number of existing users who are expecting the current output (and they already have existing CSS which works with it). If we change the output, we will break their sites when they upgrade. Is the proposed change worth it?

It also appears that you need to clean up your code to be [flake8 compliant](https://travis-ci.org/waylan/Python-Markdown/jobs/74296546).

And finally, read carefully over my [analysis](http://achinghead.com/marking-up-code.html) of how code blocks should be marked up in HTML. Anything that does not fit that model is likely to be rejected.

waylan:
Given that lack of response on this issue, I'm closing this for now. I'll reconsider if and when the issues I raised are addressed.

cyisfor:
Yeah sorry, I just don't have a reason to do this outside of one rather stupid documentation package that just isn't worth it.

PR #429, opened 2015-08-28
Fix find footnote placeholder to recurse

Tigge:
The footnotes extension actually didn't recurse correctly when looking for the place marker. If it wasn't a child of the root element it would traverse and find the place marker but still return None in the `finder` function.


waylan:
@Tigge, thanks for your contribution. Before I accept it, could you provide a test which fails before this change is applied and passes after? I'm not sure I see what this is accounting for.

Tigge:
@waylan Test added!

PR #434, opened 2015-09-25
Fully remove YAML support (to make it easier for third party extensions to take over the job)

joachimneu:
YAML support was added and then again removed from the "meta" extension -- since YAML metadata could, should and is be(ing) provided by an additional third party extension. However, the "meta" extension is still parsing metadata that looks like YAML (starting and ending with '---'), but without actually parsing the YAML correctly. This is not only misleading, but also makes it hard for "normal" metadata and YAML-formatted metadata and the corresponding extensions to co-exist (because the "meta" extension consumes the metadata that is supposed to be consumed by the YAML metadata extension).

This pull request provides:
1. Full removal of pseudo-YAML capabilities from the "meta" extension (+ updated doc)
2. The "meta" extension does not assume anymore to be the only extension to write to the `markdown.Meta` attribute, but respects values another extension might already have put there
3. Two lines I had to add to `tests/__init__.py` to make tests work on my machine

Hope this helps. Let me know if you have any suggestions. :)

### DIFF
--- a/docs/extensions/meta_data.txt
+++ b/docs/extensions/meta_data.txt
@@ -23,7 +23,7 @@ This extension is included in the standard Markdown library.
 Syntax
 ------
 
-Meta-data consists of a series of keywords and values defined at the beginning 
+Meta-data consists of a series of keywords and values defined at the beginning
 of a markdown document like this:
 
     Title:   My Document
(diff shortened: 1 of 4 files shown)


waylan:
The existing implementation is an implementation of MultiMarkdown's [meta-data format](http://fletcher.github.io/MultiMarkdown-4/metadata.html). As that format includes support for (optional) deliminators, so will Python-Markdown's implementation. If you want to support YAML meta-data you may still do so. Simply do not use the meta-data extension. In fact, note that for the brief period when this extension supported both, it did not support both simultaneously. You could only get MMD meta-data or YAML meta data, not both. Therefore, leaving the current behavior as-is creates no conflict or lose of features.

waylan:
As a reminder, you always have the option to create your own extension which does whatever you want. Just because I reject a specific feature request, does not mean you can't have the feature. It just means that it is not something I am interested in developing and maintaining myself.

joachimneu:
Any chance we can get something like https://github.com/joachimneu/Python-Markdown/commit/6731cd196d0f4762007959fbe51cd07a06ae7245#diff-52418a280f3d8bb9064dc35e3532e77aR69 (only those four lines), i.e. "meta" not deleting metadata without substitution that had already been extracted by some other third party metadata extension before?
This would allow for multiple metadata extensions to be enabled at the same time. For example, in a [Pelican](https://github.com/getpelican/pelican) website, there might be both articles with traditional and articles with yaml metadata, and they all get parsed by markdown with the same set of extensions (both "meta" and "metayaml" enabled), because we can't tell beforehand which type of metadata we will stumble upon in a particular document.

waylan:
So, you would like the following change:

```
-        self.markdown.Meta = meta
+        if hasattr(self.markdown, 'Meta'):
+            self.markdown.Meta.update(meta)
+        else:
+            self.markdown.Meta = meta
```

It is a little more complicated than that. As it is, that would not reset the meta data between multiple pages (when the same instance of the Markdown class is used to process multiple pages). If the first page has data `foo: bar` and the second page does not define `foo`, then the second page will still have `foo` set to `"bar"`. Of course, the extension could be altered to make it a registered extension and a reset method could be defined. My point is that this is more work than a simple four line change.

Additionally, we are currently in a feature freeze (especially with regard to extensions) while we work toward version 3.0. See the [3.0 Roadmap](https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap) for more info. When the extensions get split out into their own repos (third list item), re-raise this and we will address it then.

joachimneu:
Okay, I see. Thank you for taking the time for such detailed responses anyways! :)

PR #437, opened 2015-10-22
Table inline code

MustafaHaddara:
There's probably a better way to do this -- I feel like `tables.py` should import from `inlinepatterns` or something like that. 

Tests passed, and I added 4 tests for tables. On the `master` branch, the first 3 pass. These commits get the 4th one working. 

Resolves https://github.com/waylan/Python-Markdown/issues/436


markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
+                current = ''
+            else: 
+                current += i
+        if escaped:
+            ## This should never happen
+            return row.split(marker)
+        elements.append(current)
+        return elements
+
+    def _row_has_code(self, row):
```
mitya57: I think this should be named `_row_has_unpaired_backticks`, as this really has little to do with code.

markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
+        count_total_backtick = row.count('`')
+        count_escaped_backtick = row.count('\`')
+        count_backtick = count_total_backtick - count_escaped_backtick
+        ## odd number of backticks, 
+        ## we won't be able to build correct code blocks
+        return (count_backtick % 2 != 0)
```
mitya57: Just a suggestion: `return count_backtick & 1` is a more beautiful form of writing this.

markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
                 row = row[1:]
             if row.endswith('|'):
                 row = row[:-1]
-        return row.split('|')
+        return self._split(row, '|')
+
+    def _split(self, row, marker):
+        """ split a row of text with some code into a list of cells. """
+        if self._row_has_code(row):
+            ## fallback on old behaviour
```
mitya57: PEP 8 doesn't like comments starting with `##`, please use a single hash mark (`#`) instead (here and in multiple places below).

markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
+        current = ''
+        for i in row:
+            if i == '`' and (current == '' or current[-1] != '\\'):
+                ## Found an un-escaped backtick
+                escaped = not escaped
+
+            if i == marker and not escaped:
+                elements.append(current)
+                current = ''
+            else:
```
mitya57: Trailing whitespace.

markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
+            ## This should never happen
+            return row.split(marker)
+        elements.append(current)
+        return elements
+
+    def _row_has_code(self, row):
+        count_total_backtick = row.count('`')
+        count_escaped_backtick = row.count('\`')
+        count_backtick = count_total_backtick - count_escaped_backtick
+        ## odd number of backticks,
```
mitya57: Again `##` and a trailing whitespace.

mitya57:
In general the idea looks good to me :+1:, though I have left some comments on the code style issues.

Travis fails on Coverage check on Python 3.2 because Coverage doesn't seem to support that version, should we maybe exclude that check from the Travis config?

waylan:
Interesting. What about when multiple backticks are used to delimit a codespan? Might want to add to the test some more cells:

``````
`` some | code ``
``` some | code ```
```` some | code ````
`` some ` | ` code ``
``` some ` | ` code ```
```` some ` | ` code ````
``````

Note that each of the first three should result in the following HTML:

```
<code>some | code</code>
```

And the last three:

```
 <code>some ` | ` code</code>
```

File: markdown/extensions/tables.py [hunk 6]
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
             if row.endswith('|'):
                 row = row[:-1]
-        return row.split('|')
+        return self._split(row, '|')
+
+    def _split(self, row, marker):
+        """ split a row of text with some code into a list of cells. """
+        if self._row_has_code(row):
+            ## fallback on old behaviour
+            return row.split(marker)
```
waylan, line 99: According to the [coverage report](https://coveralls.io/builds/3919721/source?filename=markdown%2Fextensions%2Ftables.py), this is never tested. Might want to add a test for it.

markdown/extensions/tables.py
MustafaHaddara: On this line -- I've noted there we should never end up inside the `if` block, is there a preferred exception that should be thrown instead of silently falling back on the (potentially wrong) `string.split()`?

mitya57: No, we should not raise any exceptions, even on invalid syntax.

markdown/extensions/tables.py
```diff
@@ -85,7 +85,39 @@ def _split_row(self, row, border):
...
+                escaped = not escaped
+
+            if i == marker and not escaped:
+                elements.append(current)
+                current = ''
+            else: 
+                current += i
+        if escaped:
+            ## This should never happen
+            return row.split(marker)
```
waylan: According to the [coverage report](https://coveralls.io/builds/3919721/source?filename=markdown%2Fextensions%2Ftables.py), this is also never tested. Given the comment that this should never happen, then could you add a comment telling coverage to skip this test? -- `# pragma: no cover`

waylan:
> Travis fails on Coverage check on Python 3.2 because Coverage doesn't seem to support that version, should we maybe exclude that check from the Travis config?

Yes, Coverage 4.0 does not support  Python 3.2. We could require Coverage 3.x, but I don't really think it matters much at this point.

waylan:
@MustafaHaddara, @mitya57 already pointed out the specific issues to you, but the [Flake8 test](https://travis-ci.org/waylan/Python-Markdown/jobs/86843658) shows all those style/formatting related issues for you as well.

MustafaHaddara:
@waylan @mitya57 I've made the style fixes and updates to the test...but attempting to use multiple backticks to denote a code block is still broken.

waylan:
@MustafaHaddara this looks like a good start. When the multiple backticks issue gets worked out, either @mitya57 or I will merge it.  Thanks for the report and getting the work started on this.

markdown/extensions/tables.py
```diff
@@ -85,7 +85,38 @@ def _split_row(self, row, border):
...
+        elements = []
+        current = ''
+        for i in row:
+            if i == '`' and (current == '' or current[-1] != '\\'):
+                # Found an un-escaped backtick
+                escaped = not escaped
+            if i == marker and not escaped:
+                elements.append(current)
+                current = ''
+            else:
```
waylan: The [Flake8 test](https://travis-ci.org/waylan/Python-Markdown/jobs/86851063) is indicating this has trailing whitespace.

MustafaHaddara:
@waylan @mitya57 updated to deal with the issue of multiple backticks. I changed the implementation strategy from the manual string parsing I initially started with and instead, I use `inlinepatterns.BacktickPattern` to find inline code blocks inside of a table cell.

markdown/extensions/tables.py
```diff
@@ -85,7 +90,60 @@ def _split_row(self, row, border):
...
+                        if letter not in [' ', '\t', '\n']:
+                            # if we added a code element in this cell, add to that element's text
+                            elements[-1].text += letter
+                    else:
+                        current += letter
+                else:
+                    # match.groups()[-1] is the rest of the (unmatched) text
+                    # jump pointer to the beginning of the rest of the text
+                    i = row.index(match.groups()[-1])
+                    # build the code element
```
waylan: Please don't do this. Lets keep all inline parsing in the inline parsing step of the parser. If you want to use the regex for codespans to identify them, I understand, but when the table blockparser is done, the cell should still contain Markdown formatted inline text. If you don't do it that way, then other (inline) extensions could fail in unpredictable ways on table cells with codespans.

MustafaHaddara:
@waylan @mitya57 fixed as requested. All tests pass for me. Speaking of tests, I didn't notice there was already some tables tests in `/tests/extensions/extra/` so I had created a new tables test in `/tests/extensions/`. Do you want me to remove the test file I added and move those test cases to the older file? 

And would you like me to rebase and squash all of these commits into one commit to encompass the entire patch?

waylan:
The tables extension is part of extra, so yes I would prefer the tests be under `tests/extensions/extra/`. And yes, it would make sense to squash all of the commits.

markdown/extensions/tables.py
```diff
@@ -85,7 +90,49 @@ def _split_row(self, row, border):
...
+                escaped = False
+            else:
+                match = backtick_pattern.getCompiledRegExp().match(row[i:])
+                if not match:
+                    current += letter
+                else:
+                    groups = match.groups()
+                    delim = groups[1]  # the code block delimeter (ie 1 or more backticks)
+                    row_contents = groups[2]  # the text contained inside the code block
+                    i = row.index(groups[3])  # jump pointer to the beginning of the rest of the text
```
waylan: Any reason why you're not using `match.end(group)` here?

Every thing else looks good to me.

MustafaHaddara:
@waylan In response to your point about `match.end(group)`, that gives us the end of the group -- and we only have groups for the first code block delimiter, the text inside the code block, and the text after the code block. Asking for the end of a group won't help. 

What we really want is the beginning of the text outside the code block, ie. `match.start(group)`.

Please review, and if it is good with you I'll rebase into one commit.

waylan:
Looks good. Clean up the Flake8 errors and rebase and I'll merge.

MustafaHaddara:
The Flake8 test didn't run, and I'm not sure why. It threw `ERROR: InvocationError: '/home/travis/build/waylan/Python-Markdown/.tox/flake8/bin/flake8 /home/travis/build/waylan/Python-Markdown/markdown /home/travis/build/waylan/Python-Markdown/tests /home/travis/build/waylan/Python-Markdown/setup.py /home/travis/build/waylan/Python-Markdown/run-tests.py'` 

Is there a more detailed log I can look at somewhere?

waylan:
The [Travis Report](https://travis-ci.org/waylan/Python-Markdown/builds/87263477) (linked below) shows the details for each build. Just select the failing build and it will give you a full log. As per our earlier discussion, you can ignore the Python 3.2 build due to a version incompatibility with Coverage.

MustafaHaddara:
Rebased, all ready to go.

Issue #334, opened 2014-08-21
CodeHilite: expand option support


mingos777:
I would like to use Python Markdown in conjunction with CodeHilite plugin + Pygments to highlight some PHP code snippets, often one-liners. Currently, CodeHilite doesn't allow any lexer configuration (only some formatter configuration).

Pygments' default PHP lexer has the option `startinline`, which is set to `False` by default. This means that if I want any PHP code highlighted, I must start it with `<?php`. This is not very pretty (again: think about one-liners) and I would rather have Pygments instructed to switch this setting to `True`.

Currently it is not possible to pass in any lexer options to CodeHilite. It would be nice to be able to add lexer configuration to the lexer name, e.g.:

```
:::php(startinline=True)
```

waylan:
@mingos777, thanks for the report. I wasn't aware of Pygments behavior here.

I suppose Pygments is expecting (X)HTML with `<?php ...>` interspersed within it. The `startinline` (strange name for it btw) keyword tells it we have PHP code only. Is this keyword used by any other lexers?

My recollection is that in the past Pygments had multiple lexers for PHP, one which was for PHP only and another for PHP embedded in (X)HTML. That way, you could just specify the different lexer - no need to configure it. But is has been awhile. Is that no longer the case?

I hate to add a confiig option that is only useful for a single lexer, but I can see how this could be useful. What does the PHP lexer do if `startinline=False` and (X)HTML with `<?php ...>` embedded in it? In other words, could we always have `startinline=False` for the PHP lexer? if so, could we pass that into all Lexers (without error even if they don't know about it)?

And for the record, I just checked and the behavior is the same with the [fenced code](https://pythonhosted.org/Markdown/extensions/fenced_code_blocks.html) extension as it is with CodeHilite on indented code blocks.

mitya57:
@waylan The PHP lexer is not the only Pygments lexer that supports configs, most other lexers do.

Some time ago I have written a [patch for Docutils](http://sourceforge.net/p/docutils/patches/112) to add configs support, you can probably reuse that code

mingos777:
@waylan As far as I'm aware, only the (pure) PHP lexer uses this option. Options shared by all lexers are `stripnl` (strip  leading/trailing newlines), `stripall` (trim the input string), `ensurenl` (make sure the input ends with a newline), `tabsize` (self-explanatory and IMO very useful) and `encoding`. Several lexers have an extra set of options - this is the case for PHP.

[...]

PR #445, opened 2015-11-26
Make it possible to supply a filename for highlighted code

atodorov:
This patch makes it possible to supply a filename (or any arbitrary string for that matter) after the shebang line and it will be added to the generated code block. 

You can see it in action here (search for settings.py):
http://atodorov.org/blog/2013/03/07/python-twitter-django-social-auth-hello-new-user/

I've inherited pygments.HtmlFormatter in order to get it working quickly but maybe we'd want this to go straight into pygments. Let me know how you want to proceed. The CSS that goes with this change is as follows:

```
.codehilite .filename {
    width: 100%;
    display: block;
    font-size: 15px;
    color: #333;
    background-color: #F2F8FA;
    border: solid #CCC 1px;
    border-bottom: none;
    text-align: center;
}
```

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -23,7 +23,20 @@ from ..treeprocessors import Treeprocessor
 try:
     from pygments import highlight
     from pygments.lexers import get_lexer_by_name, guess_lexer
-    from pygments.formatters import get_formatter_by_name
+    from pygments.formatters import HtmlFormatter
+
+    class TitleHtmlFormatter(HtmlFormatter):
+        def __init__(self, **options):
+            HtmlFormatter.__init__(self, **options)
+            self.filename = options.get('filename', '')
+
+        def _wrap_pre(self, inner):
+            if self.filename:
+                yield 0, ('<span class="filename">' + self.filename + '</span>')
+
+            for tup in HtmlFormatter._wrap_pre(self, inner):
+                yield tup
+
     pygments = True
 except ImportError:
     pygments = False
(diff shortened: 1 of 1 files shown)


waylan:
Thank  you for your contribution. However, I'm unsure about this. I never understood the desire to pull the code blocks in from external pages. Why not just include the code in the document? Regardless, what is clear is that if such a feature were to be added, it should be part of #334  (see [this comment](https://github.com/waylan/Python-Markdown/issues/334#issuecomment-77912812)). And that won't happen until version 3.0 is released; when all extensions will be broken out of the main lib and maintained separately (see [the roadmap](https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap)). In fact, we are in a feature freeze until the 3.0 release anyway. Therefore, I am closing this at this time. Feel free to bring this up when the extension gets its own repo.

atodorov:
Hi @waylan ,
I think you've misunderstood me. I'm not proposing a parameter which will cause codehilite to read the source from an external file on disk, although I can see some people having a use for it. 

What I'm proposing is an additional text string, which can be used to indicate the file name (or anything really) being displayed. It's like a title for the code block. 

Btw I'd prefer if you leave this issue open and I will take care of it once the extensions are broken out into a separate repository.

Just for reference, the Pygments part is here:
https://bitbucket.org/birkenfeld/pygments-main/pull-requests/527

waylan:
Oh you want a title. Completely missed that. Also missed the fact that this requires upstream support. Given the later, this is definitely a no. No need to open a new issue (or reopen this one) as #334 should provide a way to accept any arbitrary keyword which is accepted by Pygments. In other words, Markdown should no longer need to have each keyword hardcoded for it to work.

Issue #217, opened 2013-05-19
WikiLinkExtension custom definition of wiki words regular expression


mblayman:
In using the WikiWordExtension, I discovered that only certain characters are permitted. I am trying to customize the extension to accept slashes as part of the definition, but I found that the regular expression defining what it means to be a wiki word is hardcoded.

It would be nice if user's could supply their own custom regular expression defining what the wiki word should be. As is, I had to copy the whole extension to tweak the regular expression to suit my needs.

Perhaps 'wikiword_pattern' in the configs?

waylan:
Personally, I dislike wikilinks and don't believe they belong in markdown. The existing extension is more a proof of concept to show what can be done. If you are subclassing and/or forking the extension to add your own behavior, then the extension is serving its purpose. For that reason and to avoid feature creep, I'm not interested in adding new features and/or maintaining new features.

I realize others disagree with me and that's fine. My suggestion would be to fork the extension and maintain your own as a third-party-extension. Feel free to add a link to the the [wikipage](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) so others can find it. In fact, if a decent alternative becomes available, I'm inclined to remove the existing extension completely.

mblayman:
Ok, I understand. I did have to copy the extension to make it work. If my project ever takes off and people are clamouring for the wikilinks extension, then I can pull it out into its own thing. Otherwise, I'll just go with what I've got.

PR #455, opened 2016-02-29
modified: markdown/extensions/wikilinks.py

luxcem:
Adds unrestricted regex for title because the current regex doesn't
allow the large variety of titles on wikipedia.

Adds the labeled links syntax from wikipedia : [[a|b]] is labeled "b"
on this page but links to page "a".


waylan:
I would consider this a duplicate of the request made in #217. My response is the same here as is was there.

luxcem:
Ok thanks for the review :)

PR #457, opened 2016-03-03
Nesting <pre> in <code> blocks instead

ParthKolekar:
When I make a markdown message in the form of 4-space indent. 

```
    <4 spaces>test
```

The output is shown in the form of 

```
    <pre><code>test</code></pre>
```

This breaks some tagging in the wikimedia editor which I am using. (Quite an old version.). It interprets the nested code  tag as `<code>`

Simply inverting this should make it work again.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -243,10 +243,10 @@ class CodeBlockProcessor(BlockProcessor):
             )
         else:
             # This is a new codeblock. Create the elements and insert text.
-            pre = util.etree.SubElement(parent, 'pre')
-            code = util.etree.SubElement(pre, 'code')
+            code = util.etree.SubElement(parent, 'pre')
+            pre = util.etree.SubElement(code, 'code')
             block, theRest = self.detab(block)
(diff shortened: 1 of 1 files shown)


waylan:
Are you suggesting that the output should be:

```
 <code><pre>test</pre></code>
```

If so, that is not valid Markdown output as defined in the [rules](http://daringfireball.net/projects/markdown/syntax#precode). For that matter it is not valid HTML as per the [HTML5 Spec](https://www.w3.org/TR/html5/text-level-semantics.html#the-code-element) either. Perhaps I am misunderstanding what you are trying to accomplish here.

waylan:
I just took another look at your change and you don't actually change anything except the variable names. It is not clear how that makes any difference. Nor do I understand what you are trying to accomplish. Could you please provide the input, the desired output and the actual output.

ParthKolekar:
Yes. I completely misinterpreted the code and missed the point of changing variable names and tag type.  If it is not a valid HTML5 spec, then I think I should try to hack about in the old wikimedia entry rather than here. 

This PR is then invalid.

PR #459, opened 2016-03-11
Make admonition classname and classname-title configurable

pbertera:
Making classname and classname-title configurable will make the usage of admonition more easy with mkdocs and templates with class names already defined for admonitions.

### DIFF
--- a/markdown/extensions/admonition.py
+++ b/markdown/extensions/admonition.py
@@ -39,9 +39,12 @@ class AdmonitionExtension(Extension):
 
 class AdmonitionProcessor(BlockProcessor):
 
-    CLASSNAME = 'admonition'
-    CLASSNAME_TITLE = 'admonition-title'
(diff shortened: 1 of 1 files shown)


waylan:
I'm not saying configurable its bad, but when do you actually need this? The class names are copied directly from ReST and are pretty standardized in my experience. Seems to me encouraging others to maintain that standard (in their templates, etc) is the better route.

Regardless, this project is currently in a feature freeze as we move towards version 3.0. In version 3.0 all extensions will be removed and distributed as separate packages. I'll leave this open for now, but I'm not intending on doing anything with it until the extensions are broken out. That being the case, you might be better off forking it as long as you do so within the confines of the license.

pazz:
+1 from me!
One would need this for instance when using markdown through pelican, using a theme such as bootstrap which pre-defines nice css for warning boxes etc.

PR #461, opened 2016-03-17
Add failing toc test for double inline code

minddust:
noticed this strange behavior here:

http://www.django-rest-framework.org/topics/3.0-announcement/#the-data-and-query_params-properties

so here is a failing testcase.

**data**

``` markdown
# This is `code` and `this` too.
```

**genetated**

``` html
<h1 id="this-is-code-and-this-too"><code>this</code> too.<a class="toclink" href="#this-is-code-and-this-too">This is <code>code</code> and </a></h1>
```

**expected**

``` html
<h1 id="this-is-code-and-this-too"><a class="toclink" href="#this-is-code-and-this-too">This is <code>code</code> and <code>this</code> too.</a></h1>
```

### DIFF
--- a/tests/test_extensions.py
+++ b/tests/test_extensions.py
@@ -775,8 +775,23 @@ class TestTOC(TestCaseWithAssertStartsWith):
             '<h2 id="header-2"><a class="toclink" href="#header-2">Header <em>2</em></a></h2>'
         )
 
-    def testAnchorLinkWithInlineCode(self):
-        """ Test TOC Anchorlink with inline code. """
+    def testAnchorLinkWithSingleInlineCode(self):
+        """ Test TOC Anchorlink with single inline code. """
+        md = markdown.Markdown(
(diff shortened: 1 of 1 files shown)


waylan:
Thanks for the report (and clear test case). I'm curious if this problem persists if you use the `permalink` config option rather than `anchorlink` (as documented [here](https://pythonhosted.org/Markdown/extensions/toc.html#usage)).

minddust:
`permalink` is looking good

minddust:
single inline code is also ok. 

2+ inline code + anchor makes the trouble

mitya57:
I have pushed a [fix for this](https://github.com/waylan/Python-Markdown/commit/803ab63609d6a148b01d6afc12f841501a5fd84c) to [`toc_bug`](https://github.com/waylan/Python-Markdown/tree/toc_bug) branch. If @waylan is OK with the fix, I can merge it to master.

waylan:
I suspected it was something simple like that. @mitya57 your fix makes sense to me.

PR #469, opened 2016-04-11
Support CodeHilite option use_pygments in fenced_code

martinmo:
Using `markdown.extensions.fenced_code` in combination with `CodeHiliteExtension(use_pygments=False)` does not yield the expected behavior. Pygments is used anyway because the corresponding config item is not copied. This PR adds a test to reproduce this bug and the corresponding fix.


waylan:
Thanks. I was about to say just don't use CodeHilite and you get the desired behavior (probably why it hasn't come up before), but a user should be able to use both is they desire. This is clearly a bug.

martinmo:
Thanks for merging!

But let me add a small note. There is a subtle difference if one disables the codehilite extension completely: the CSS class for the `<code>` tags then won't have the `language-` prefix (unfortunately, as documented). Most of the Javascript highlighters I have tried out there depend on that. To avoid writing some hacky post processing step prefixing the classes, I enabled codehilite and stumbled across this small issue :-)

PR #473, opened 2016-05-03
return converted markdown from markdownFromFile()

cyriac:
markdownFromFile() function currently doesn't return the converted markdown. Adding return to make the converted markdown available to the caller.

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -524,6 +524,6 @@ def markdownFromFile(*args, **kwargs):
                       DeprecationWarning)
 
     md = Markdown(**kwargs)
-    md.convertFile(kwargs.get('input', None),
-                   kwargs.get('output', None),
-                   kwargs.get('encoding', None))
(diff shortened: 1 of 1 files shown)


waylan:
> markdownFromFile() function currently doesn't return the converted markdown.

...which is by design. In fact, the function already writes the converted Markdown to the output file. Why should it output to two locations at once? As the documentation [states](https://pythonhosted.org/Markdown/reference.html#markdownFromFile):

> If this rather naive solution does not meet your specific needs, it is suggested that you write your own code to handle your ... needs.

Besides, your change still doesn't return the converted HTML. It returns the instance of the Markdown class, which does not retain the output.

Regardless, given that this was never the intended use case for this function and in all the years you are the first to ever request such a change, I'm going to say no.

cyriac:
Yes, I overlooked the entire thing. My changes don't make sense based on your notes.

PR #477, opened 2016-05-28
Fixed regex pattern for attr_list extension

semiversus:
The given regex pattern is matching "{: ... }" and also "{ ... }" (without the colon). The documentation is not mentioning this possibility and I had problems in inline elements where I use "{ ... }". I think the pattern "{:" is wisely chosen to prevent such a false detection.

### DIFF
--- a/markdown/extensions/attr_list.py
+++ b/markdown/extensions/attr_list.py
@@ -72,7 +72,7 @@ def isheader(elem):
 
 class AttrListTreeprocessor(Treeprocessor):
 
-    BASE_RE = r'\{\:?([^\}]*)\}'
+    BASE_RE = r'\{\:([^\}]*)\}'
     HEADER_RE = re.compile(r'[ ]+%s[ ]*$' % BASE_RE)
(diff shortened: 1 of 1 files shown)


semiversus:
Ok, may be not that good idea... There are unit tests checking for working "{ ... }" . But why?

waylan:
I forget the specific reason now, but if memory serves me right, there are other implementations which use both, so we accept both to be compatible with as many others as possible.

semiversus:
Is it a good idea to add a parameter to the attr_list extension for specifying a strict mode?
Like `attr_list(strict=True)`. In this mode only '{: ... }' is recognized. I'm willing to do the changes of implementation, unit test and documentation. At least the '{ ... }' pattern should be mentioned in documentation.

waylan:
> Is it a good idea to add a parameter to the attr_list extension for specifying a strict mode?
> Like `attr_list(strict=True)`. 

Probably, however, given we are in a feature freeze right now, such a feature would not be added until [3.0](https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap), when each extension will be broken out into its own separate package.

PR #480, opened 2016-06-06
edited setup docs for CodeHilite with Pygments

krother:
Improved the setup documentation when rendering Python code with Pygments and CodeHilite. Added a link to the Pygments CSS files and emphasized that not installing pygments does not cause an error message.

### DIFF
--- a/docs/extensions/code_hilite.txt
+++ b/docs/extensions/code_hilite.txt
@@ -20,20 +20,27 @@ This extension is included in the standard Markdown library.
 Setup
 -----
 
+### Step 1: download and install Pygments ###
+
 You will also need to [download][dl] and install the Pygments package on your
-`PYTHONPATH`. You will need to determine the appropriate CSS classes and create
-appropriate rules for them, which are either defined in or linked from the
-header of your HTML templates. See the excellent [documentation][] for more
+`PYTHONPATH`. The CodeHilite extension will produce HTML output without Pygments, 
+but it won't highlight anything.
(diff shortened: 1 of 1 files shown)


mitya57:
I don't like the idea of linking to third-party CSS. Instead, one can get the CSS using Pygments API:

from pygments.formatters import HtmlFormatter
HtmlFormatter().get_style_defs('.codehilite')

krother:
I agree - that would introduce another dependency. I would find it useful though to include the above sniplet in the documentation for the extension.

waylan:
I have been aware of and intending to link to the third party CSS for some time now. Those CSS files have been generated from calling the Pygments API. But I think @mitya57's concern is valid. I would prefer to leave the note in place and link to the CSS files from there. As a note, it is more like an "aside" and not part of the main body. Or maybe it should be in a "see also" style note.

In any event, the current documentation is lacking. Back when I first drafted it, Pygments documentation was different than it is today and we linked to instructions on how to generate the CSS from Pygment's API. Today, I don't even know where Pygments documents that. Even if we find it, what's to stop them from moving it again. We should just document it ourselves.

Thanks for kickstarting this @krother.

krother:
Let me know how I can help with this. I am using Python and Markdown daily for writing things. Your library is a key component that glues my blog server together.

waylan:
I made some additional changes in f3f7687. That should address any concerns raised here. Thanks again.

PR #486, opened 2016-08-05
Add css_class config to extensions/tables.py

buaabyl:
Add a config option 'css_class' to tables.py.

### DIFF
--- a/markdown/extensions/tables.py
+++ b/markdown/extensions/tables.py
@@ -55,6 +55,7 @@ class TableProcessor(BlockProcessor):
                 align.append(None)
         # Build table
         table = etree.SubElement(parent, 'table')
+        table.set('class', self.config['css_class'])
         thead = etree.SubElement(table, 'thead')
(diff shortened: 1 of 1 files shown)


waylan:
Thanks for your submission, however, we are currently in a feature freeze as we work toward version 3.

That said, I still have a few concerns. For starters, the tests were not updated and are failing. Second, the default behavior should remain as it was before this change (do not add a class unless the user specifically provides one). And third, I'm having a hard time seeing the value.  As all tables in a document will all receive the same class, you can just select all `table` elements (from CSS or JavaScript). And if you have other tables in your page template (not from the Markdown document), just wrap the Markdown output in a container and select all child `table` elements of that container. Without a clear use-case, I'm inclined to reject this.

buaabyl:
because pygment hightlight using table, and have the same level of other tables, so the wrap not help. I use markdown to write user manual. Tables have different style from highlight table. So I add a custom class to table.

PR #487, opened 2016-08-05
Add 'inlcude' extension, syntax is {!file.md!}, original from Christo…

buaabyl:
original version from cmacmackin@gmail.com,
I add pygment highlight to it.
Means than {!test.c!} can highlight.

### DIFF
--- a/markdown/extensions/include.py
+++ b/markdown/extensions/include.py
@@ -1,0 +1,189 @@
+#!/usr/bin/env python
+# -*- coding: utf-8 -*-
+#
+#  include.py
+#
+#  Copyright 2015 Christopher MacMackin <cmacmackin@gmail.com>
+#  Modify by William <buaabyl@gmail.com>
+#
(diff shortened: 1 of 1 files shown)


waylan:
Thanks, but I have no interest in maintaining this extension. There is no reason why it can't be a third party extension which is maintained by someone who actually wants such a feature. I would suggest you package it up as a Python package (as explained in the [tutorial](https://github.com/waylan/Python-Markdown/wiki/Tutorial:-Writing-Extensions-for-Python-Markdown)) and list it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) for others to find.

buaabyl:
Thanks, by the way, can you set up a list of  exrensions, because finding a extension is not that easy, I have to write by myself sometime...

waylan:
I linked to the list on the wiki in my previous comment. And that list is publicly editable so anyone can add their own items to the list.

PR #488, opened 2016-08-11
remove unnecessary footnotes postprocessor

Larivact:
The current implementation using these placeholders seems very weird. Am I missing something?

### DIFF
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -24,8 +24,6 @@ from ..util import etree, text_type
 from ..odict import OrderedDict
 import re
 
-FN_BACKLINK_TEXT = "zz1337820767766393qq"
-NBSP_PLACEHOLDER = "qq3936677670287331zz"
 DEF_RE = re.compile(r'[ ]{0,3}\[\^([^\]]*)\]:\s*(.*)')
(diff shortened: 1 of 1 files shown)


Larivact:
@waylan What do you think?

waylan:
Thanks for your work on this. I haven't had an opportunity to look at this closely. I wouldn't be surprised if some old code was around that was unnecessary in the footnotes extension. That extension has existed since before I was involved in the project and is one of the few pieces of code I haven't given much attention to (my changes have always been the minimum required to keep it working).

Regardless, as we move to version 3.0 I was expecting to do a complete rewrite of this extension and am not eager to start making changes before then. In fact, we are actually in a feature freeze right now (while working toward 3.0) and I am only interested in fixing bugs in the 2.x branch. So unless this fixes a clearly defined bug, I'll wait till I start refactoring this extension for 3.0. If you're interested, you can review the [3.0 Roadmap](https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap) on the wiki.

waylan:
I don't consider this to be "unnessecary". Note that inline parsing can muck up HTML entities. Therefore, it is less likely to break things if we use placeholders. See the discussion to #493 where one of those placeholders was added. If the tests pass after removing the placeholders, then that is more likely an indication that we have incomplete tests than anything else.

PR #489, opened 2016-08-15
Fix table alignment when seperator contains spaces

painyeph:
Currently table alignment is correct for seperator like `----:|:----:|:----`, but incorrect for `----: | :----: | :----`, which is commonly used.


waylan:
@painyeph good catch. Thanks. It appears that our test output was wrong and now that you fixed the bug, the test needs to be updated. If you could include the fix in your PR, I'll merge it in. You can update the test by running the following command.:

```
./run-tests.py update tests/extensions/extra/tables.txt
```

That will update the output file `tests/extensions/extra/tables.html`. Just push that change.

waylan:
It would appear that you did not use the tool to update the test. Due to an annoying minor difference in (otherwise insignificant) whitepsace, the test is still failing. You can either delete the newline at the end of the file or use the tool to generate the file.

waylan:
Excellent. Thanks.

PR #490, opened 2016-08-25
Add ability to specify custom Formatter for Pygments code highlighting

dprothero:
In our use of the Python-Markdown library, we discovered the need to be able to use a custom Pygments Formatter class. Pygments supports creating custom formatters, but there wasn't a way to have the `CodeHiliteExtension` utlize them.

I've added `pygments_formatter` as a config option for the `CodeHiliteExtension` class (which is also used by `FencedCodeExtension`). It can take a string, which will be passed to the call to Pygments' `get_formatter_by_name` function. It defaults to 'html', which is what was hard-coded before.

Also, the option can take a reference to a class. This makes it convenient to specify a custom class in code without having to create a separate package for your custom formatter. Usage is something like this:

```
def get_html_from_markdown(value):
    # Use the extra extensions
    from code_samples.code_formatters import MarkdownHtmlFormatter
    return markdown.markdown(value, extensions=[
        'markdown.extensions.extra',
        TocExtension(anchorlink=True),
        CodeHiliteExtension(pygments_formatter=MarkdownHtmlFormatter)])
```

The last line of code demonstrates the usage.

All tests pass (aside from those marked as Skip), including the additional test I added. Let me know if you'd like any modifications.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -17,6 +17,7 @@ License: [BSD](http://www.opensource.org/licenses/bsd-license.php)
 
 from __future__ import absolute_import
 from __future__ import unicode_literals
+import inspect
 from . import Extension
 from ..treeprocessors import Treeprocessor
 
@@ -75,7 +76,8 @@ class CodeHilite(object):
 
     def __init__(self, src=None, linenums=None, guess_lang=True,
                  css_class="codehilite", lang=None, style='default',
-                 noclasses=False, tab_length=4, hl_lines=None, use_pygments=True):
+                 noclasses=False, tab_length=4, hl_lines=None, use_pygments=True,
+                 pygments_formatter='html'):
         self.src = src
         self.lang = lang
         self.linenums = linenums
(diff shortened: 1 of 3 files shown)


atbaker:
I work with @dprothero. As front end design evolves, I'm sure we're not the only ones who could benefit from specifying custom formatters.

Thanks for your maintenance of the project and for giving this PR a look!

waylan:
Cool! I like it. And I would merge this if we weren't in a feature freeze right now. We are only accepting bug fixes at this time. As we move toward version 3.0 (see the [roadmap](https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap)) the extensions will be broken out into separate packages and will no longer be tied to the development/release cycle of the main package. At that time these sorts of things won't hold up a good contribution like this. Unfortunately, until then, this will have to wait. Of course, in the meantime, you are welcome to fork the extension as long as you honor the licence.

waylan:
Oh, and before you ask, the timeline is "when it's done."

dprothero:
That sounds totally fine to me! We'll just keep my fork in our requirements.txt until this all comes together.

PR #500, opened 2016-09-28
replace `getiterator` function for Python 3.6

### LINKED ISSUE #499: Fix for Python 3.6
I'm trying to upgrade my project to Python 3.6 so that, even though it's in beta, when it does get released to its full version, it'll be a simple pull request to update python versions. But currently, installing Markdown gives me:

```
      File "<string>", line 1, in <module>
      File "/tmp/pip-build-hiv4fz31/Markdown/setup.py", line 270, in <module>
        'Topic :: Text Processing :: Markup :: HTML'
      File "/usr/local/lib/python3.6/distutils/core.py", line 148, in setup
        dist.run_commands()
      File "/usr/local/lib/python3.6/distutils/dist.py", line 955, in run_commands
        self.run_command(cmd)
      File "/usr/local/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/usr/local/lib/python3.6/site-packages/setuptools/command/install.py", line 61, in run
        return orig.install.run(self)
      File "/usr/local/lib/python3.6/distutils/command/install.py", line 545, in run
        self.run_command('build')
      File "/usr/local/lib/python3.6/distutils/cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "/usr/local/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/usr/local/lib/python3.6/distutils/command/build.py", line 135, in run
        self.run_command(cmd_name)
      File "/usr/local/lib/python3.6/distutils/cmd.py", line 313, in run_command
        self.distribution.run_command(command)
      File "/usr/local/lib/python3.6/distutils/dist.py", line 974, in run_command
        cmd_obj.run()
      File "/tmp/pip-build-hiv4fz31/Markdown/setup.py", line 184, in run
        out = template % self._get_context(src, outfile)
      File "/tmp/pip-build-hiv4fz31/Markdown/setup.py", line 116, in _get_context
        c['body'] = self.md.convert(src)
      File "build/lib/markdown/__init__.py", line 375, in convert
        newRoot = treeprocessor.run(root)
      File "build/lib/markdown/treeprocessors.py", line 361, in run
        brs = root.getiterator('br')
    SystemError: Python/getargs.c:1508: bad argument to internal function
```

Is there a quick fix to this solution already, or is this a new bug with 3.6?

brandonchinn178:
Fixes #499 

The `getiterator` function doesn't accept arguments in Python 3.6 (raises a TypeError), so this manually filters elements based on the tag name

### DIFF
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -358,14 +358,14 @@ class PrettifyTreeprocessor(Treeprocessor):
         self._prettifyETree(root)
         # Do <br />'s seperately as they are often in the middle of
         # inline content and missed by _prettifyETree.
-        brs = filter(lambda e: e.tag == 'br', root.getiterator())
+        brs = root.iter('br')
         for br in brs:
             if not br.tail or not br.tail.strip():
                 br.tail = '\n'
             else:
                 br.tail = '\n%s' % br.tail
         # Clean up extra empty lines at end of code blocks.
-        pres = filter(lambda e: e.tag == 'pre', root.getiterator())
+        pres = root.iter('pre')
         for pre in pres:
             if len(pre) and pre[0].tag == 'code':
                 pre[0].text = util.AtomicString(pre[0].text.rstrip() + '\n')
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -358,14 +358,14 @@ class PrettifyTreeprocessor(Treeprocessor):
         self._prettifyETree(root)
         # Do <br />'s seperately as they are often in the middle of
         # inline content and missed by _prettifyETree.
-        brs = root.getiterator('br')
+        brs = filter(lambda e: e.tag == 'br', root.getiterator())
         for br in brs:
             if not br.tail or not br.tail.strip():
                 br.tail = '\n'
             else:
                 br.tail = '\n%s' % br.tail
         # Clean up extra empty lines at end of code blocks.
-        pres = root.getiterator('pre')
+        pres = filter(lambda e: e.tag == 'pre', root.getiterator())
         for pre in pres:
             if len(pre) and pre[0].tag == 'code':
                 pre[0].text = util.AtomicString(pre[0].text.rstrip() + '\n')


waylan:
Thanks for doing the work on this. I don't doubt your right, but did you find any comments about this change in Python's documentation/release notes/commit logs? I'm curious about the reasoning behind that change.

brandonchinn178:
There are no explicit changes in the [changelog](https://docs.python.org/3.6/whatsnew/changelog.html#changelog), so I suspect there must've been some change in the way Python 3.6 interfaces with the underlying C library?

mitya57:
The [documentation for 3.6](https://docs.python.org/3.6/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getiterator) still mentions `tag` as a valid argument.

According to the stacktraces this looks like a bug in Python 3.6 to me:

```
>>> from xml.etree.ElementTree import Element
>>> el = Element('foo')
>>> el.getiterator('bar')
SystemError: ../Python/getargs.c:1508: bad argument to internal function
>>> el.getiterator(tag='bar')
TypeError: iter() takes at most 1 argument (140172072006928 given)
```

Please report it to Python issue tracker (or I can do it for you, if you prefer).

brandonchinn178:
Wait so `.iter('div')` works... since `getiterator` is being deprecated anyway, I think this is a worthwhile change. It seems like maybe there's a bug in `getiterator` when it passes the call to `iter`?

waylan:
Ah, `getiterator` has been [deprecated](https://docs.python.org/2.7/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element.getiterator) since Python 2.7,  and `iter` was added in 2.7. As 2.7 is the lowest supported version, we should be using `iter`.

mitya57:
Getting rid of deprecated calls is indeed the good thing.
But I filed http://bugs.python.org/issue28314 for the regression anyway.

PR #503, opened 2016-10-13
Release as a universal wheel

adamchainz:
By releasing as a [Python wheel](http://pythonwheels.com/) as well as a source distribution, you can speed up end user’s installs. After merging this command, to release you just need to run `python setup.py clean sdist bdist_wheel upload`.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,1 +1,4 @@
+[bdist_wheel]
+universal = 1
+
 [nosetests]


waylan:
Thanks for your contribution. This is planned for version 3.0. Version 2 is currently in maintenance mode and only accepting bug fixes at this time. I'll leave this open to remind us to do this for version 3.

adamchainz:
Cool thanks

benjaoming:
This change doesn't affect the codebase, but I think `setup.py` looks fairly advanced and would require special testing on at least Windows in order to see that the .whl can still be installed. Python-Markdown doesn't use `data_files` for  `setup()` which would definitely break.

@waylan I don't think you should push it all the way to 3.0 but maybe introduce it in the 2.7 series?

waylan:
@benjaoming there will be no 2.7 series. Which is why I'm rejecting all changes which are more than simple bugfixes. All such work should be on the 3.0 branch (see the [Roadmap][1] for details).

[1]: https://github.com/waylan/Python-Markdown/wiki/3.0-Roadmap

benjaoming:
> @benjaoming there will be no 2.7 series.

Ah okay, well that would have been another amendment to the [release notes](https://pythonhosted.org/Markdown/release-2.6.html#safe_mode-deprecated) ;)

jdufresne:
You could include the license file in the generated wheel using the `license_file` in `setup.cfg`:

```ini
[metadata]
license_file = LICENSE.md
```

See the documentation for information on this feature of wheel:

https://wheel.readthedocs.io/en/stable/#including-the-license-in-the-generated-wheel-file

This would satisfy the license's requirement:

>  Redistributions of source code must retain the above copyright notice, this list of conditions and the following disclaimer.

waylan:
This was addressed in #606.

PR #504, opened 2016-10-21
Allow restricting the list of attributes able to be set with attr_list

puiterwijk:
Signed-off-by: Patrick Uiterwijk patrick@puiterwijk.org

### DIFF
--- a/docs/extensions/attr_list.txt
+++ b/docs/extensions/attr_list.txt
@@ -85,4 +85,7 @@ Usage
 See [Extensions](index.html) for general extension usage, specify `markdown.extensions.attr_list`
 as the name of the extension.
 
-This extension does not accept any special configuration options.
+The following options are provided to configure the output:
+
+* **`allowed_attributes`**:
+    List of attributes to allow setting. `['*']` means all. Defaults to `['*']`.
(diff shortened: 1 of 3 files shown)


waylan:
Thanks for your contribution. It hadn't occurred to me to do this. Regardless, I'm not sure its a good idea (or that it fits with our goals). 

For example, some attributes are only valid on some HTML elements, but not others. We don't do anything to address that. Perhaps a user wants to allow a certain attribute on some elements, but not others. 

Additionally, you didn't include an `exclude` option. What if a user wants to include everything except one attribute. With the current approach, she has to list every single possible attribute except that one. An `exclude` option with one item listed would be much preferable.

All of the above sounds an awful lot like an HTML sanitizer. A tool which has a very valid use case. However, such tools already exist and can be run on the HTML output by Markdown. As it stands, the Markdown parser cannot guarantee that its output is safe (see this [excellent explanation](https://michelf.ca/blog/2010/markdown-and-xss/)). In fact, in the upcoming version 3.0, the current `safe_mode` will no longer be supported. If you want "safe" output, you need to use an HTML sanitizer. I recommend [Bleach](https://bleach.readthedocs.io/en/latest/) with [bleach whitelist](https://github.com/yourcelf/bleach-whitelist) as a good place to start for a set of rules to pass to Bleach.

Given the above, and the current direction that Python-Markdown is heading, I don't think it makes sense to make this change. The same can be accomplished much more thoroughly via a HTML sanitizer.

waylan:
As an aside, as long as you conform to the license, you are always welcome to fork the built-in extension and maintain your own which offers whatever features you desire.

PR #508, opened 2016-11-03
Enable tables without headers

pfertyk:
Ruby's `kramdown` allows for creating tables without headers, I think Python's `markdown` should not be worse.

I know that this might not be a part of the standard, but I also think that this is a very useful feature (I need it and I don't want to switch from Pelican to Jekyll just for that).

Please let me know if the code quality is not good enough, I will do my best to improve it.

### DIFF
--- a/markdown/extensions/tables.py
+++ b/markdown/extensions/tables.py
@@ -28,20 +28,40 @@ class TableProcessor(BlockProcessor):
 
     def test(self, parent, block):
         rows = block.split('\n')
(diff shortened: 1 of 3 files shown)


waylan:
Thanks for the submission. However I will not be merging this pull request. Please see my response to issue #74 for a full explanation. As you appear to have a working extension, you are free to fork it as long as you conform to the license. You could even release it as a third party extension and list it on the [wiki](https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions) for others to find.

PR #510, opened 2016-11-11
Replace homegrown OrderedDict with purpose-built Registry.

### LINKED ISSUE #418: Replace OrderedDict with prioritized List?
I'm looking for feedback on a possible Extension API change which might be introduced in version 3.0.

Currently (version 2.x), all extensions register where they are run within the parser with our homegrown OrderedDict. Each piece of code is assigned a name in the dict and an extension inserts itself before or after a given name. 

``` python
patterns.add(SomePattern(), '<emphasis')  # insert before "emphasis" pattern
del patterns['emphasis']                  # remove "emphasis" pattern
```

What I am proposing is that instead of an Ordered Dict, we use a list (as we did in version 1.x). However, each item in the list is assigned a "priority" attribute which is used to sort the list in order.  For example, each inlinepattern class would have a priority set (10, 15, 20, 25, 30, ...). Higher numbers get run first. An extension could set a priority of 22 to get placed between items with priorities of 20 & 25. If a second extension needed to also be between 20 & 25 but before the extension with priority 22, is could use priority 23 or 24, and we don't have the possible conflicts that exist now.

The tricky part would be in removing existing patterns. It is easy to do with the named keys. It might be a little more tricky without. And we can't hardcode index position as that can be changed by other extensions. The entire list would need to be searched for the given class instance. Do we set a "name" property on each class for this reason, rely only on the "priority" property, or something else? Perhaps the built-ins could all be assigned to constants. That way, the constant (a text-based name) could be used for reference purposes, but the value would be the integer (much like the logging modules error codes). Or maybe the constants could point to the class instances themselves.

Therefore, where `patterns` is the list of patterns, one might alter the list like this:

``` python
patterns.register(SomePattern(), priority=23)
patterns.deregister(inlinepatterns.EMPHASIS)
```

I'm using register/deregister as opposed to register/unregister for the reasons stated [here](http://english.stackexchange.com/questions/25931/unregister-vs-deregister) (although it could change by popular demand). However, what I'm not sure about is the best way to define the priority:

This is odd as the priority is a function of the registration process, not creation of the class instance:

``` python
myinstance = MyPattern(priority=23)
patterns.register(myinstance)
```

This is easy to understand but then requires the parser to monkeypatch the class instance to attach the priority to the class:

``` python
myinstance = MyPattern()
patterns.register(myinstance, priority=23)
```

In the first example, `register` is simply an alias to `list.append`. However, the second example would require something like this:

``` python
class PriorityList(list):
    def register(self, item, priority):
        item.priority = priority  # the monkeypatch
        self.append(item)
```

Of course, regardless of implementation details, we would only sort once after all extensions are loaded.

**WHY?**
1. Provides more flexibility to extension authors. Multiple third-party-extensions would be less likely to conflict with each other. 
   
   For example, in the current situation, Extension A removes "emphasis" and extension B tries to insert before emphasis (`<emphasis`). If the user lists the extensions `[B, A]`, everything works fine, but if she lists them as `[A, B]`, then a `KeyError` will be raised when setting up extension B. 
   
   Or two extensions might use the same name. For example multiple "math" extensions currently exist but each does something slightly different. A user could conceivably try to use two of them together, but one might replace the other.
2. Gets rid of the awkward `<emphasis` syntax. Ugh.
3. Removes the homegrown (and mostly untested) `OrderedDict` (the implementation that ships with Python only allows adding to the end so it is useless for this purpose). The fact that the new implementation is just a sorted list is an implementation detail that does not even need to be mentioned in the docs. Extension authors only need to know about and use the two methods `register` and `deregister` on a registry.

waylan:
All processors and patterns now get "registered" to a Registry.
Each item is given a name (string) and a priority. The name is for
later reference and the priority can be either an integer or float
and is used to sort. A Registry instance is a list-like iterable with
the items auto-sorted by priority. If two items have the same priority,
then they are listed in the order there were "registered". Registering
a new item with the same name as an already registered item replaces
the old item with the new item (however, the new item is sorted by
its priority). To remove an item, "deregister" it by name or index.

Fixes #418.


waylan:
Well, Travis is getting some weird failure that I can't replicate locally. And as far as I can tell, it should be working properly, unless there is some sort of circular import issue. But why can't I replicate then?

mitya57:
I also cannot reproduce this error locally. There are however other tests failures caused by Pygments…

waylan:
Yeah, the code highlighting tests depend on a specific version of Pygments. However if Pygments is not installed the testing pass with the "no Pygments" code path. Tox is configured to not install Pygments so the tests should pass if run in tox-although I had to force tox to recreate the env.

waylan:
Closed in favor of #688.

PR #515, opened 2016-12-21
Add a start_level option

estan:
The `start_level` will be added to the level of all headings.

This is useful if the markdown rendering is done in some context where the content to be generated is already at some level.

### DIFF
--- a/docs/reference.txt
+++ b/docs/reference.txt
@@ -249,6 +249,20 @@ The following options are available on the `markdown.markdown` function:
           <li>Pears</li>
         </ol>
 
+* __`start_level`__{: #start_level }: Start level for headings in the output.
+  Default: 0
+
+    Given the following headers:
+
+        # Header 1
+
+        Header 2
+        --------
+
+    By default markdown will turn Header 1 and Header 2 into `h1` and `h2`
+    elements, respectively. The `start_level` option specifies an offset that
+    will be added to the levels. For example, with a `start_level` of 2,
+    Header 1 would result in a `h3` element and Header 2 in a `h4` element.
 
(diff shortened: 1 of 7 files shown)


docs/reference.txt
```diff
@@ -249,6 +249,20 @@ The following options are available on the `markdown.markdown` function:
...
+
+        # Header 1
+
+        Header 2
+        --------
+
+    By default markdown will turn Header 1 and Header 2 into `h1` and `h2`
+    elements, respectively. The `start_level` option specifies an offset that
+    will be added to the levels. For example, with a `start_level` of 2,
+    Header 1 would result in a `h3` element and Header 2 in a `h5` element.
```
mitya57: You probably mean `h3` and `h4` (not `h3` and `h5`).

estan: Bah, indeed. Good catch. Will fix.

estan: Pushed fixed commit.

waylan:
Thanks for the submission, however, this should be implemented as an extension. In fact, it is already part of the [TOC][0] extension (as well as the deprecated [HeaderID][1] extension). Note that in the upcoming version 3.0, all keywords will be removed and the only way to alter the parser's behavior will be via extensions. That being the case, I see no need to do this. if you really want this behavior outside of the existing extensions, then I would suggest implementing your own third-party extension.

[0]: https://pythonhosted.org/Markdown/extensions/toc.html
[1]: https://pythonhosted.org/Markdown/extensions/header_id.html

estan:
Aha, I see. I didn't know options were being removed. I'll try to make my own extension then.

Issue #449, opened 2015-12-19
Another table and inline code issue


facelessuser:
Inline code in tables can't escape backtics like they can outside of tables.

```
``Test` | passes``

| Table | Test     |
| ----- | ------- |
| Broken | ``Test` | fails`` |
```

You can see it only accepts a surrounding backtic of 1.  Then it splits on the pipe.  The rest is lost because the table is only 2 columns wide.

``` html
<p><code>Test` | passes</code></p>
<table>
<thead>
<tr>
<th>Table</th>
<th>Test</th>
</tr>
</thead>
<tbody>
<tr>
<td>Broken</td>
<td><code>`Test</code></td>
</tr>
</tbody>
</table>
```

I do have intentions of looking into the reason why and providing a fix when I get time.

facelessuser:
I guess I don't under stand this design.  Maybe someone can fill in the gaps. 

So this doesn't work (a real world example I am trying to use -- just showing some regex):

| Doesn't work | Works |
| --- | --- |
| ```[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^_`{|}~]``` | ```[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^`_`{|}~]```. |

This is extremely limited as you can only show grave accents in a table's inline code block if all accents have a pair.  Now I am sure there was some reasoning here, I just don't understand it.  What is more, I figured that I could probably put it in an html code block, but the html code block was split as well.

I am not sure why we alter the behavior in tables.  I feel like it should capture just as it would outside of tables.  I understand we are treating backticks as a special case in tables, but I don't understand the additional requirement of only if all grave accents are paired.  This is jarring for a user as the rules have clearly changed inside a table.  I feel like the row should be processed with a normal code pattern and then split.  We are already parsing code patterns special in tables, we might as well do it proper.

Before I put in work, I want to know if allowing normal backtick logic would be rejected; I don't want to waste time.  As it stands, I am simply forced to throw out my entire table and do raw html tables as Python Markdown can't handle an odd number of grave accents in a table, and I am afraid a pull to fix this might be rejected.

waylan:
@facelessuser, yes, normal markdown inline markup should work in tables just the same as it does outside of tables.

[...]

Issue #521, opened 2017-01-10
FYI: Spelling silently failing in Travis


facelessuser:
Just an FYI. Aspell is failing.  The dictionary is missing. I just went through this myself.  All I needed to do was also install aspell-en in apt.

facelessuser:
If I get a chance, I'll issue a pull soon since this is such an easy patch.

PR #522, opened 2017-01-10
Fix Issue #449 and Issue #521

facelessuser:
- At  some point the logic of counting backtics and determining if they
are odd or even was used to parse a row's text into cells.
Unfortunately this approach broke expected code parsing logic in a
table.  We essentially traded one bug for another.  This fixes table
backtick handling and restores sane backtick logic while preserving
existing fixes. (#449)

- Travis should checkout aspell-en (#521)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ addons:
     packages:
     - libtidy-0.99-0
     - aspell
+    - aspell-en
 install:
   - pip install tox
   - pip install coveralls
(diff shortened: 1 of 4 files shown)


facelessuser:
Sorry for the multiple commits.  I had forgotten that inline code is evaluated before escapes.  With squash and merge I might be okay, but if you need a rebase anyways, let me know.

waylan:
I'd prefer the different things to be in different commits. If you could rebase that would be preferred.

facelessuser:
Will do.  I kind of figured, but I was hoping :).

facelessuser:
Close for rebase.

facelessuser:
New pull requests are:

- https://github.com/waylan/Python-Markdown/pull/524
- https://github.com/waylan/Python-Markdown/pull/523

PR #524, opened 2017-01-11
Better handling of backticks

facelessuser:
At some point the logic of counting backticks and determining if they
are odd or even was used to parse a row's text into cells.
Unfortunately this approach broke expected code parsing logic in a
table. We essentially traded one bug for another. This fixes table
backtick handling and restores sane backtick logic while preserving
existing fixes. (issue #449)

### DIFF
--- a/tests/extensions/extra/tables.txt
+++ b/tests/extensions/extra/tables.txt
@@ -83,7 +83,7 @@ Lists are not tables
 
 Add tests for issue #449
 
-Odd backtics | Even backtics
+Odd backticks | Even backticks
 ------------ | -------------
 ``[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^_`{|}~]`` | ``[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^`_`{|}~]``
 
--- a/tests/extensions/extra/tables.txt
+++ b/tests/extensions/extra/tables.txt
@@ -80,3 +80,13 @@ Lists are not tables
 
  - this | should | not
  - be | a | table
+
+Add tests for issue #449
+
+Odd backtics | Even backtics
+------------ | -------------
+``[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^_`{|}~]`` | ``[!\"\#$%&'()*+,\-./:;<=>?@\[\\\]^`_`{|}~]``
+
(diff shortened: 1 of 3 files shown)


tests/extensions/extra/tables.txt
```diff
@@ -80,3 +80,13 @@ Lists are not tables
 
  - this | should | not
  - be | a | table
+
+Add tests for issue #449
+
+Odd backtics | Even backtics
```
mitya57: Backtic**k**s (also in the commit message).

facelessuser: I can update the test, but I don't think there is any thing I can do about the commit message though.

mitya57: Why? `git commit --amend` && `git push --force`.

facelessuser: Wasn't sure how that affects the pull as I think GitHub uses the message to calculate the hash.  Good to know it can be done.  I'll take care of it.

facelessuser: Done.  Commit update worked great, thanks for letting me know.  I was under the incorrect assumption that GitHub didn't support it. Wish I had know about this sooner :).

mitya57: Just a note for your future PRs: you could also squash the two commits into a single one by using `git commit -a --amend` (or `git add` && `git commit --amend`). Sorry for not being so precise in my previous reply.

mitya57: Though @waylan probably uses GitHub UI for merging pull requests, and that UI has an option to squash all commits and edit the message, anyway :)

facelessuser: > Just a note for your future PRs: you could also squash the two commits into a single one by using git commit -a --amend (or git add && git commit --amend). Sorry for not being so precise in my previous reply.

Oh.  That makes sense.  I'm clearly not a git command line warrior; I use the GitHub app to manage my workflow (and it is very limited compared to other git GUIs).  I should really play with this more and work it into my workflow.  Getting out of my comfort zone and learning git better could really help.

> Though @waylan probably uses GitHub UI for merging pull requests, and that UI has an option to squash all commits and edit the message, anyway :)

I hope so.  Hoping to avoid another rebase :).

PR #525, opened 2017-01-17
Support for <main> tag in raw HTML to suppress an extra <p> tag.

daniel-j-mac:
When using the [HTML5 `<main>` tag](https://www.w3.org/WebPlatform/docs/html51/grouping-content.html#the-main-element) as raw HTML in my Markdown file, an extra `<p>` tag was output.

For example:
`<main>...some HTML</main>`

turned into:
`<p><main>...some HTML</main></p>`

The `<p>` tag ended up creating extra space above the main content.

I added "main" to the BLOCK_LEVEL_ELEMENTS constant in markdown/util.py and this seemed to work well.  The `<main>` tag is recognized and the extra `<p>` tag is suppressed.

Might this be useful for others?

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -32,7 +32,7 @@ BLOCK_LEVEL_ELEMENTS = re.compile(
     "|hr|hr/|style|li|dt|dd|thead|tbody"
     "|tr|th|td|section|footer|header|group|figure"
     "|figcaption|aside|article|canvas|output"
-    "|progress|video|nav)$",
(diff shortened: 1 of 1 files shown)


waylan:
Can you point to the `<main>` tag in the HTML spec somewhere? Last I checked, we support all of the block level tags in the spec. I'm not interested in adding any arbitrary tag.

waylan:
Opps, I see you already linked to the spec. Sorry. Not sure how I missed the `<main>` tag before.

PR #527, opened 2017-01-18
codehilite: detect languages with funny names

grantisu:
Extend the language regex to match things like 'C#' and 'VB.Net'.


waylan:
Not technically necessary as the relevant languages have aliases which work, but I suppose we might as well support them all.

facelessuser:
Probably should update fenced extension as well...for consistency.

Issue #526, opened 2017-01-17
Question: Approach to be taken for a pull


facelessuser:
@waylan I had a question about an approach I wanted to run by you, and your general opinion on which direction you'd like me to take for a pull.

Recently you accepted a pull to fix table backticks.  In that pull, I hadn't realized that Python Markdown doesn't currently allow for the escape of `|` in `Tables`, but the logic I added accounts for it. [PHP Markdown Extra](https://michelf.ca/projects/php-markdown/extra/#backslash) **does** allow this (which makes sense based on the sytax `Tables` adds).  So in general, I would call this a **bug**, but before I continue, I want to make sure that this is something you want to address, or if you consider this a feature/enhancement that you would prefer not to introduce into the 2.6 series.  In that case, I would just issue a pull to remove the logic that accounts for `\|` and be done.

If you do  consider the handling of `|` to be a bug. The fix is pretty easy, but I wanted to ask about the approach.

1. I could just inject an escape pattern for `|` and call it done.

2. We could append `|` to the current escape list and let the current escape pattern handle it.  The reason why I like this is that an extension can just see what's already in that list and avoid adding redundant escapes that have already been added by a previous extension.  The only issue with this is that a global list is used in the Markdown class.  It would be nice if we copied the global array for that instance of Markdown instead so that you could append and not modify the global.  I think regardless of whether it is a condoned approach to append to that list, it should probably not be a global copy to protect against people's monkey patching.  At the very least make it a tuple so people would have to physically construct a new one if they tried to modify it.

In short:

1. Do you want to handle escaped `|` for Extra Tables?

2. If so, which approach would you prefer?

Based on your response, I will take the appropriate action.

waylan:
My intention has always been to have extensions add to the global escape list on the Markdown instance. My thinking has been that if an extension introduces a new character with special meaning, then users should be able to escape that character in the same manner as any other escapable character. And the way to do that is in the global escape list. Yes, that means that you can also escape the character outside of a table, but that is the way escape works for the existing characters anyway. I consider this to be the least surprising behavior for end users.

As an aside, I realize that many of the existing extensions have not done this. That has been an annoyance of mine and part of why I moved the escape list to the class. I just never went through the extensions and made sure they all make use of it.

[...]

PR #528, opened 2017-01-18
Tables should add the escaping of pipes

facelessuser:
The table extension should have escaped pipes in order to adhere to the
PHP Markdown extra implementation.  In order to properly add escaped
pipes, a bug relating to persisting appended escapes also had to be
patched. Reference: #526

### DIFF
--- a/markdown/extensions/tables.py
+++ b/markdown/extensions/tables.py
@@ -160,6 +160,8 @@ class TableExtension(Extension):
 
     def extendMarkdown(self, md, md_globals):
         """ Add an instance of TableProcessor to BlockParser. """
+        if '|' not in md.ESCAPED_CHARS:
+            md.ESCAPED_CHARS.append('|')
         md.parser.blockprocessors.add('table',
                                       TableProcessor(md.parser),
                                       '<hashheader')
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -75,9 +75,6 @@ class Markdown(object):
         'xhtml5': to_xhtml_string,
     }
 
-    ESCAPED_CHARS = ['\\', '`', '*', '_', '{', '}', '[', ']',
-                     '(', ')', '>', '#', '+', '-', '.', '!']
-
     def __init__(self, *args, **kwargs):
         """
         Creates a new Markdown instance.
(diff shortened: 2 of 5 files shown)


facelessuser:
I'm going to need to adjust border striping to account for escaped chars and the header test.  Let me rebase and finish it proper.

markdown/extensions/tables.py
```diff
@@ -160,6 +160,8 @@ class TableExtension(Extension):
 
     def extendMarkdown(self, md, md_globals):
         """ Add an instance of TableProcessor to BlockParser. """
+        if '|' not in md.ESCAPED_CHARS:
```
waylan, line 163: This makes me wonder if `ESCAPED_CHARS` should be a set. Then it will always only contain one of each item and extensions don't need to check before appending. Just an idea.

facelessuser: Great idea.  That would be even better :).

waylan:
@facelessuser there is no need to close a PR. Just add a comment that it is not ready. You can just rebase locally and then do a 'force push' to the same remote branch and GitHub will update the PR properly.

waylan:
Oh, I just noticed you committed to your master branch. If you use a separate unique branch for a PR, then rebasing etc, becomes easier.

facelessuser:
> Oh, I just noticed you committed to your master branch. If you use a separate unique branch for a PR, then rebasing etc, becomes easier.

Yeah, I need to get out of that habit.

facelessuser:
I have it working.  I'll get the pull in this evening.

I'll also incorporate changing the escaped list to a set...are we worried about breakage on the change to a set (append vs add)? If you aren't, I have no problem.  I don't know who actually uses the escaped char list.  I know I do, but I'll prepare for it before the next Python Markdown release.

facelessuser:
If we care about preventing breakage on the migration from list to set (at least in the 2.6 series), we could just derive from set and add the append/extend options as that is probably all anyone ~~would~~ should ever use.  If we don't care about breakage in this area, then it doesn't really matter.  We could also just stick to a list for 2.6.

For 3.0 though, I think a set should be used.

I'll add the set if I don't hear otherwise tonight when I create a new pull.

waylan:
> are we worried about breakage on the change to a set (append vs add)?

I forgot about that. Let's leave it a list for now. I agree, a set makes sense for 3.0.

Issue #468, opened 2016-04-09
markdown.extensions.footnotes -- multiple references to the same footnote generate duplicate HTML IDs


a3nm:
Consider the following minimal Markdown document which contains two references to the same footnote:

``` markdown
[^a][^a]

[^a]: b
```

I compile it to HTML with Python-Markdown with the following code:

``` python
#!/usr/bin/python3

import markdown
markdown.markdownFromFile("minimal.md", "minimal.html",
        extensions=['markdown.extensions.footnotes'])
```

The result is the following:

``` html
<p><sup id="fnref:a"><a class="footnote-ref" href="#fn:a" rel="footnote">1</a></sup><sup id="fnref:a"><a class="footnote-ref" href="#fn:a" rel="footnote">1</ a></sup></p>
<div class="footnote">
<hr />
<ol>
<li id="fn:a">
<p>b&#160;<a class="footnote-backref" href="#fnref:a" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

This HTML is invalid (and will be rejected by validators) because the ID "fnref:a" is multiply defined.

The footnotes extensions has a parameter `UNIQUE_IDS` which looks related (though the description is not very clear), but it does not appear to work for this purpose. If I set it to true as follows:

``` python
#!/usr/bin/python3

import markdown
markdown.markdownFromFile("minimal.md", "minimal_param.html",
        extensions=['markdown.extensions.footnotes(UNIQUE_IDS=True)'])
```

I obtain the following output:

``` html
<p><sup id="fnref:2-a"><a class="footnote-ref" href="#fn:2-a" rel="footnote">1</a></sup><sup id="fnref:2-a"><a class="footnote-ref" href="#fn:2-a" rel="footnote">1</a></sup></p>
<div class="footnote">
<hr />
<ol>
<li id="fn:2-a">
<p>b&#160;<a class="footnote-backref" href="#fnref:2-a" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

This time, the duplicate ID is "fnref:2-a", but the problem remains.

I think Python-Markdown should be generating different HTML IDs for the different invocations of the same footnote, with the backlink pointing, e.g., to the first one, or (ideally) offering backlinks to all of them as Wikipedia does (see https://en.wikipedia.org/wiki/Help:Footnotes#Footnotes:_using_a_source_more_than_once).

[...]

PR #534, opened 2017-01-23
Create additional references for duplicate footnotes

facelessuser:
Track when we find duplicate footnote references and create unique ids
for them.  Then add an additional tree-processor after inline to go
back and update the footnotes with additional back references that link
to the duplicate footnote references. Ref: #468.

### DIFF
--- a/tests/extensions/extra/footnote.html
+++ b/tests/extensions/extra/footnote.html
@@ -1,5 +1,8 @@
 <p>This is the body with a footnote<sup id="fnref:1"><a class="footnote-ref" href="#fn:1" rel="footnote">1</a></sup> or two<sup id="fnref:2"><a class="footnote-ref" href="#fn:2" rel="footnote">2</a></sup> or more<sup id="fnref:3"><a class="footnote-ref" href="#fn:3" rel="footnote">3</a></sup> <sup id="fnref:4"><a class="footnote-ref" href="#fn:4" rel="footnote">4</a></sup> <sup id="fnref:5"><a class="footnote-ref" href="#fn:5" rel="footnote">5</a></sup>.</p>
 <p>Also a reference that does not exist[^6].</p>
+<p>Duplicate<sup id="fnref:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> footnotes<sup id="fnref2:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> test<sup id="fnref3:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup>.</p>
+<p>Duplicate<sup id="fnref:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> footnotes<sup id="fnref2:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> test<sup id="fnref3:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup>.</p>
+<p>Single after duplicates<sup id="fnref:c"><a class="footnote-ref" href="#fn:c" rel="footnote">8</a></sup>.</p>
 <div class="footnote">
 <hr />
 <ol>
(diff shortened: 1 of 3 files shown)


tests/extensions/extra/footnote.html
```diff
@@ -1,5 +1,8 @@
 <p>This is the body with a footnote<sup id="fnref:1"><a class="footnote-ref" href="#fn:1" rel="footnote">1</a></sup> or two<sup id="fnref:2"><a class="footnote-ref" href="#fn:2" rel="footnote">2</a></sup> or more<sup id="fnref:3"><a class="footnote-ref" href="#fn:3" rel="footnote">3</a></sup> <sup id="fnref:4"><a class="footnote-ref" href="#fn:4" rel="footnote">4</a></sup> <sup id="fnref:5"><a class="footnote-ref" href="#fn:5" rel="footnote">5</a></sup>.</p>
 <p>Also a reference that does not exist[^6].</p>
+<p>Duplicate<sup id="fnref:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> footnotes<sup id="fnref2:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> test<sup id="fnref2:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup>.</p>
+<p>Duplicate<sup id="fnref:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> footnotes<sup id="fnref2:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> test<sup id="fnref2:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup>.</p>
```
waylan: Footnote 6 is repeated three times on the first line (once with `id="fnref:a"` and twice with `id="fnref2:a"`.

Footnote 7 is repeated three times on the second line (once with `id="fnref:a"` and twice with `id="fnref2:a"`.

However, in the actual footnotes below, we have three backreference links to each line line as:

footnote 6: `href="#fnref:a"`, footnote 7: `href="#fnref2:a"` and footnote 8: `href="#fnref3:a" `
footnote 9: `href="#fnref:a"`, footnote 10: `href="#fnref2:a"` and footnote 11: `href="#fnref3:a" `

We have two problems. 

1. The footnote numbers don't match between the references and the backrefs. If we "jump back to footnote 8", then we should jump back to a footnote labeled "8" (i.e.; `<sup>8</sup>`)

2. There appears to be a bug in generating ids for the duplicates when there are more than two. Curiously, the backrefs point to the correct ids, but the references (after the first) are all assigned `"fnref2:a"` rather than counting up (`"fnref2:a", "fnref3:a", ...`).

waylan (changes requested):

tests/extensions/extra/footnote.html
waylan: Regarding problem 1 above, because the three references all point to the same footnote, they all *should* be labeled  "6" (i.e.; `<sup>6</sup>`). So in this case, the bug is in the backrefs. All three should have identical tiles: "Jump back to footnote 6" (they do not), but point to different IDs (which they do).

facelessuser: > "fnref2:a" rather than counting up ("fnref2:a", "fnref3:a", ...).

I missed that.  I'll take a look.

waylan: So given the input in the test (repeated here in relevant part):

```md
Duplicate[^a] footnotes[^a] test[^a].
 
Duplicate[^b] footnotes[^b] test[^b].
 
Single after duplicates[^c].

[^a]: 1
[^b]: 2
[^c]: 3
```

I would expect the following output:

```html
<p>Duplicate<sup id="fnref:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> footnotes<sup id="fnref2:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup> test<sup id="fnref3:a"><a class="footnote-ref" href="#fn:a" rel="footnote">6</a></sup>.</p>
<p>Duplicate<sup id="fnref:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> footnotes<sup id="fnref2:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup> test<sup id="fnref3:b"><a class="footnote-ref" href="#fn:b" rel="footnote">7</a></sup>.</p>
<p>Single after duplicates<sup id="fnref:c"><a class="footnote-ref" href="#fn:c" rel="footnote">8</a></sup>.</p>

<div class="footnote">
<hr />
<ol>
<li id="fn:a">
<p>1&#160;<a class="footnote-backref" href="#fnref:a" rev="footnote" title="Jump back to footnote 6 in the text">&#8617;</a><a class="footnote-backref" href="#fnref2:a" rev="footnote" title="Jump back to footnote 6 in the text">&#8617;</a><a class="footnote-backref" href="#fnref3:a" rev="footnote" title="Jump back to footnote 6 in the text">&#8617;</a></p>
</li>
<li id="fn:b">
<p>2&#160;<a class="footnote-backref" href="#fnref:b" rev="footnote" title="Jump back to footnote 7 in the text">&#8617;</a><a class="footnote-backref" href="#fnref2:b" rev="footnote" title="Jump back to footnote 7 in the text">&#8617;</a><a class="footnote-backref" href="#fnref3:b" rev="footnote" title="Jump back to footnote 7 in the text">&#8617;</a></p>
</li>
<li id="fn:c">
<p>3&#160;<a class="footnote-backref" href="#fnref:c" rev="footnote" title="Jump back to footnote 8 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

facelessuser: > The footnote numbers don't match between the references and the backrefs. If we "jump back to footnote 8", then we should jump back to a footnote labeled "8" (i.e.; <sup>8</sup>)

I think I understand.  I'll have both these resolved this evening.

facelessuser: > I would expect the following output:

Excellent, that is what I was looking for.

I know what went wrong with issue 1 as I had iterated through multiple changes.  While I track how many duplicates I have, I accidentally remove my tracker for unique ids.

Issue 2 is even easier.  I just need to clone the first link and don't touch it's title and only update the link back.

facelessuser: These should be fixed now.  Just force pushed the changes.

waylan (approved):

Issue #452, opened 2016-01-28
Nested (and indented) raw HTML with blank lines breaks


waylan:
Simple example:

``` markdown
<div>
    <div>
        Div with a blank line

        in the middle.
    </div>
    <div>
        This gets treated as a code block
    </div>
</div>
```

Results:

``` html
<div>
    <div>
        Div with a blank line

        in the middle.
    </div>

<pre><code>&lt;div&gt;
    This gets treated as a code block
&lt;/div&gt;
</code></pre>
</div>
```

Removing the blank line in the first child `div` resolves the problem. But this should work without altering the HTML. This bug was first reported to [MkDocs](/mkdocs/mkdocs/issues/791).

Felk:
on 2.6.8 a very similar bug still occurs.

	>>> x = """<i>
	...     <i>
	...
	...     </i>
	... </i>"""
	>>> print(md.convert(x))
	<p><i>
		<i></p>
	<pre><code>&lt;/i&gt;
	</code></pre>
	<p></i></p>

Workaround is to either avoid empty lines in html, or don't indent with tabs or multiples of 4 spaces

facelessuser:
Isn't that an inline tag?

Felk:
It doesn't matter what tag it is. Even tables work. After the empty line the closing tag gets treated as a markdown code block.

EDIT: oh, it actually works properly with `<div>` for the example above. However, here's a slightly different example that still breaks, and I have no idea why (i only added a headline above):

	>>> x = """# headline
	... <div>
	...     <div>
	...
	...     </div>
	... </div>"""
	>>> print(md.convert(x))
	<h1>headline</h1>
	<p><div>
		<div></p>
	<pre><code>&lt;/div&gt;
	</code></pre>
	</div>

If there's an additional newline after the headline, it works again. In my case it's a nested table with some jinja2 directives that implicitly cause some newlines in the html if I don't explicitly suppress them, which I could do of course.

facelessuser:
> EDIT: oh, it actually works properly with <div> for the example above.

Yup, inline are handled differently.

> However, here's a slightly different example that still breaks, and I have no idea why (i only added a headline above):

If you put a space after the header, it works.  I'd have to look at the Markdown spec to see if this is a bug.

facelessuser:
[...]

PR #537, opened 2017-01-24
Fix HTML parse with empty lines

facelessuser:
If both open and close was not found in first block, additional blocks
were evaluated without context of previous blocks.  The algorithm needs
to evaluate a buffer with the left bracket present.  So feed in all
items and get the right bracket, then adjust the data_index to be
relative to the last block. Ref: #452

### DIFF
--- a/markdown/preprocessors.py
+++ b/markdown/preprocessors.py
@@ -258,7 +258,13 @@ class HtmlBlockPreprocessor(Preprocessor):
             else:
                 items.append(block)
 
-                right_tag, data_index = self._get_right_tag(left_tag, 0, block)
+                # Need to evaluate all items so we can calculate relative to the left index.
+                right_tag, data_index = self._get_right_tag(left_tag, left_index, ''.join(items))
+                # Adjust data_index: relative to items -> relative to last block
+                prev_block_length = 0
+                for item in items[:-1]:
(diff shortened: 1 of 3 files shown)


facelessuser:
Wasn't planning on fixing anything else, but I overlooked this as one that could be fixed pre 3.0.  I don't know about the HTML parsing algorithm on a whole (seems it takes some shortcuts), but I was at least able to identify why it couldn't resolve the right bracket across a gap.

facelessuser:
Personal opinion on this.  For 3.0, I would either rewrite this to not use recursion and also track a state of depth. So when you call it in subsequent blocks, it knows how many openings it has to resolve until done.  I don't recall if this is already on your roadmap or not.  This current algorithm gets slower the more blocks it has to process. For every block it doesn't find the end tag it is looking for, it has to process i + 1 blocks.  So if you had to process 3 blocks to find your close tag, you actually processed 1 + 2 + 3 = 6 blocks (re-iterating blocks you already processed).  I think this is a fine fix for now as that is how the algorithm is designed, but not ideal for the future.  I wanted to fix the current algorithm, but I wasn't willing at this time to tackle rewriting.  Maybe if I'm more ambitious in the future or you haven't gotten to it first :).

PR #559, opened 2017-04-10
Automatically register extension.

aeslaughter:
It doesn't seem necessary to require the extension to register itself when it can be done automatically. Calling the registration method manually makes it easy to forget (e.g., the meta extension) thus not allowing the list to be inspected by other extensions that may require an extension.

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -185,6 +185,8 @@ class Markdown(object):
                 ext = self.build_extension(ext, configs.get(ext, {}))
             if isinstance(ext, Extension):
                 ext.extendMarkdown(self, globals())
+                if ext not in self.registeredExtensions:
+                    self.registeredExtensions.append(ext)
                 logger.debug(
                     'Successfully loaded extension "%s.%s".'
                     % (ext.__class__.__module__, ext.__class__.__name__)


waylan:
The existing behavior was an intensional design decision. Not all extensions need to be registered and we don't force them to be. In fact for many extensions, there is absolutely no benefit to registering them. It is up to each extension developer to decide whether that extension needs to be registered.

aeslaughter:
Just because one developer doesn't need it to be registered doesn't mean another doesn't. For example, I would like to check for dependencies and it certainly doesn't hurt to register an extension. Personally, I think the inconsistency is problematic, but I will find another way.

waylan:
@aeslaughter when the extension registration feature was added, its sole envisioned purpose was to provide for the "reset" feature. As it turns out, it is useful for other purposes and may be due for reassessment. However, as the 2.x branch is in a feature freeze while we work on 3.0, this would be something to address in 3.0.

waylan:
We will not be changing this. An extension is "registered" if it needs to utilize the "reset" feature. and that would only ever be a decision made by the developer of the extension. If you are trying to determine if a certain extension is in use, then that should be done in another way as it is outside the scope of the "register" feature.

PR #579, opened 2017-09-07
Create smart_del.py

zzir:
The Smart_Del extension adds smarter handling of double `~` within words.

example:

```python
md_text = "How ~~is a~~ nice weather!"
md = markdown.Markdown(['extra', 'smart_del'])
html = md.convert(text)
print(html)
# <p>How <del>is a</del> nice weather!</p>
```

### DIFF
--- a/markdown/extensions/smart_del.py
+++ b/markdown/extensions/smart_del.py
@@ -1,0 +1,31 @@
+"""
+example:
+
+```python
+md_text = "How ~~is a~~ nice weather!"
+md = markdown.Markdown(['extra', 'smart_del'])
+html = md.convert(text)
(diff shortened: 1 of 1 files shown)


facelessuser:
From what I understand, Python Markdown has no intentions of adding new extensions.  Also, there are already plenty of extensions that offer this as a solution:

- http://facelessuser.github.io/pymdown-extensions/extensions/tilde/
- https://github.com/aleray/mdx_del_ins

You can always search the [wiki](https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions) to see listed 3rd party extensions, but there may even be more that are not listed.

PR #581, opened 2017-09-08
Make coveralls work again

mitya57:
Testing coveralls, please do not merge until it passes.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist = py27, py32, py33, py34, pypy, flake8, checkspelling
 
 [testenv]
+passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps = -rtest-requirements.txt
 commands = coverage run --source=markdown {toxinidir}/run-tests.py {posargs}
            coverage report --show-missing
--- a/tox.ini
+++ b/tox.ini
@@ -5,7 +5,7 @@ envlist = py27, py32, py33, py34, pypy, flake8, checkspelling
 passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps = -rtest-requirements.txt
 commands = coverage run --source=markdown {toxinidir}/run-tests.py {posargs}
-           coverage report --show-missing
+           coveralls
 
 [testenv:flake8]
 deps = flake8
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,8 +23,6 @@ install:
   - pip install coveralls
 script:
   - tox
-after_success:
-  - coveralls
 branches:
   only:
     - master
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,7 +20,6 @@ addons:
     - aspell-en
 install:
   - pip install tox
-  - pip install coveralls
 script:
   - tox
 branches:
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,6 @@
 nose
 coverage<4.0
+coveralls
 pyyaml
 pytidylib
 pygments


facelessuser:
I'm not really sure why Coveralls broke in the first place, but if it seems to be a PITA to get working again, maybe we should consider moving to Codecov?  Not sure if there are specific features we are trying to take advantage of in Coveralls or not, but I've found Codecov to be pretty nice and easy to use.

With that said, if there is indeed something superior to Coveralls, maybe it is worth the effort.

mitya57:
I will try it too, but I will return to this a bit later.

waylan:
No specific reason we use coveralls. I don't recall why it was chosen originally. Any working solution is fine by me. Just glad its getting addressed, especially now that I fixed Travis yesterday.

facelessuser:
Cool.  I think Coveralls was originally one of the leading solutions for a bit, so I imagine that was probably why.  Codecov has recently given it some good competition though, and in general I've grown to like its interface more.

mitya57:
I fixed coveralls. Moving to codecov is more complicated, because I do not have permission to add it to this repository.

Let’s merge this for now and move to codecov in a next pull request if needed.

waylan:
One thing: Can we avoid the comments being added to the PR each time coveralls runs? Its unnecessary noise IMO. The "check" like we have with Travis is sufficient.

facelessuser:
I agree with the message noise part.  I wish most of these guys would disable such things by default instead of enabling them by default.

It looks like it may be a repo maintainers job to disable notifications; I don't think there is a config that can be placed in the repo to do that: https://coveralls.zendesk.com/hc/en-us/articles/201774895-Notifications.

waylan:
Unfortunately, the documentation seems to be out-of-date. What I actually see is...

coveralls

That doesn't match up at all with the documentation. I can't find a "Notifications" option anywhere.

One nice thing about Codecov is that you can configure it with a `.codecov.yml` config file in the repo. Once its enabled, there's no need for the project owner to adjust settings.

facelessuser:
Yup, that's what I do with codecov as well. Just another reason to consider it in the long term.

mitya57:
Ok, if you don’t manage to disable comments, then please disable coveralls and enable codecov, I will try to add a configuration for it.

tox.ini
```diff
@@ -2,9 +2,10 @@
 envlist = py27, py32, py33, py34, pypy, flake8, checkspelling
 
 [testenv]
+passenv = TRAVIS TRAVIS_JOB_ID TRAVIS_BRANCH
 deps = -rtest-requirements.txt
 commands = coverage run --source=markdown {toxinidir}/run-tests.py {posargs}
-           coverage report --show-missing
+           coveralls
```
waylan, line 8: Not sure I'm a fan of this. I like looking at the coverage report in the tox logs. In fact, I generally run tox locally and will review the coverage report. With this change I can't do that anymore. And as coverage was run in a now deactivated tox venv, I can't just run `coverage report` manually either. Now I have to manually run `coverge run ...` and `coverage report`  to get coverage reports locally before pushing to GitHub. This is a non-starter for me.

waylan: Although, is there any reason why we can't leave the `coverage report ` line? Just call `coveralls` after?

facelessuser: I personally don't think coveralls should be run in the test suite, but should instead be run by Travis. If I want to run the test suite locally, it's going to try to run coveralls which doesn't make sense to me.

waylan: @facelessuser good call. Let's not run coveralls from tox. It we can't make it work directly from travis, then we need to switch to another provider. I'll try to find some time later this evening to enable codecov.

PR #582, opened 2017-09-08
Switch to codecov from coveralls

waylan:
Fixes #581.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,6 @@
 sudo: false
 language: python
+
 matrix:
   include:
     - python: '2.7'
@@ -12,19 +13,25 @@ matrix:
       env: TOXENV=pypy
     - env: TOXENV=flake8
     - env: TOXENV=checkspelling
+
 addons:
   apt:
     packages:
     - libtidy-0.99-0
     - aspell
     - aspell-en
+
 install:
   - pip install tox
-  - pip install coveralls
+  - pip install 'coverage<4.0'
+  - pip install codecov
+
 script:
   - tox
+
 after_success:
-  - coveralls
+  - codecov
+
 branches:
   only:
     - master
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,0 +1,1 @@
+comment: off
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 ===================
 
 [![Build Status](http://img.shields.io/travis/Python-Markdown/markdown.svg)](https://travis-ci.org/Python-Markdown/markdown)
-[![Coverage Status](https://img.shields.io/coveralls/Python-Markdown/markdown.svg)](https://coveralls.io/r/Python-Markdown/markdown?branch=master)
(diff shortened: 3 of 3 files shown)


waylan:
[codecov](https://codecov.io/gh/Python-Markdown/markdown)

Not sure why nothing is showing up in the checks below or on codecov.io.

facelessuser:
Doesn't appear that there is a repo added: https://codecov.io/gh/Python-Markdown

waylan:
Each time I  try to "add" the repo, it tells me the next step is to upload a report. Perhaps something is failing in Travis?

facelessuser:
Maybe it doesn't kick in on pulls until it has a base from master.

facelessuser:
Maybe because it's coming from waylan instead of a branch directly off PythonMarkdown....That's probably it.

facelessuser:
Hey, we got something :).

.travis.yml
```diff
@@ -12,19 +13,25 @@ matrix:
...
   apt:
     packages:
     - libtidy-0.99-0
     - aspell
     - aspell-en
+
 install:
   - pip install tox
-  - pip install coveralls
+  - pip install coverage<4.0
```
facelessuser: Is there a reason for this (<4.0)?  I'm using coverage 4.0+ on all my projects without issue.

waylan:
Yeah, I push a report from my local machine. Still failing from GitHub/Travis.

facelessuser:
Hmm.  It was passing before your last commit (except it would then fail with no base to compare to).

https://codecov.io/gh/Python-Markdown/markdown/tree/908c2252e0ecb6056eb328f225887c78eecc8adf/markdown

facelessuser:
I don't think you'll get a badge until it gets merged.

.travis.yml
waylan: See 20b17c29081. And the `.coverage` file format changed in 4.0. As the tox envs output the older format, we need the older version to upload as well.

facelessuser:
Oh, I think I get what you mean...you pushed the report from your machine which is why we have a report...

facelessuser:
Looks like that worked.  The "No report found to compare against" will go away once this is merged to master.  Though, why are we avoiding 4.0?

waylan:
It works!

waylan:
> why are we avoiding 4.0?

Because 4.0 dropped support for Python 3.2, which we still support (see 20b17c2). And the `.coverage` file format changed in 4.0. As the tox envs output the older format, we need the older version to upload as well.

I figured out that was the problem when I first tried to push a report from my local machine. Once I downgraded coverage, it worked.

.travis.yml
facelessuser: Interesting. I must not be covering 3.2... Or for some reason it doesn't cause an issue for me :/. Now I'm going to need to audit my repos.

facelessuser:
Yup, I don't formally support 3.2 on any of my repos, so that is why I can get away with 4.0+.  Thanks for the explanation.

Issue #584, opened 2017-10-02
IndexError when defining links in nested block elements


smathot:
When using nested block elements of the following kind:

~~~
<div markdown="1">

[link]: a_link

<div markdown="1">

</div>

</div>
~~~

An `IndexError` occurs:

~~~
doc-pelican (0.7)$ python3 test.py 
Markdown 2.6.9
Traceback (most recent call last):
  File "test.py", line 22, in <module>
    print(md.convert(text))
  File "/usr/lib/python3/dist-packages/markdown/__init__.py", line 371, in convert
    root = self.parser.parseDocument(self.lines).getroot()
  File "/usr/lib/python3/dist-packages/markdown/blockparser.py", line 65, in parseDocument
    self.parseChunk(self.root, '\n'.join(lines))
  File "/usr/lib/python3/dist-packages/markdown/blockparser.py", line 80, in parseChunk
    self.parseBlocks(parent, text.split('\n\n'))
  File "/usr/lib/python3/dist-packages/markdown/blockparser.py", line 98, in parseBlocks
    if processor.run(parent, blocks) is not False:
  File "/usr/lib/python3/dist-packages/markdown/extensions/extra.py", line 130, in run
    block = self._process_nests(element, block)
  File "/usr/lib/python3/dist-packages/markdown/extensions/extra.py", line 97, in _process_nests
    self.run(element, block[nest_index[-1][0]:nest_index[-1][1]],  # last
IndexError: list index out of range
~~~

This happens only under very specific circumstances (but reliably so). For example, adding an extra blank line below the link definition will fix the error.

Hope this is useful!

facelessuser:
Hmm, I may take a look at this in the next couple of days if I have a chance.  Issue doesn't look obvious to me.

waylan:
Hmm, there could be a number of things going on here. Some initial thoughts without verifying anything...

Do you have the `extra` extension enabled?

I also find it peculiar that the only content within the first `<div>` is a link reference. Of course, link references get removed from the document, so in the output the `<div>` would have no content. I realize this may be a minimum test case (which we appreciate receiving, thank you), but this specific scenario would make no sense in the real world. Does the error occur with other content or only a link reference?

Finally, the second `div` is nested within the first. <del>I don't recall off-hand, but it may not be necessary to use `markdown=1` in the nested `div`.</del> Does the error still occur if that is the case?

facelessuser:
[...]

PR #585, opened 2017-10-07
Fix raw html reference issue

facelessuser:
Preserve the line which a reference was on to prevent raw HTML indexing issue. Ref #584.

### DIFF
--- a/markdown/extensions/abbr.py
+++ b/markdown/extensions/abbr.py
@@ -53,6 +53,9 @@ class AbbrPreprocessor(Preprocessor):
                 title = m.group('title').strip()
                 self.markdown.inlinePatterns['abbr-%s' % abbr] = \
                     AbbrPattern(self._generate_pattern(abbr), title)
+                # Preserve the line to prevent raw HTML indexing issue.
+                # https://github.com/Python-Markdown/markdown/issues/584
+                new_text.append('')
             else:
                 new_text.append(line)
         return new_text
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -234,7 +234,7 @@ class FootnotePreprocessor(Preprocessor):
                 fn.insert(0, m.group(2))
                 i += _i-1  # skip past footnote
                 footnote = "\n".join(fn)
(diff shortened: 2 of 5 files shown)


facelessuser:
Just occurred to me that abbreviations and footnotes should be checked for this kind of breakage and fixed if it occurs.  I will verify and patch those if necessary in this pull as well.

facelessuser:
Abbreviations and footnotes are now handled. Footnotes were a little special as we had to preserve a line for each block. We also had to account for unnecessary trailing empty lines. Now that I say it out loud, I should get a test in for trailing empty lines as well....

facelessuser:
Looks like footnotes changed in the tests slightly. Ugh, forgot to run the tests one more time...

facelessuser:
Tests are now passing.

waylan:
This looks good. One question though: why not just modify the raw HTML processor so that this didn’t matter? To be clear, I haven’t looked into it myself. It just seems like that might have been the first approach I would have explored. Maybe there’s a good reason?

facelessuser:
> One question though: why not just modify the raw HTML processor so that this didn’t matter?

I am open to suggestions, but the reason this issue occurs is that the raw HTML preprocessor is not aware of references.  It parses the blocks and such first populating the tag_data.  **Then** the footnote, abbr, and link reference preprocessor gets run and then removes entire blocks from the preprocessed file.  Then the raw HTML parser references indexes it thought was good that are no longer good.

At least with abbr and link references, maybe you could post process the tag data after the references get stripped and rebuild them proper. With the footnote references, the extension actually utilizes the tag_data when constructing the footnote which adds even more complications.

Honestly, its a messy situations and this was the easiest way to solve the issue.  Maybe there is a "better" way to approach it, but this seemed like the least invasive approach.  Maybe if I spent more time getting to understand the raw HTML parser another cleaner (more involved) approach may make itself manifest.

Any ideas?

facelessuser:
I'm not in a hurry to get this merged. I can mull over this more and see if there is a better way. We can consider this a first draft. If we can't come up with something better, this may be decent stopgap solution.

Issue #590, opened 2017-10-18
Allow spaces in image urls


sedrubal:
Currently `markdown.markdown` does not recognize images with spaces in their url:

This is not working:
```py
>>> markdown.markdown('![](/test test.jpg)')
'<p>![](/test test.jpg)</p>'
```

The same without space is working:
```py
>>> markdown.markdown('![](/testtest.jpg)')
'<p><img alt="" src="/testtest.jpg" /></p>'
```

A possible workaround is to replace the space with `%20`, but this is ugly.

If you think this should work, I can try to fix it :wink:

facelessuser:
URLs/URIs don't allow spaces.  The link syntax expects a valid URL/URI etc.  This is pretty universal with all Markdown parsers I've used.  If I try to enter your example here in GitHub, it won't even accept it, not because the image isn't found, but because it is invalid syntax.

![](/test test.jpg)

waylan:
Actually, there is quite a bit of variety (see [Babelmark][1]), although most of the "old-school" implementations seem to agree. What I find most curious is that Python-Markdown has different behavior for links and images. Seems the behavior should be the same here. If we are changing anything, we should follow the lead of the "old-school" implementations (after all Python-Markdown fits that category).

[1]: http://johnmacfarlane.net/babelmark2/?normalize=1&text=!%5Bimage%5D(%2Ftest+test.jpg)%0A%0A%5Blink%5D(%2Ftest+test.html)

facelessuser:
Interesting.  I didn't realize that normal links allowed spaces...I'd hate to break all the documents out there by reverting it though.  Especially since it has been like that for a while now...

waylan:
Truth be told, I didn't realize links allowed spaces either. At least I don't ever recall that coming up before.

facelessuser:
I have to say, I generally would lean to the the "old-school" behavior, but since normal links have had this functionality for some time now, and I'd hate to strip out functionality people are used to, it kind of sways me in favor of image links doing the same only for unification of behavior.  Had normal not already had this behavior, I would have voted no for this.

waylan:
I'm torn. I understand we shouldn't break existing links, so changing the link behavior is out of consideration. But the image behavior is technically "correct," so I hesitate to alter it, even if it "unifies" the behavior. We probably should make the change. I'm just not crazy about it.

facelessuser:
[...]

PR #592, opened 2017-10-26
Allow spaces in image url

facelessuser:
Preserving general pre-existing title output, ensure that image URL can have spaces (ref #590).

### DIFF
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -131,7 +131,7 @@ LINK_RE = NOIMG + BRK + \
     r'''\(\s*(<.*?>|((?:(?:\(.*?\))|[^\(\)]))*?)\s*((['"])(.*?)\12\s*)?\)'''
 
 # ![alttxt](http://x.com/) or ![alttxt](<http://x.com/>)
-IMAGE_LINK_RE = r'\!' + BRK + r'\s*\(\s*(<.*?>|([^"\)\s]+\s*"[^"]*"|[^\)\s]*))\s*\)'
+IMAGE_LINK_RE = r'\!' + BRK + r'\s*\(\s*(<.*?>|(?:([^"\)]+)\s*("[^"]*")|[^\)]*))\s*\)'
 
 # [Google][3]
 REFERENCE_RE = NOIMG + BRK + r'\s?\[([^\]]*)\]'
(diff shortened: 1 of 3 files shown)


facelessuser:
Failing issues should be ignored as this issue is not introducing any new failures, but the flake8 issues are being addressed here: https://github.com/Python-Markdown/markdown/pull/593.

waylan:
Interesting. I didn't remember that we actually had tests confirming the preexisting behavior. I'm curious, do we have tests confirming that spaces work in links? If not, then perhaps we're making the wrong change here.

facelessuser:
> I'm curious, do we have tests confirming that spaces work in links?

I don't know yet, but more importantly, we don't have any checks to ensure that they *don't* work or it would already be failing. Omission is some times as bad as explicitly doing something wrong.

The problem is that (for who knows how long) we've allowed them, and possibly given the expectation that this is something that Python Markdown supports.  I'm not sure how many documents we might break.

I'm not sure the motivation against specifically checking that it doesn't support it in images and not links was.

Anyways, the work is done, if we decide to not accept it that is fine.  I think we should either explicitly disallow this in the future, or or go ahead and relax image tests as done here.  Again we may break a good number of document URLs explicitly asserting on this situation, but if we are okay with that, then we should do that instead.

waylan:
My thinking is that if we have tests explicitly requiring that spaces in links be supported, then we should go forward with this change. However, if no such tests exist, my assumption is that the support for it was accidental and we should not go forward with this change. Either way, I agree that changing the link behavior could break too many existing documents so we won't do that, even if that is my preferred way forward.

facelessuser:
If we don't add a test to fail on it, how do we ensure it doesn't get accidentally changed in the future? Or is that not a concern?

Either way, I'll take a look at allowed links.

facelessuser:
I guess we could add tests to ensure it keeps working without updating image links.

waylan:
> If we don't add a test to fail on it, how do we ensure it doesn't get accidentally changed in the future?

What? I think we're talking about different things. When I say "links" I am not referring to images. If links (eg: `[foo](http://example.con/foo bar.html "title")`) already have tests with spaces then we should go forward with this change.  However, if no such tests exist, we should not go forward with this change in image behavior.

Whether we add tests for links with spaces is another matter  which I have not commented on.

facelessuser:
Understood, I was thinking about both images and non images. I'm just looking for a way to make sure we're handling both in a sane manner.  We're on the same page.

waylan:
> I'm just looking for a way to make sure we're handling both in a sane manner. 

To me, being "right" is more important that being consistent in this instance. If the current link behavior was an accident (based on the fact that there are no tests to support it), then I'm saying that we should not change images to match. I'd rather have the inconsistency (for the previously stated backward compatibility reasons) than change the behavior of images. However, if there are supporting tests for the current link behavior, then that was intentional and we should change images to be consistent. 

In other words, before I accept this, someone needs to look through the existing tests and point out any tests which enforce support for spaces in links. If no such tests can be found, then I'm inclined to close instead.

facelessuser:
Understood. That's why I said sane instead of consistent. I just want to make sure whatever we decide on is preserved, but yes these are two separate things. I was just being unclear on what I was talking about, so I'll keep the topic on images.

facelessuser:
I don't see any tests that ensure link spaces work.  Even commonmark doesn't support spaces in their links 🙂 .  I guess the question now is what to do with with the normal links.

facelessuser:
I'm going to close this since there is no actual tests that ensure spaces are allowed for normal links, it makes no sense to add support for spaces in image links.

PR #598, opened 2017-11-18
Feature ancestry

facelessuser:
Feature adds the ability for an inline pattern to define a list of ancestor tag names that should be avoided.  If a pattern would create a descendant of one of the listed tag names, the pattern will not match.

### DIFF
--- a/docs/extensions/api.txt
+++ b/docs/extensions/api.txt
@@ -53,7 +53,7 @@ A pseudo example:
 Inline Patterns {: #inlinepatterns }
 ------------------------------------
 
-Inline Patterns implement the inline HTML element syntax for Markdown such as
+Inline Patterns implement the inline HTML element syntax for Markdown such as 
 `*emphasis*` or `[links](http://example.com)`. Pattern objects should be 
 instances of classes that inherit from `markdown.inlinepatterns.Pattern` or 
 one of its children. Each pattern object uses a single regular expression and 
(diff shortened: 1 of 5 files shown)


docs/extensions/api.txt
```diff
@@ -68,6 +68,12 @@ must have the following methods:
     Accepts a match object and returns an ElementTree element of a plain 
     Unicode string.
 
+* **`getExcludes()`**: 
+
+    Returns an array of tag names that are undesirable ancestors. The pattern 
+    should not match if it would cause the content to be a descendant of one 
+    of the tag names in the list.
```
waylan: "Returns an ***array***... tag names in the ***list***." Let's be consistent here. And in Python we call them lists, not arrays.

markdown/inlinepatterns.py
```diff
@@ -207,6 +207,10 @@ def __init__(self, pattern, markdown_instance=None):
         if markdown_instance:
             self.markdown = markdown_instance
 
+    def getExcludes(self):
```
waylan: Any reason why this is a method? The end user of API (extension developer) is defining a list, not "getting" the list. The fact that the list is being retrieved is not relevant to the user. Rather, the user is interested in defining what to exclude. Why not a property `exclude = [...]`? Does it even need to be specific to an instance of the class?

waylan (commented):
Overall it looks good. I'm leaving for a week tomorrow so I probably won't look at this again until next weekend at the earliest.

docs/extensions/api.txt
facelessuser: I work in many languages, so things blur together at times and I make mistakes like this.

markdown/inlinepatterns.py
facelessuser: Something like `exclude` is fine (maybe something a bit more unique in case existing extensions happen to use that name for something else).  We can make the default a tuple to be safe so no one will accidentally modify the base list and ripple through all instances that use the base.  That will force people to override the default opposed to appending to it.

To be safe, we can probably just wrap the treeprocessor access with a try/except to be safe as well.

facelessuser:
Suggested changes have been made, so I'm all done unless there are more suggestions.  Whenever you get back, just let me know if there is anything else that needs to be addressed.

Issue #601, opened 2017-11-30
Switch documentation host


waylan:
As pointed out in #554 I didn't update the publicly hosted docs with the 2.6.9 release. I now remember that I tried but ran into problems uploading the docs. Then I ran out of time and forgot to come back to it later.

In any event as I explain [here][1]:

> The homepage at http://pythonhosted.org states: 
> 
> > To upload documentation, go to your package edit page (`http://pypi.python.org/pypi?%3Aaction=pkg_edit&name=yourpackage`), and fill out the form at the bottom of the page. 
> 
> However, there is no longer a form at the bottom of the edit page for uploading documentation. Instead I only see: 
> 
> > If you would like to DESTROY any existing documentation hosted at `http://pythonhosted.org/ProjectName` Use this button, There is no undo. 
> > 
> > [Destroy Documentation] 
> 
> I also went to pypi.org and logged in there. But I don't see any options for editing my projects or uploading documentation on that site. 
> 
> So, how do I upload an update to my documentation? 

In addition to the above, pythonhosted seems to have been rather flakey lately. I'm considering switching to GitHub Pages (probably using [Organization Pages][2] so the URL would be `https://Python-Markdown.github.io`). I'm not interested in readthedocs as they require you to have them build your docs and then we cannot use our own tooling to build our own docs (they do support MkDocs, but an older version with less than complete support for Markdown extensions, which is a nonstarter IMO). Any thoughts?

[1]: https://groups.google.com/forum/#!topic/comp.lang.python/QZ-CSfIbRes
[2]: https://help.github.com/articles/user-organization-and-project-pages/#user--organization-pages

mitya57:
In case you want to update the `pythonhosted.org` page: last time I tried, setuptools’ `upload_docs` command worked. This means you can either use POST requests directly (see the [implementation](https://github.com/pypa/setuptools/blob/master/setuptools/command/upload_docs.py#L144)) or adapt that command to use your custom tarball instead of a Sphinx-generated one.

waylan:
As a reminder, we are not yet using setuptools, so we don't have the `upload_docs` command available. We could certainly switch over (the md3 branch already has been), but that doesn't help for a previous release.

However, I do see some value in using the mechanism in that script to upload something redirecting people to the new docs.

waylan:
[...]

PR #602, opened 2017-12-02
Switch to MKDocs

waylan:
This is a work in progress.

I merged in 6f87b32 from the md3 branch. That branch has all of the
extension docs removed, so they haven't been converted yet. And I
only resolved conflicts and updated mkdocs.yml to the current MkDocs
format.

Addresses #601.


waylan:
I still need to do an audit of the docs to make sure I didn't miss anything, but I pushed the deployment up to GitHub Pages. However, weirdly the homepage is not showing unless I include `index.html` in the URL. So this works https://python-markdown.github.io/index.html but this doesn't https://python-markdown.github.io/. Even more strange, this works fine https://python-markdown.github.io/extensions/. Not sure what's going on with the homepage there. Unless its some weird caching thing.

facelessuser:
https://python-markdown.github.io/ works for me

facelessuser:
Just my 2 cents, but links are hard to see in that theme.  If you can make them stand out a little, it might help with navigation.

waylan:
The homepage works for me from a different location/machine. Must have been a caching thing causing the problem.

> Just my 2 cents, but links are hard to see in that theme. If you can make them stand out a little, it might help with navigation.

The CSS is completely unmodified from the Sphinx theme. That is the "problem" with navigation? I didn't have any problems, but then I'm intimately aware of the structure of the theme. 

I did notice that tables seem to be completely unstyled, so that may need to be investigated and possibly addressed. But that would likely happen upstream on the theme, not here.

Regardless, at this point I'm mostly concerned with the MkDocs config, whether all internal links are properly converted, and if the rendered Markdown works with the theme (for example, some code blocks might need to have a language identified as language guessing appears to be getting it wrong). I'll be doing an audit page-by-page, so if anyone sees anything, let me know in a comment and I'll be sure to address it.

facelessuser:
> The CSS is completely unmodified from the Sphinx theme. That is the "problem" with navigation? I didn't have any problems, but then I'm intimately aware of the structure of the theme.

Dark navy blue and black is hard to distinguish...at least for me (at first).  I had to mouse around to get the link to turn red on hover to see where they were.  Once I stared at it a bit, I more easily noticed the subtle differences, but I'd argue they are a bit *too* subtle.  It won't really affect me much as I don't often have to reference the docs, just my general impression.  Lightening it up to `#007fb4` makes them easy to see, but even `#006d9b` would do it less though but give it a bit more contrast.

It's up to you, just figured I'd mention it, I'll deal regardless.  After a bit my eyes can see the difference.  I just personally dislike sites where clickable things are too subtle (this happens a lot with the flat theme trend).  I feel like I just flail my mouse around until I see the cursor or the link change to some easy to see hover state.

waylan:
After a lot of cleanup, I merged this. We can deal with theme related issues separately.

makefile
```diff
@@ -22,6 +22,7 @@ install:
 deploy:
 	python setup.py sdist --manifest-only
 	python setup.py sdist --formats zip,gztar upload
+	mkdocs gh-deploy -r pages -b master
```
waylan, line 25: Turns out this doesn't work. It hoses your local master branch the first time and then (after more commits are added) results in merge conflicts. See davisp/ghp-import#65 for details.

PR #604, opened 2017-12-07
Upped version to 2.6.10

### LINKED ISSUE #601: Switch documentation host

waylan:
This version was released to force PyPI to point to the new docs location.
Closes #601. The old PyPI hosted docs can be deleted after this.

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -1,7 +1,9 @@
-title:      Change Log
+title: Change Log
 
 Python-Markdown Change Log
 =========================
+
+Dec 7, 2017: Released version 2.6.10 (a documentation update).
 
 Aug 17, 2017: Released version 2.6.9 (a bug-fix release).
 
--- a/markdown/__version__.py
+++ b/markdown/__version__.py
@@ -5,7 +5,7 @@
 # (major, minor, micro, alpha/beta/rc/final, #)
 # (1, 1, 2, 'alpha', 0) => "1.1.2.dev"
 # (1, 2, 0, 'beta', 2) => "1.2b2"
-version_info = (2, 6, 9, 'final', 0)
+version_info = (2, 6, 10, 'final', 0)
 
 
 def _get_version():
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -9,7 +9,7 @@ use_directory_urls: true
 theme:
   name: nature
   icon: py.png
-  release: 2.6.9
+  release: 2.6.10
   issue_tracker: https://github.com/Python-Markdown/markdown/issues
 
 pages:


waylan:
Arg! Apparently we can't upload to PyPI via HTTP anymore (only HTTPS). But that only works with newer Python versions, which I don't have installed on this machine (and can't easily update for reasons). I can't even get twine to work due to a [InsecurePlatformWarning][1].

Guess the deployment to PyPI will need to wait to I have access to another machine (in a few hours). I understand PyPI's end goal, but their path to getting there is extremely frustrating and feels very user hostile.

[1]: http://urllib3.readthedocs.io/en/latest/advanced-usage.html#ssl-warnings

facelessuser:
I personally let [Travis deploy](https://github.com/facelessuser/pymdown-extensions/blob/master/.travis.yml#L53) when I cut a release.  Makes it easier for me (until the next drastic change).

waylan:
I also discovered that my (clever, I thought) [solution][1] for deploying the docs is faulty. See davisp/ghp-import#65. I hosed my local repo and had to do a hard reset. Then, after merging some upstream commits, the deploy failed as it was causing a merge conflict. Unless/until my request in davisp/ghp-import#65 is met, I need to script the deploy.

[1]: https://github.com/Python-Markdown/markdown/pull/602/commits/dd56b87d2be8ed67a388dd8f4522be430abce399

facelessuser:
Unfortunately, I don't have much experience with deploying to group.  I hadn't realized how much different it was.

waylan:
> I personally let Travis deploy when I cut a release. Makes it easier for me (until the next drastic change).

Deploying with Travis means dealing with deploy/auth keys. Until now, it was always easier to just run the command locally. But deploy keys won't work anyway. According to GitHub's [docs][0], "Deploy keys aren't supported for Organization Pages." You have to "[set up a machine user as a member of your organization][1]". I don't even know if that is possible on Travis (and haven't looked into it) as you need to "run the `ssh-keygen` procedure on your server".

Of course, that doesn't apply to PyPI, but PyPi still requires an auth key, which I'd rather not have to deal with.

[0]: https://help.github.com/articles/user-organization-and-project-pages/#building-your-user--organization-pages
[1]: https://developer.github.com/v3/guides/managing-deploy-keys/#machine-users

facelessuser:
Ugh. What a headache.

waylan:
See also my followup with MkDocs in [this comment](https://github.com/mkdocs/mkdocs/issues/543#issuecomment-350138685). If/when mkdocs/mkdocs#543 gets implemented, deploying to user/org pages should be easier than they are now.

waylan:
Well, I pushed the release to PyPI, but I got this response:

```
running upload
Submitting dist/Markdown-2.6.10.zip to https://upload.pypi.org/legacy/
Server response (200): OK
Submitting dist/Markdown-2.6.10.tar.gz to https://upload.pypi.org/legacy/
Upload failed (400): Only one sdist may be uploaded per release.
error: Upload failed (400): Only one sdist may be uploaded per release.
```

The line I've been using for years in my makefile is:

```
python setup.py sdist --formats zip,gztar upload
```

It appears that the `zip` file uploaded and is now blocking the `tar.gz` file from uploading. So I guess 2.6.10 will have a `zip` file only.

facelessuser:
That's weird.  So if the base name is the same and the extension is different it rejects it?  I guess the name is technically different for win, osx, linux, and with x86 etc.

What a mess.

waylan:
I suppose it is all part of their push for wheels, which I'd use of we were using setuptools. But it seems ridiculous that you need non-standard-library libs to use the standard package manager's preferred defaults. In any event, I'm mostly posting things here to document them for future reference, not to complain.

facelessuser:
Very interesting.  I'm glad you are documenting this.  The information will be useful for the future.  Most of the time I don't know about these issues until I run into them (like what is happening here).

jimporter:
> I can't even get twine to work due to a InsecurePlatformWarning.

`pip install 'urllib3[secure]'` is how I solved this. You can just install it in a virtualenv (or whatever you like) and then happily use `twine` to upload to PyPI even on older Pythons; I've done it on 2.7.6 (I tend to use 2.7 as my dev env because it's easier to make sure I'm not accidentally breaking things for 2.7).

PR #606, opened 2017-12-09
Switch to setuptools and wheels

waylan:


setup.py
```diff
@@ -98,9 +65,10 @@ def run(self):
     maintainer_email='waylan.limberg@icloud.com',
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
-    scripts=['bin/%s' % SCRIPT_NAME],
-    cmdclass={
-        'install_scripts': md_install_scripts
+    entry_points={
+        'console_scripts': [
+            'markdown = markdown.__main__:run',
```
mitya57, line 70: Should not this be `markdown_py = ...`? From Debian packaging perspective, I would still like to avoid conflicts with the Perl implementation.

waylan: Yes, you're correct, it should. I forgot about that. Feel free to push a fix for that yourself. Probably should restore the old comment to remind us as well.

PR #608, opened 2017-12-10
Restore code to use markdown_py as the console script name.

mitya57:
We still need to avoid conflicts with the original Perl implementation.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -30,6 +30,11 @@ def get_version():
 version, version_info, DEVSTATUS = get_version()
 
 
+# The command line script name.  Currently set to "markdown_py" so as not to
+# conflict with the perl implimentation (which uses "markdown").
+SCRIPT_NAME = 'markdown_py'
+
+
 long_description = '''
 This is a Python implementation of John Gruber's Markdown_.
 It is almost completely compliant with the reference implementation,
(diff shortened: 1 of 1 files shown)


waylan:
Looks good. Although I was wondering what your thoughts are about using a shorter name going forward?  Perhaps `pmd`, `pymd`, `md.py` or similar. I realize anyone could create their own alias with such a name, but why not use it from the start?

If we do, then should we leave the existing script name in place? Should we deprecate it? Thoughts?

setup.py
```diff
@@ -52,6 +52,13 @@ def get_version():
 '''
 
 
+# The command line script name.  Currently set to "markdown_py" so as not to
+# conflict with the perl implimentation (which uses "markdown").  We can't use
+# "markdown.py" as the default config on some systems will cause the script to
+# try to import itself rather than the library which will raise an error.
```
waylan: I'm not sure the second sentence is accurate for an entry point. We might want to remove that one.

mitya57:
There is existing code that uses `markdown_py` name. See [Debian codesearch output](https://codesearch.debian.net/search?q=markdown_py) for examples. I think it would be good to not break it and continue providing that name.

If you think we should add a new shorter name, then I don’t mind, but I’m also not sure how many users will use it.

setup.py
mitya57: Do you mean the third sentence? The second still applies, I think, as the entry point usually gets installed into the same namespace as a normal script (something on $PATH).

waylan: Sorry, yes, third sentence.

mitya57: Ok, the third sentence removed.

Issue #611, opened 2017-12-17
Invalid escape sequences in regular expressions


samuelharmer:
Something I noticed in some of my unit tests:
```
[...]\markdown\extensions\toc.py:28: DeprecationWarning: invalid escape sequence \w
  value = re.sub('[^\w\s-]', '', value.decode('ascii')).strip().lower()
[...]\markdown\extensions\toc.py:29: DeprecationWarning: invalid escape sequence \s
  return re.sub('[%s\s]+' % separator, separator, value)
```

There might be more occurrences of this problem than those I've found, but making the regular expressions on [lines 28 and 29][1] into raw strings (i.e. `'[^\w\s-]'` into `r'[^\w\s-]'`) silenced these warnings. I'm not sure whether this is the correct fix.

(Python 3.6.3)

[1]: https://github.com/Python-Markdown/markdown/blob/1edbb9dcb998966e43e02f5e345fe3d08c0307dc/markdown/extensions/toc.py#L28-L29

facelessuser:
Yes, these should all get fixed.  PY 3.6 is much more strict with its tokenizer (I ran into this on another project), and PY 3.7 will actually throw exceptions on these making it more critical to address.  The motivation behind this is they want to start discouraging people from using unsupported ASCII escapes in case they want to support other back references in replaces in the future, like `\N` etc., but those are long term goals.  This is specifically referring to `Re`, but Python in generally may also be more strict with normal strings in general.

Anyways, we should probably scan the code for all cases like this were regex strings were lazily using non-raw strings.

waylan:
Yup, I agree as well. One thing we need to be aware of is that many regex are build up by concatenating multiple strings together. That's not a problem in itself, it just adds another level of complexity to the process. For example, what happens when you concatenate a raw string and a regular string? Does it matter if one or the other comes first (the way Python operators work it's a possibility depending on the implementation)? I expect all we really care about is that the final string (which gets passed to re) is raw, regardless of what the individual pieces started out as. However, it would probably be good form to have each piece start out raw.

Of course PRs are welcome.

facelessuser:
Good news is that it doesn't really matter. Raw or not raw, it all concatenates the same. The r in front only matters at compile time, then it's just a string. I concatenate raw strings for regex all the time. By time the operator are working on the objects, they're just strings.

facelessuser:
I will tentatively assign this task to me assuming someone doesn't get a pull together before I do.

PR #614, opened 2017-12-21
Make sure regex patterns are raw strings

facelessuser:
Python 3.6 is starting to reject invalid escapes.  Regular expression patterns should be raw strings to avoid having regex escapes being mistaken for invalid string escapes. Ref #611.

Address almost all strings being passed as regex to be a reminder that they are patterns.  Avoided things like STX and ETX as they are used in patterns and in replaces, and while re will convert Unicode escapes, raw strings in general do not.


facelessuser:
Any thoughts on this? Or can we merge this?

PR #615, opened 2017-12-22
Test case for nested paragraph in list

browniebroke:
When doing a nested paragraph in a list element, the first following list item has a nested paragraph, but not the other ones.

I've added a test case to cover the expected output and I'm currently getting the following diff:
```
-<li>That item is simple</li>
+<li>
+<p>That item is simple</p>
+</li>
```

I noticed this when looking at a [recent DRF release note](http://www.django-rest-framework.org/topics/release-notes/#374), where the output feels a bit weird.

Not sure yet how to fix it though... I'll keep digging for a bit.

### DIFF
--- a/tests/basic/list-nested-paragraph.html
+++ b/tests/basic/list-nested-paragraph.html
@@ -1,0 +1,11 @@
+<ul>
+<li>
+<p>This is a list</p>
+<p>This is a nested multi-line content</p>
+<p>Another nested paragraph in the same
+  list item</p>
+<p>And yet another one</p>
+</li>
+<li>That item is simple</li>
+<li>So is that one</li>
+</ul>
--- a/tests/basic/list-nested-paragraph.txt
+++ b/tests/basic/list-nested-paragraph.txt
@@ -1,0 +1,11 @@
+* This is a list
+
+      This is a nested multi-line content
+
+      Another nested paragraph in the same
+      list item
+
+      And yet another one
+
+* That item is simple
+* So is that one


waylan:
The existing behavior is actually correct. To understand, ignore the multi-paragraph item for now and consider what many refer to as a "loose list". The rules [state][1]:

> If list items are separated by blank lines, Markdown will wrap the
items in `<p>` tags in the HTML output. For example, this input:
>
>     *   Bird
>     *   Magic
>
> will turn into:
>
>     <ul>
>     <li>Bird</li>
>     <li>Magic</li>
>     </ul>
>
> But this:
>
>     *   Bird
>
>     *   Magic
>
> will turn into:
>
>     <ul>
>     <li><p>Bird</p></li>
>     <li><p>Magic</p></li>
>     </ul>

Notice that both the list item before and the list item after the blank line are wrapped in `<p>` tags. Obviously, by it's nature, a multi-paragraph list item is also a "loose" list item, therefore the item following it will also be a "loose" list item. This is supported by [Babelmark][2], where Python-Markdown's output matches the reference implementation (markdown.pl).

Of course, as you note, this looks weird in the output. In fact, you may have noticed that Babelmark shows that the Commonmark implementations maintain a consistency. In fact, the Commonmark spec is different than the original Markdown rules in this regard. The spec [states][3]:

> A list is loose if any of its constituent list items are separated by blank lines, or if any of its constituent list items directly contain two block-level elements with a blank line between them. Otherwise a list is tight. (The difference in HTML output is that paragraphs in a loose list are wrapped in `<p>` tags, while paragraphs in a tight list are not.)

In other words, either the entire list is "loose" or the entire list is "tight". However, Python-Markdown is a Markdown implementation, not a Commonmark implementation. Regardless, the logic in this part of the Commonmark spec makes sense. Of course, we aren't going to change the parser (as we are a Markdown implementation not a Commonmark implementation), but I always encourage document authors to follow that rule in principle. When authoring your Markdown lists, if any item is "loose, then make all items in the list "loose". In others words, add a blank line between each of the simple items to maintain consistency with the more complex items. Therefore, your example list should be formatted like this:

```
* This is a list
 
    This is a nested multi-line content
 
    Another nested paragraph in the same
    list item
 
    And yet another one
 
* That item is simple

* So is that one
```

Note that [Babelmark][4] shows a very consistent rendering of that source text. This is the sort of thing that a [Markdown linter][5] would have caught (by suggesting that a blank line be added between the last two items). It's one of those things that is not strictly required by the rules, but is a good practice to avoid surprising results.

[1]: https://daringfireball.net/projects/markdown/syntax#list
[2]: http://johnmacfarlane.net/babelmark2/?normalize=1&text=%23+A+tight+list%0A%0A*+++Bird%0A*+++Magic%0A%0A%23+A+loose+list%0A%0A*+++Bird%0A%0A*+++Magic%0A%0A%23+A+mixed+list%0A%0A*+This+is+a+list%0A+%0A++++This+is+a+nested+multi-line+content%0A+%0A++++Another+nested+paragraph+in+the+same%0A++++list+item%0A+%0A++++And+yet+another+one%0A+%0A*+That+item+is+simple%0A*+So+is+that+one
[3]: http://spec.commonmark.org/0.28/#loose
[4]: http://johnmacfarlane.net/babelmark2/?normalize=1&text=*+This+is+a+list%0A+%0A++++This+is+a+nested+multi-line+content%0A+%0A++++Another+nested+paragraph+in+the+same%0A++++list+item%0A+%0A++++And+yet+another+one%0A+%0A*+That+item+is+simple%0A%0A*+So+is+that+one
[5]: https://github.com/markdownlint/markdownlint#markdown-lint-tool

browniebroke:
Waw! Thank you for your thorough answer, after digging a bit I started to expect it, but I don't understand enough of the dark corner of Markdown. I appreciate you taking the time to write this detailed explanation!

Issue #618, opened 2018-01-03
Deprecation Warnings


jkempinger-dv:
I just updated to 2.6.10 and now I'm seeing DeprecationWarnings

```
...markdown/treeprocessors.py:275: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.

...markdown/treeprocessors.py:276: FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
```

This seems to be the same issue as #254, and changing lines 275 and 276 to

```
while parent is not None:
    if parent is not None:
```

eliminates these warnings.

waylan:
Thanks for the report.

The fix is easy enough, but those lines were introduced in #598 very recently. Perhaps our CI tests should be catching that sort of thing (the warnings don't even show up in our [test log][1]). Maybe we should update our tests to check for that sort of thing.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/304174819

jkempinger-dv:
I'm using python 3.6.4 with PyInstaller.  I see the warnings when I run my built application.  I don't recall seeing the warnings when running the source files through the python interpreter on the CLI.  Let me know if you would like any other information/debugging about my setup!

waylan:
Apparently, the Python interpreter has these set to be ignored by [default][0]. However, in other environments, they can still get issued. So we need to change the default when running our tests if we want to prevent introducing them.  Using Python [`-W`][1] flag or setting the [`PYTHONWARNINGS`][2] env variable should do the trick. According to [this][3], setting either to `default` (or `d` for short) will override any ignored warnings.

I'm thinking of adding the following line to the `tox.ini` file:

    setenv = PYTHONWARNINGS = error::DeprecationWarning,default::PendingDeprecationWarning

That way `PendingDeprecationWarning` get listed (but don't cause the tests to fail) and `DeprecationWarning` cause the tests to fail.

What is not clear to me is why we don't already get the warnings. We do get depreciation warnings raised from the Markdown lib, so why not from the standard libs? Or is nose specifically changing the defaults for the tested libs, but not the standard libs? I'll have to experiment.

[0]: https://docs.python.org/3/library/warnings.html?highlight=warn#default-warning-filters
[1]: https://docs.python.org/3/using/cmdline.html#cmdoption-w
[...]

PR #619, opened 2018-01-04
Warnings should cause tests to fail.

waylan:
Related to #618.

Although, I don't see that this any offers any value. Mostly doing this for later reference.

Note that this change does not force the use of the pure Python Elementtree lib. But in my testing, even when doing so, there was no difference.

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -2,6 +2,7 @@ import os
 import markdown
 import codecs
 import difflib
+import warnings
 try:
     import nose
 except ImportError as e:
@@ -179,4 +180,10 @@ def generate_all():
 
 
 def run():
+    # Warnings should cause tests to fail...
+    warnings.simplefilter('error')
+    # Except for the warnings that shouldn't
+    warnings.filterwarnings('default', category=PendingDeprecationWarning)
+    warnings.filterwarnings('default', category=DeprecationWarning, module='markdown')
+
     nose.main(addplugins=[HtmlOutput(), Markdown()])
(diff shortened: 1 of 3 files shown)


tests/__init__.py
```diff
@@ -179,4 +180,10 @@ def generate_all():
 
 
 def run():
+    # Warnings should cause tests to fail...
+    warnings.simplefilter('error')
+    # Execpt for the warnings that shouldn't
```
mitya57: Typo: Ex**ce**pt

mitya57 (commented):
In my opinion it is nice to have as it actually makes change on PyPI.

waylan:
After making some changes, rebasing, and and force pushing, the complete development history no longer shows here. So, for reference, with only the first commit the tests failed as can be seen [here][1]. The DeprecationWarnings from ElementTree resulted in failed tests and the PendingDeprecationWarnings were included in the output (but did not cause failures). Note that the warnings were issued in the PyPy job as it is the only one which uses pure Python ElementTree. All others use the C implementation which does not issue the warnings.

The second commit updates the code and avoids any warnings being issued except warnings issued by the Markdown lib. We want those deprecation warnings to be issued but not cause the tests to fail. We can see it works properly [here][2]. 

I feel confident that this should help prevent us from accepting any future regressions of this type. We do need to be careful to check the test logs for PendingDeprecationWarnings though.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/324864321
[2]: https://travis-ci.org/Python-Markdown/markdown/jobs/325171829

mitya57:
Looks good! :+1:

PR #620, opened 2018-01-04
Up version to 2.6.11

waylan:
Only a couple minor bugs, but this will give us a wheel on PyPI, etc. Seeing how the last release went, it would be good to not keep it around.


waylan:
I'm happy to report we have a successful release. Both a wheel and tar.gz source file uploaded to [PyPI][1] and updated [documentation][2]. All without incident. Looks like I worked all the kinks out.

[1]: https://pypi.python.org/pypi/Markdown/2.6.11
[2]: https://python-markdown.github.io/change_log/

PR #622, opened 2018-01-06
New test tools

waylan:
The new test tools are designed to both be used by Python-Markdown and by third party extensions.  As all tests are based on the Python standard lib `unittest`, no third party dependencies exist (we no longer rely on `nose`). See the included documentation for the details.

The following is my proposal for how we should proceed with Python-Markdown's tests going forward:

Any new syntax tests for Python-Markdown should be developed with the `markdown.test_tools.TestCase` class. Any old file-based tests which need updating should be converted to this new type of test. A few new tests have been included with this PR and can be found in the `tests/test_syntax` directory. That directory should contain two subdirs: `block` and `inline`. The `block` subdir should contain files of tests for block-level elements and the `inline` subdir should contain files of tests for inline-level elements. Each file should contain tests for a single type of element.

The old file-based tests still exist but are now wrapped in `markdown.test_tools.LegacyTestCase` class instances. All of the wrapping classes, which include all of the settings and config for each tests are contained in the `tests/test_legacy.py` file. Over time, as these tests get moved to the new format, they should be removed from the legacy tests (none have been removed yet). Eventually, only the Perl and PHP test suites should remain (no edits should happen to those test files; if a test won't pass for good reason, then it should be skipped as some already are).


waylan:
Hmm, it seems that Travis has had a lot of jobs time-out recently. Usually it is somewhere in the setup process (downloading deps or similar). If I restart them they run fine. Anyone else seeing that?

facelessuser:
Yes. Right around the new year, Travis went to crap. It's been timeouts since on multiple of my repos. So it's not specific to this repo.

PR #623, opened 2018-01-07
Correct spelling mistakes.

EdwardBetts:


tests/test_extensions.py
```diff
@@ -92,8 +92,8 @@ def testNestedAbbr(self):
                '*[ABBR]: Abreviation'
```
waylan: If you are going to change the output, you also need to change the input or the test [fails](https://travis-ci.org/Python-Markdown/markdown/jobs/325987980#L591).

tests/php/extra/Footnotes.text
```diff
@@ -50,7 +50,7 @@ footnote test[^reference].
 	This footnote has a footnote of its own.[^nested]
 
 [^nested]:
-	This footnote should appear even though as it is refered
+	This footnote should appear even though as it is referred
```
waylan: The `php` test files are borrowed directly from the PHP project and should always remain unchanged. Please do not make any changes to these files. These changes need to be undone before I can accept this.

tests/pl/Tests_2004/Markdown Documentation - Basics.html
```diff
@@ -114,7 +114,7 @@ <h2>Lists</h2>
 
 <p>Unordered (bulleted) lists use asterisks, pluses, and hyphens (<code>*</code>,
 <code>+</code>, and <code>-</code>) as list markers. These three markers are
-interchangable; this:</p>
+interchangeable; this:</p>
```
waylan: The `pl` test files are borrowed directly from the Perl Markdown project and should always remain unchanged. Please do not make any changes to these files. These changes need to be undone before I can accept this.

waylan (changes requested):
Wow, that was a lot of misspelled words in places they would never bother me. I'm happy to accept the changes with a few exceptions. If you could make the requested changes, I'll be happy to accept this.

EdwardBetts:
I've made the changes that you asked for. I ran the tests and they passed.

waylan (approved):

Issue #613, opened 2017-12-20
Image Tag does not work if the file name contains '()'


spokvulcan:
Example:

```
![image](sometext(sometext)sometext.png)

[link](sometext(sometext)sometext.png)
```

Convert to html image:

```
<p>
<img alt="images" src="sometext(sometext">sometext.png)
</p>
```

Convert to html link:

```
<p>
<a href="sometext(sometext)sometext.png">link</a>
</p>
```

waylan:
Thanks for the report. It looks like you found a bug.

Sigh. Another one of the differences between links and images. Sometimes I think we should just use the same partial regex for both, but then there is some behavior of links we don't want to replicate elsewhere (for example see #590).

facelessuser:
If/when we ever get 3 out the door, we can probably make that change and remove spaces from the original links.

waylan:
Fixed in #629.

Issue #551, opened 2017-03-05
Infinite execution on some input


dattaz:
With some input, i have a infinite execution, with markdown function, no exception raise.

Step to reproduce: https://gist.github.com/anonymous/ffab9ad433127893f04b9d009cd21444

waylan:
Thanks for the report.

First from a practical standpoint, why is that not in a code block or span? That is not Markdown code but is close enough collisions would not be surprising. As it turns out, most implementations just pass it through unaltered, although there are some variations. Interestingly, at least three other implementations fail to parse it at all (according to Babelmark [^1]).

That said, you shouldn't be able to crash a server by submitting some plain text, so we do need to fix this. However, the fix will be to simply have it not crash/enter an infinite loop. What is actually returned is unpredictable as this is not really "valid" Markdown.

Without doing any debugging, my best guess is that the nested brackets regex (for the link syntax) is the problem. Unlike in Perl (which the reference implementation was written in), Python's regex engine doesn't do recursion so we fake it with six levels of nested hard-coded in. The assumption was that no one would ever try to create more than that. Personally, I was never crazy about the solution, but it was in-place long before I got involved and nothing better has ever been proposed.

[^1]: I did not include the Babelmark link as it causes my server to freeze and I have to manually reset it. Please don't copy the input to Babelmark yourself. Thank you.

dattaz:
I'm totally agree with you, it's should be code block, but it's data from stackoverflow dump that hasn't been format in the good way by the author of the comment.

> However, the fix will be to simply have it not crash/enter an infinite loop. What is actually returned is unpredictable as this is not really "valid" Markdown

Yes it's logical, because main problem is infinite loop

waylan:
So it does not appear to be an infinite loop. If you wait long enough it will finish, but most servers would time out by then. And I seem to have narrowed the minimum input down to this:

```
[[&quot;][&quot;][&quot;][&quot;][&quot;][&quot;][&quot;][&quot;][&quot;]]
```

As you remove each `[&quot;]` from the string, it runs noticeably faster. In order to replicate it, the opening and closing double brackets are necessary as is having an HTML entity in each set of brackets.

waylan:
Interestingly, this seemed to not be a noticeable problem even though it contains the same number of sub-strings:

```
[[&quot;][&quot;][&quot;][&quot;][&quot;][&quot;]] [[&quot;][&quot;][&quot;][&quot;][&quot;][&quot;]]
```

[...]

Issue #161, opened 2012-11-18
Render time is superlinear in the size of a paragraph


kmcallister:
It's easiest to see with some slightly complex markup, a link for example.  Using the latest Python-Markdown from Git I ran:

``` .py
import time
import sys
import markdown

def time_conv(md):
    engine = markdown.Markdown()
    t0 = time.time()
    engine.convert(md)
    return time.time() - t0

link = '[link](http://google.com)'

for n in xrange(1,1000,10):
    md = ' '.join([link] * n)
    print n, time_conv(md)
    sys.stdout.flush()
```

When I plot the output in Gnuplot, it's clearly superlinear.  Taking the square root of the time produces a nice line, so I suspect it is quadratic in the size of the paragraph.

This doesn't happen when I separate the links with `'\n\n'` to make separate paragraphs.  I first noticed this problem when using single `'\n'`s with the `nl2br` extension.

I'm using Python 2.7.3 from Debian unstable.

waylan:
I'm not sure what to make of this report.

Python-Markdown has never claimed to be fast. Just fast enough. If you want
high performance, I'd suggest a C implementation with python bindings.

That said, I suppose your results (well, what I'm assuming your results
were based on your description because you actually didn't provide the
results) make sense. Inline markup is processed one paragraph at a time. So
the regex will parse a paragraph with one link pretty quickly. However, a
single paragraph with 1,000 links will take some time. Not sure how to
speed that up aside from a complete re-factor - which is not happening any
time soon - besides I'm not even sure how to re-factor for that use case
anyway.

It appears from your report that this issue was brought on by using nl2br.
My personal opinion is that that is the worst addon to markdown ever -
which shouldn't have existed anywhere in any implementation - but that's
just me. Point is I'm not really interested in doing anything about this.
Of course, patches are always welcome.

kmcallister:
My complaint isn't "it's slow" but rather the quadratic scaling.  In particular, this makes it harder for us to bound the rendering time by bounding the input length.

The performance with separate paragraphs is indeed "fast enough".  I would love to keep using Python-Markdown; it's remarkably flexible and we have already used that flexibility in numerous ways.  Porting all of those customizations to a C library seems like a world of pain.

[...]

PR #629, opened 2018-01-13
Flexible inline

facelessuser:
Add new InlineProcessor class that handles inline processing much better and allows for more flexibility.  This adds new InlineProcessors that no longer utilize unnecessary pretext and posttext captures.  New class can accept the buffer that is being worked on and manually process the text without regex and return new replacement bounds.  This helps us to handle links in a better way and handle nested brackets and logic that is too much for regular expression.  The refactor also allows image links to have links/paths with spaces like links. Ref #551, #613, #590, #161.


facelessuser:
Figured I'd get some discussion going on this and start moving this forward.

facelessuser:
Seems pypy is failing for the recently discussed issue Future warnings with elementtree:

```
FutureWarning: The behavior of this method will change in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.
```

While this is a separate issue, I might take care of it here as well.

waylan:
> Seems pypy is failing for the recently discussed issue Future warnings with elementtree:

Yeah, we keep fixing these, and then every time we do more work on ElementTree related code, we introduce them. So I made sure a test fails to stop us from reintroducing the warnings. I consider these warnings a reason to not accept a PR, so yeah, they should be addressed (which I see you did).

tests/misc/image.html
```diff
@@ -1,5 +1,5 @@
 <p><img alt="Poster" src="http://humane_man.jpg" title="The most humane man." /></p>
 <p><img alt="Poster" src="http://humane_man.jpg" title="The most humane man." /></p>
 <p><img alt="Blank" src="" /></p>
-<p>![Fail](http://humane man.jpg "The most humane man.")</p>
-<p>![Fail](http://humane man.jpg)</p>
\ No newline at end of file
+<p><img alt="Image" src="http://humane man.jpg" title="The most humane man." /></p>
+<p><img alt="Image" src="http://humane man.jpg" /></p>
```
waylan: Rather than editing these, lets delete them and recreate the tests in the new syntax tests.

tests/misc/image.txt
```diff
@@ -7,6 +7,6 @@
 
 ![Blank]()
 
-![Fail](http://humane man.jpg "The most humane man.")
+![Image](http://humane man.jpg "The most humane man.")
 
-![Fail](http://humane man.jpg)
+![Image](http://humane man.jpg)
```
waylan: Rather than editing these, lets delete them and recreate the tests in the new syntax tests.

tests/test_syntax/inline/test_advanced_images.py
```diff
@@ -0,0 +1,99 @@
+from markdown.test_tools import TestCase
+
+
+class TestAdvancedImages(TestCase):
```
waylan: I suppose we can leave the class named "advanced" but let's just name the file `images.py`. Eventually all "image" related tests should be moved here.

tests/test_syntax/inline/test_advanced_links.py
```diff
@@ -0,0 +1,98 @@
+from markdown.test_tools import TestCase
+
+
+class TestAdvancedLinks(TestCase):
```
waylan: I suppose we can leave the class named "advanced" but let's just name the file `links.py`. Eventually all "link" related tests should be moved here.

waylan (changes requested):
I still need to take a closer look at what some of the code is actually doing, but here are a few general organization issues I spotted. Overall it looks good. Thanks for doing the work on this.

tests/misc/image.html
facelessuser: Makes sense. I can probably move the whole file to the new format.

tests/test_syntax/inline/test_advanced_images.py
facelessuser: Sounds good.

tests/test_syntax/inline/test_advanced_links.py
facelessuser: Sounds good.

tests/misc/image.html
waylan: The whole file woull be even better

waylan:
Oh and while it may seem obvious, I should note that the docs need to be updated for these changes. Specially, the Extension API docs.

facelessuser:
Makes sense. It's a pretty big change. I figured it'd take a little bit to work through the review.

facelessuser:
Yeah, I didn't make changes to the docs yet only because I wanted to make sure we were all in agreement on the new processors behavior. Figured I'd wait until we were good with the code.

facelessuser:
Went ahead and addressed the test changes suggested, and provided some API documentation.  I'll address spelling issues if Travis notifies of any.

docs/extensions/api.md
```diff
@@ -110,8 +112,113 @@ implemented with separate instances of the `SimpleTagPattern` listed below.
...
+been added which users are encouraged to migrate to. The new inline processor
+is found at `markdown.inlinepatterns.InlineProcessor`.
+
+The new processor is very similar to legacy with two major distinctions.
+
+1. Patterns no longer need to match the entire block, so patterns no longer
+    start with `r'^(.*?)'` and end with `r'(.*?)!'`. This was a huge
+    performance sink and this requirement has been removed. The returned match
+    object will only contain what is explicitly matched in the pattern, and
+    extension pattern groups now starts with `m.group(1)`.
```
waylan: should be ``extension pattern groups now start with `m.group(1)`.``

You have an extra `s` on the end of `start`.

docs/extensions/api.md
```diff
@@ -110,8 +112,113 @@ implemented with separate instances of the `SimpleTagPattern` listed below.
...
+* **`SimpleTextInlineProcessor(pattern)`**:
+
+    Returns simple text of `group(2)` of a `pattern` and the boundaries of
+    where the data to replace resides.
+
+* **`SimpleTagInlineProcessor(pattern, tag)`**:
+
+    Returns an element of type "`tag`" with a text attribute of `group(3)`
+    of a `pattern`. `tag` should be a string of a HTML element (i.e.: 'em').
+    Also returns the boundaries of where the data to replace resides.
```
waylan: > returns the boundaries of where the data to replace resides.

I understand that last line, but if feels very awkward. I had to re-read it a couple times the first time I saw it. Same applies for the other few places it is used. Sorry, I don't have a better suggestion.

waylan (changes requested):
Thanks for the documentation. I noted a couple minor issues. Otherwise looks good

docs/extensions/api.md
facelessuser: Cool, will fix.

docs/extensions/api.md
facelessuser: Yeah, I'll rephrase it.

waylan (approved):

waylan:
@facelessuser, if you rebase this against master (resolving merge conflicts), I think we can accept this.

facelessuser:
Cool, I'll pull the latest changes into this pull request and resolve any conflicts.  I've pulled in most of the changes already, so imagine it should go fairly smooth.

facelessuser:
Branch is synced up.

Issue #575, opened 2017-08-08
Support custom block-level elements


akvadrako:
HTML has always worked fine with custom elements, but with the new [Custom Elements](https://www.w3.org/TR/custom-elements/) spec they have official support. We use some custom elements, some of which are designed for block-level displaying

Python-Markdown should allow configuring BLOCK_LEVEL_ELEMENTS. For now, I'm using this workaround:

```py

def patch_markdown():
  """Add additional block level elements."""
  orig = markdown.util.isBlockLevel

  def patched(tag):
    return orig(tag) or str(tag).startswith('x-')

  markdown.util.isBlockLevel = patched
```

waylan:
We've always been proud of the fact that anything is customization via our Extension API. However, this is actually something that is not easily changed. Even a subclass of the Markdown class woudn't address this. PR's are welcome against the md3 branch.

akvadrako:
The md3 branch looks awfully old - no changes for 2 years - are you sure that's the right one?

waylan:
Yep, that's the one. My time has been extremely limited the last few years. For the most part, I've only been reviewing pull requests for bug fixes. From my perspective, the library is feature complete and anything else is just nice-to-have. The md3 branch just makes the nice-to-have stuff easier.

akvadrako:
Okay, but I don't want to spend time on a pull request without expecting it to be part a release in the next couple months.

waylan:
That's fair, which is why I'm being upfront. However, version 2 is in a feature freeze (until version 3 is released) and we are only doing maintenance (bug fix) releases. So such a change would not be accepted on the version 2 branch (master).

PR #634, opened 2018-01-25
Update block level elements.

### LINKED ISSUE #543: HTML5 <details>
Having an issue where:

```markdown
<details>
<summary>Some summary</summary>

...content either md or html if separated by new line from html tags

</details>
```

breaks a [Sublime Text](https://www.sublimetext.com) [`mdpopups`](https://github.com/facelessuser/sublime-markdown-popups) phantom from the first `<details>` tag on. Everything before that is fine.

---

* compare facelessuser/sublime-markdown-popups#25

waylan:
Also refactor from regex to a list and add comments to explain why the
elements are in the list for future reference. Fixes #543.

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -28,15 +28,18 @@ Constants you might want to modify
 """
 
 
-BLOCK_LEVEL_ELEMENTS = re.compile(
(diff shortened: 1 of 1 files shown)


waylan:
This should probably also address #575 before being accepted.

PR #641, opened 2018-02-20
Better check of allowed TOC location

cbeauchesne:
These markdown should not raise an exception : 

    # *[toc]*

and
    
    [toc]
    # [toc]

Sorry if I miss something, I've tried a "blind" commit.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -137,11 +137,17 @@ class TocTreeprocessor(Treeprocessor):
 
         self.header_rgx = re.compile("[Hh][123456]")
 
-    def iterparent(self, root):
-        ''' Iterator wrapper to get parent and child all at once. '''
-        for parent in root.iter():
-            for child in parent:
-                yield parent, child
+    def iterparent(self, node):
+        ''' Iterator wrapper to get allowed parent and child all at once. '''
+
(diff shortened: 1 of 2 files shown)


markdown/extensions/toc.py
```diff
@@ -142,22 +142,30 @@ def iterparent(self, root):
         for parent in root.iter():
             for child in parent:
                 yield parent, child
+                
+    def iter_allowed_parent(self, node):
```
facelessuser: Why did you not just modify iterparent?  Now we have iterparent which is no longer used and then iter_allowed_parent.  This should be consolidated into one function, and if you have dead code, it should be removed.

cbeauchesne: Because it's public, and I was not sure that it's not used anywhere. But if it's not the case, it's better to simply replace it.

cbeauchesne: Fixed

facelessuser: If it isn't overriding a function in the base class, it probably isn't used externally.

facelessuser:
Should probably add a unittest to ensure this bug isn't regressed in the future.

cbeauchesne:
@facelessuser  : unittests added !

Issue #635, opened 2018-01-31
Empty first cell in a table


vincentbernat:
Hey!

With Markdown 2.6.8, I was able to have an empty first cell by using a non-breakable space. With 2.6.11, this doesn't work anymore. I can use a zero width non-breaking space instead but I fear I would forget it about after some time. Instead, I am using `<i></i>`. But maybe there should be a cleaner way?

```markdown
<i></i>       | 1 core  | 2 cores | 4 cores | 6 cores
------------- | ------- | ------- | ------- | -----------------
CPU 1, Core 1 | network | network | network | network + haproxy
CPU 1, Core 2 | haproxy | haproxy | haproxy | TLS
CPU 1, Core 3 | TLS     | TLS     | TLS     | TLS
CPU 1, Core 4 | -       | TLS     | TLS     | TLS
CPU 2, Core 5 | -       | -       | network | TLS
CPU 2, Core 6 | -       | -       | TLS     | TLS
CPU 2, Core 7 | -       | -       | TLS     | TLS
CPU 2, Core 8 | system  | system  | system  | system + haproxy
```

facelessuser:
There was a recent rework in to improve table handling. I guess one of the side effects is that I added calls to `strip()` in a couple of places, but that will strip out things like non-breaking spaces.  If all we want to strip is spaces, we could change it to a less aggressive stripping.  I'd be interested to hear @waylan's thoughts on it.

Edit: Specifically, we could just call `.strip(' ')` which would strip only spaces.

waylan:
@vincentbernat, have you tried using an HTML entity? As the output is HTML, you really should be using an HTML entity, not the Unicode character. I realize that you were using a *zero length* char, but what's wrong with a single width non-breaking space (`&nbsp;`)? The column will always be wider than that anyway due to the other rows.

> Specifically, we could just call `.strip(' ')` which would strip only spaces.

I'm of two minds about this. If someone is inserting something other than a normal space, then it is probably intentional and maybe we should leave it. Of course, there are those annoying cases where a user copy and pastes something not realizing they copied something other than a normal space char. In that scenario the current behavior is better. But the current behavior limits those who know what they are doing. However, users should be using HTML entities, not various Unicode chars... I can go back and forth all day.

I suppose in the end, as the current behavior was an unintentional behavioral change, we should probably change it back.

facelessuser:
Cool, I was kind of on the fence about it myself which is why I asked. The change is trivial, so I can throw together pull.

[...]

PR #644, opened 2018-02-22
Only strip spaces in tables

facelessuser:
Strip only the space character and not things like nbsp in tables. Ref #635


tests/test_syntax/blocks/test_tables.py
```diff
@@ -0,0 +1,43 @@
+# -*- coding: utf-8 -*-
```
waylan: As this is for an extension, let's put it at `tests/test_syntax/extensions/test_tables.py`.

waylan (changes requested):
Looks good.

tests/test_syntax/blocks/test_tables.py
facelessuser: Will do.

waylan (approved):

PR #659, opened 2018-05-10
toc.md baselevel doc fix

mustafa0x:


docs/extensions/toc.md
```diff
@@ -106,7 +106,7 @@ The following options are provided to configure the output:
 
     The `baselevel` setting allows the header levels to be automatically
     adjusted to fit within the hierarchy of your HTML templates. For example,
-    suppose the Markdown text for a page should not contain any headers higher
+    suppose the Markdown text for a page should not contain any headers lower
```
waylan, line 109: I'm not sure this change is an improvement. While the number specifying the level is "lower," the level in the hierarchy is "higher." Therefore, "higher" makes more sense to me. But I can also see how others might disagree. What language does the HTML spec use for this? Perhaps we should mirror that.

mustafa0x: Found this:

> These elements have a rank given by the number in their name. The h1 element is said to have the highest rank, the h6 element has the lowest rank, and two elements with the same name have equal rank.

[source](https://www.w3.org/TR/2011/WD-html5-author-20110705/the-h1-h2-h3-h4-h5-and-h6-elements.html)

waylan: Yes, that's matches my use. I suppose we could say "... any headers **with a higher rank** than level 3..." to clarify that it is the "rank" rather than the numerical value which is higher. But personally, I'm meh on the change from what we already have.

PR #662, opened 2018-05-20
Add missing comma in HTML_EMPTY

zyxw59:
Because of the missing comma, the last two tags in the list, `meta` and `param` are interpreted as a single string literal `"metaparam"`, so the serializer fails to correctly produce empty elements for `<meta>` and `<param>` tags.


waylan:
So it appears that the failing test is [caused][1] by the latest version of wheel not supporting Python 3.3. Otherwise this looks fine. Thanks.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/381219613#L590

waylan:
We probably should add a test for this.

PR #664, opened 2018-05-26
Add class to ul tag for css

ghost:
I added `class` in `ul tag` to make `toc` CSS design better.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -6,7 +6,7 @@ This parser handles basic parsing of Markdown blocks.  It doesn't concern
 itself with inline elements such as **bold** or *italics*, but rather just
 catches blocks, lists, quotes, etc.
 
-The BlockParser is made up of a bunch of BlockProcessors, each handling a
+The BlockParser is made up of a bunch of BlockProssors, each handling a
 different type of block. Extensions may add/replace/remove BlockProcessors
 as they need to alter how markdown blocks are parsed.
 """
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -6,7 +6,7 @@ This parser handles basic parsing of Markdown blocks.  It doesn't concern
 itself with inline elements such as **bold** or *italics*, but rather just
 catches blocks, lists, quotes, etc.
 
-The BlockParser is made up of a bunch of BlockProssors, each handling a
+The BlockParser is made up of a bunch of BlockProcessors, each handling a
 different type of block. Extensions may add/replace/remove BlockProcessors
 as they need to alter how markdown blocks are parsed.
 """
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -207,6 +207,7 @@ class TocTreeprocessor(Treeprocessor):
 
         def build_etree_ul(toc_list, parent):
             ul = etree.SubElement(parent, "ul")
+            ul.attrib["class"] = "tocstyle-%d" % toc_list[0]['level']
             for item in toc_list:
                 # List item link, to be inserted into the toc div
                 li = etree.SubElement(ul, "li")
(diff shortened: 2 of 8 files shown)


markdown/blockprocessors.py
```diff
@@ -6,7 +6,7 @@
 itself with inline elements such as **bold** or *italics*, but rather just
 catches blocks, lists, quotes, etc.
 
-The BlockParser is made up of a bunch of BlockProcessors, each handling a
+The BlockParser is made up of a bunch of BlockProssors, each handling a
```
waylan: This was correct to begin with. Now it is misspelled.

tests/test_extensions.py
```diff
@@ -573,8 +573,6 @@ def testRE(self):
 class TestTOC(TestCaseWithAssertStartsWith):
     """ Test TOC Extension. """
 
-    unittest.TestCase.maxDiff = None
```
waylan: Please don't remove this. I want to see a complete diff of any test failure when debugging a problem.

markdown/extensions/toc.py
```diff
@@ -207,6 +207,7 @@ def build_toc_div(self, toc_list):
 
         def build_etree_ul(toc_list, parent):
             ul = etree.SubElement(parent, "ul")
+            ul.attrib["class"] = "tocstyle-%d" % toc_list[0]['level']
```
waylan: I'm not crazy about the name of the class. Why do we need "style" in the class name. Yes, you might use the class as a styling hook, but you could also use it from JavaScript or for some other purpose. A more generic name seems better to me. Sorry, I don't have any suggestions.

waylan (changes requested):
Thanks for your submission. Although, I'm not 100% convinced that this is something we need. But if we do add it, it seems like the class should be configurable with a config option. Of course, a sensible default should be set. In addition to that, the following issues need to be addressed as well.

tests/test_extensions.py
ghost: It is what I added and it does not exist in the previous commit.

markdown/extensions/toc.py
ghost: My consideration was not enough. How about `toclist-*`?

tests/test_extensions.py
waylan: Setting [maxDiff][1] to `None` causes it to display the complete diff (unlimited length). By removing the line, `maxDiff` falls back to the default of 80*8 characters. Therefore, please restore the line so we have complete diffs.

[1]: https://docs.python.org/2/library/unittest.html#unittest.TestCase.maxDiff

markdown/extensions/toc.py
waylan: I suppose that works. Why not just `toc` on the top level only? The nested levels seem redundant to me.

ghost: Thank you for review. Also most of the cases are top level `tos` is fine. Sometime wants to adjust font-style of nested place.

waylan:
Previously I noted that:

> Although, I'm not 100% convinced that this is something we need. But if we do add it, it seems like the class should be configurable with a config option. Of course, a sensible default should be set.

I really think this needs to be a config option or not added at all. Personally, I would be okay with not adding it at all. So you want this to be added, I'll need to include a config option to set the name of the class.

ghost:
I want to this function but it isn't always necessary. If it's possible, would you add the config option  to set the name of the class?

waylan:
As I stated previously I have no personal need for this feature. Therefore, I have no motivation to spend any time on it. @bluelief if you would like this feature to be added, I'm willing to review it, but you need to do the work. As I previously stated, without this being a config option, I'm not adding it. As some months have passed, I'm closing this. If you update it with the requested changes, I'll consider reopening it.

Issue #671, opened 2018-06-19
Meta extension does not reset attribute


Glandos:
Hi,
Here is a test case:

```ipython
In [1]: from markdown import Markdown

In [2]: md = Markdown(extensions=['markdown.extensions.meta'])

In [3]: md.convert('''Title:   My Document
   ...: Summary: A brief description of my document.
   ...: Authors: Waylan Limberg
   ...:          John Doe
   ...: Date:    October 2, 2007
   ...: blank-value:
   ...: base_url: http://example.com
   ...: 
   ...: This is the first paragraph of the document.''')
Out[3]: '<p>This is the first paragraph of the document.</p>'

In [4]: md.Meta
Out[4]: 
{'title': ['My Document'],
 'summary': ['A brief description of my document.'],
 'authors': ['Waylan Limberg', 'John Doe'],
 'date': ['October 2, 2007'],
 'blank-value': [''],
 'base_url': ['http://example.com']}

In [5]: md.reset()
Out[5]: <markdown.Markdown at 0x7fc5a9b784e0>

In [6]: md.convert('''Title: Small doc
   ...: 
   ...: Content''')
Out[6]: '<p>Content</p>'

In [7]: md.Meta
Out[7]: {'title': ['Small doc']}

In [8]: md.reset()
Out[8]: <markdown.Markdown at 0x7fc5a9b784e0>

In [9]: md.convert('')
Out[9]: ''

In [10]: md.Meta
Out[10]: {'title': ['Small doc']}

```

A call to `reset()` does not remove the `Meta` attribute from the `Markdown` class. And since a call to `convert()` with an empty string does not call preprocessors (maybe it should?), the `Meta` attributes has invalid data.

waylan:
Interesting edge case. I specifically recall looking at this before and determining that each call to `convert` would overwrite the meta data. Apparently I hadn't considered the case of an empty document. When `convert` detects an empty document, it immediately returns and does not run anything. Of course, that means it doesn't overwrite the meta data either.

facelessuser:
What would need to happen is the `meta` extension should:

1. Register itself like the `footnotes` extension does: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/footnotes.py#L70

2. Provide a reset method that clears `md.Meta`: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/footnotes.py#L102.

3. A test case would also need to be provided.

[...]

PR #672, opened 2018-06-19
Implement reset() for Meta extension

Glandos:
This is a fix for #671

### DIFF
--- a/markdown/extensions/meta.py
+++ b/markdown/extensions/meta.py
@@ -36,10 +36,18 @@ class MetaExtension (Extension):
 
     def extendMarkdown(self, md, md_globals):
         """ Add MetaPreprocessor to Markdown instance. """
+        md.registerExtension(self)
+        self.md = md
         md.preprocessors.add("meta",
                              MetaPreprocessor(md),
                              ">normalize_whitespace")
 
+    def reset(self):
+        try:
+            del self.md.Meta
+        except AttributeError:
+            # Maybe it has already been deleted
+            pass
 
 class MetaPreprocessor(Preprocessor):
     """ Get Meta-Data. """
--- a/markdown/extensions/meta.py
+++ b/markdown/extensions/meta.py
@@ -43,11 +43,7 @@ class MetaExtension (Extension):
                              ">normalize_whitespace")
 
     def reset(self):
-        try:
-            del self.md.Meta
-        except AttributeError:
-            # Maybe it has already been deleted
-            pass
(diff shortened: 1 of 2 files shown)


markdown/extensions/meta.py
```diff
@@ -36,10 +36,18 @@ class MetaExtension (Extension):
...
         """ Add MetaPreprocessor to Markdown instance. """
+        md.registerExtension(self)
+        self.md = md
         md.preprocessors.add("meta",
                              MetaPreprocessor(md),
                              ">normalize_whitespace")
 
+    def reset(self):
+        try:
+            del self.md.Meta
```
facelessuser: No need to delete it with a try/except. Just set it to `self.md.Meta = {}`.

facelessuser:
Make sure to address the failing flake8 lint errors that are occurring on Travis-CI.  You can ignore the Py3.3 failures as that is being removed moving forward.

markdown/extensions/meta.py
Glandos: That's a good remark. However, there would be a discrepancy between the initial state (when there is no `Meta`) and after a call to `reset()`, even if there is no call to `convert()` first.

I am not used to this library, but I'd expect a `reset()` method to remove every useless parts that are inexpensive to rebuild.

facelessuser: Fair enough.  It looks like the resets don't get called when the class is initialized.  We could just set `md.Meta = {}` in `extendMarkdown` and have a consistent interface that doesn't require deleting the `Meta` attribute.  I think that would be my personal preference, but I'll let @waylan make the final decision.

facelessuser: Never mind, reset does get called on initialization.  I personally think it makes sense to initialize `Meta` as `{}`.  Then you don't have to remember if you set reset or not.  It's filled when it has something important, and empty when it doesn't (reset being one of these cases). waylan will let you know what he prefers, but that is my preference.  I think meta should initialize all of the variables it is going to use (in this case `Meta`).

waylan: I agree with @facelessuser. Set `Meta` at `{}` in `reset`. `reset` gets called when the instance is first created, so that will ensure that the attribute always exists and always contains a dict.

Glandos: Done in 7f8ce8a

facelessuser:
@Glandos , LGTM.  Thanks for your work on this!

PR #688, opened 2018-07-25
Replace homegrown OrderedDict with purpose-built Registry.

### LINKED ISSUE #418: Replace OrderedDict with prioritized List?

waylan:
All processors and patterns now get "registered" to a Registry.
Each item is given a name (string) and a priority. The name is for
later reference and the priority can be either an integer or float
and is used to sort. A Registry instance is a list-like iterable with
the items auto-sorted by priority. If two items have the same priority,
then they are listed in the order there were "registered". Registering
a new item with the same name as an already registered item replaces
the old item with the new item (however, the new item is sorted by
its priority). To remove an item, "deregister" it by name or index.

Fixes #418.

Note that this is an adaptation of #510 which has been rebased onto master.


waylan:
@facelessuser, @mitya57 and any other extension authors, I would appreciate some feedback on this. Specifically on how I'm handling backward compatibility. The reasons for the change were discussed extensively in #418 and the implementation is documented in the [doc strings][1].

After getting a working implementation (the first commit), I worked on a backward compatible shim so that existing extensions can continue to work until we remove the shim in some future version. The shim won't be able to accommodate some of the more complex stuff, but I believe it should address the needs of most basic extensions. I'm asking if there is anything I missed.

# Supported Methods

I reviewed the existing documentation for our homegrown OrderedDict and implemented the following methods:

* ## add(name, item, location)

  As a reminder, `location` can be one of `_begin`, `_end`, `<name` or `>name`. 

  For any location which means 'insert new item at beginning/end", the new item is assigned a priority 5 more/less than the first/last existing item, respectively. 

  For any location which inserts an item between existing items, the priority of those existing items are obtained, and the new item is assigned a priority half-way between them.

  If no items exist in the collection (the first item is being added), then the new item is given a priority of `50`.

  A `DecrecationWarning` will be raised when the `add` method is called.

* ## delete item

  You can do `del collection['name']` to remove an item from the collection. This simply passes the name to the `deregister` method and then a `DecrecationWarning` will be raised.

* ## set item

  You can do `collection['name'] = NewItem()` to add an item to the collection. The new item is given a priority five less that the last existing item. If no items exist in the collection (the first item is being added), then the new item is given a priority of `50`.  After adding the item, `DecrecationWarning` will be raised.

# Unsupported Methods

In addition to the above, there are a few documented methods which I did not re-implement, which are discussed below:

* ## link(name, location)

  The `link` method of our homegrown `OrderedDict` allowed the position of an existing item to be moved to a new location using the location syntax of the `add` method. The new priority based registry does not include a method for changing priority, therefore it doesn't make sense to include this method. Is that a problem? If so, then we should probably implement the method for changing priority and then implement this as a deprecated wrapper. Thoughts?

* ## Various other methods.

  The documentation mentions that various other methods are available common to lists and dicts which can be used to manipulate instances of an `OrderedDict`. Specificaly, `keys()`, `values()`, `items()`, `insert()`, `append()`, and `index()` are mentioned by name, although usage is not demonstrated. However, as those methods are very general purpose and not specific to the intended use case, they are not implemented on the new registry and I see no value in providing a shim for them. If an existing extension is using these methods, then it is complex enough that there likely won't be a simple migration path and the extension is not likely to work in the new release of Markdown anyway. Thoughts?

[1]: https://github.com/Python-Markdown/markdown/pull/688/files#diff-0f2fbb4d1e47771e120a801756331eceR191

facelessuser:
I will provide feedback as soon as I am able to evaluate this, thanks.

mitya57:
The change and the backwards compatibility shim look good to me. It is fine to remove some methods — after all the next version is 3.0 so some breakage is expected. Anyway, my third-party extensions' tests pass fine with this branch.

waylan:
> after all the next version is 3.0 so some breakage is expected.

I agree, however, I believe our most valuable asset is the rich collection of extensions (both first and third party). If we cut off all existing third party extensions, we may find people will migrate to a competing lib rather than update their extensions. Therefore, I'd like to make the transition as painless as possible. To be clear, I am not suggesting that we should avoid the changes, simply that the transition is smooth.

It is great to hear that your "third-party extensions' tests pass fine with this branch." That's what I was aiming for.

facelessuser:
I'm going to need some time to sort through all of my errors with the 3.0.dev merge.  I'm getting a number of failing tests, but I'm not sure if any are related to this specific change.

mitya57:
You are right, extensions are a very great feature of Python-Markdown :)

To be more precise, I tested my [math] and [posmap] extensions. The latter was affected by another change in master, see retext-project/retext@37923cb1b74eb813714854503990b8a204b5610f where I fixed it. That extension is a great hack so I think it's normal that it breaks.

@facelessuser has a much larger collection of extensions, so results of his experiments would be more interesting.

[math]: https://github.com/mitya57/python-markdown-math
[posmap]: https://github.com/retext-project/retext/blob/master/ReText/mdx_posmap.py

facelessuser:
Yeah, I have a number of issues.  Two related to this change `AttributeError: 'Registry' object has no attribute 'link'`. And `AttributeError: 'Registry' object has no attribute 'keys'`.

Various other changes are causing failures as well, like link inline patterns no longer having `sanitize_url`, which is probably no big deal if I remember correctly...I think by default it did nothing?  I'll have to dig into it.

`ModuleNotFoundError: No module named 'markdown.odict'`, this is to be expected.

Changes in stash: `TypeError: store() got an unexpected keyword argument 'safe'`.

A couple of syntax failures which are most likely related to one or more of the above issues.  So I have some things to work through to better understand the new changes.

waylan:
> Yeah, I have a number of issues.  Two related to this change `AttributeError: 'Registry' object has no attribute 'link'`. And `AttributeError: 'Registry' object has no attribute 'keys'`.

Interesting. These are exactly the sorts of things I am look for. WHen you get a chance, some more info on those would be helpful.

facelessuser:
As far as the two errors concerning the Registry, the first (`AttributeError: 'Registry' object has no attribute 'link'`) deals with overrides of existing keys.  I was overriding a number of the emphasis keys and injecting new ones.  This required me to move an exiting  key.  `link` made that easy, but is now gone requiring me to do it manually now:

```py
        if Registry is not None:
            value = md.inlinePatterns["em_strong"]
            md.inlinePatterns.add("em_strong", value, ">strong_em2")
        else:
            md.inlinePatterns.link("em_strong", ">strong_em2")
```

This one might be worth addressing.  I actually get burned by this one in a couple of extensions.  There are some that I need to ensure are after certain extensions, and I can't really determine this until a reset because that is the only time all extensions are loaded, so I will "link" if I don't meet that criteria during reset.

The second error (`AttributeError: 'Registry' object has no attribute 'keys'`) was simply due to me treating the registry as an Ordered Dict, which it used to be.  I was simply searching for an existing key to determine if an extension was loaded.  I don't have to use `'ext' in registry.key()` to compare a key though, I could have just as easily just done `'ext' in registry`.  Not exactly sure why I specifically called `key`.

This second one isn't really a something I think you need to specifically address.

I need to finish cleaning up

waylan:
I have never needed to use `link` myself. So it is good to see that it has a useful application in the real world. We can provide a replacement (and temporary shim) for it. Any suggestions for a name? As the new method will be changing priority (rather than position), I'm not sure "link" makes sense. 

Maybe "update"? Perhaps if `update` receives a "name" and "priority" it changes the priority; if it receives a "name" and "item" is replaces the item preserving the priority; and if it receives all three it updates both the "item" and "priority". `update` should error if the provided name does not exist.

facelessuser:
Yeah, `update`, `move`, `relocate`, anything like that would be fine. I'm fine with `update`.

facelessuser:
Turns out there is no in place replace.  I have to hack an in place replace as the behavior is now unintuitive.  It always inserts 5 less than the last.

```python
        if util.MD3:
            md.inlinePatterns._data['autolink'] = auto_link_pattern
        else:
            md.inlinePatterns['autolink'] = auto_link_pattern
```

waylan:
Took me a minute to understand what you meant, but I think I got it. Previously, when using setitem on an existing key, the position information was not altered (the item was simply replaced in-place). That doesn't happen with the shim. The shim assumes a new item is always being added. Yep, that's a bug.

facelessuser:
Yeah, I'm still playing around with it, I'm going to explore a bit more with the moving of an item.  There were some things that didn't quite work like I expected them.  This may have been due to a misunderstanding on my part, but I'm not sure.

Originally I was using the legacy notation of moving an item with `>` or `<` to get right after a certain item.  That worked (the tests passed), but I tried to move away from that with utilizing `get_index_for_name` and then incrementing that for priority.  While the tests worked there, I realized that the index returned is its position in the list, not its priority.  I then used that index to extract the priority from `self._priority` and incremented that and registered the item with the new priority and tests failed. This seemed odd.

There are still some things for me to figure out and work out.  I'm still trying to figure out what went wrong as what I was doing seemed like it should have worked. It may be my own mistake/misunderstanding, but I am still investigating.

waylan:
Right, there is no public way to get an item's priority. Not sure why we need that. I was trying to get away from using any existing items position to calculate a new position programicaly. Each item's priority should be a known constant which never changes. In fact, I expect that is why I originally did not provide a mechanism for changing an item's priority. 

Do you have a situation where Item A needs a lower priority that item B in some situations but needs a higher priority than item B is other situations? If not, then you should never need to change an item's priority.

As a reminder, each priority in the Markdown lib is incremented by 10. So if you need to insert multiple items between two existing items, you have 9 possible whole numbers and an infinite number of floats to use as priorities. If you have two items to insert, simply give item A a priority higher than item B within that range. Even if the items are loaded from two separate extensions, they are sorted correctly regardless of which is loaded first. As each priority is a known constant, you can give each item a priority which works with all other known items. This breaks when you start changing priorities.

In fact, after thinking on this some more, I'm thinking perhaps we should not allow priorities to be changed.

waylan:
Let me illustrate. Suppose the Markdown lib provides 2 items: A & B. Multiple extensions need to insert items between A & B, but due to conflicts between those extensions, order matters among them. The old way required examining what was already there and altering location accordingly, which was rather cumbersome. 

Instead, each item has a priority assigned to it as a known constant, such as:

```
A (priority=20)
ext_a (priority=18)
ext_b (priority=15)
ext_c (priority=12)
B (priority=10)
```

Whether all, some or none of those extensions are loaded by the user, they will always be sorted in the correct order, irrespective of the order that they are loaded. If a new extension comes along which needs to be between ext_a and ext_b, then that new item should simply be assigned a priority somewhere between 15 and 18. And as every priority is a known constant, you don't need to worry about that changing and breaking yours.

waylan:
I fixed the bug where `__setitem__` was altering the priority of an existing name when a new item was assigned to that name. This is the only place that can happen. The expectation is that extension devs will update their code to use the `register` method and define a priority, which may or may not be the same as the item they are replacing.

> What I was trying to illustrate was that when I acquired the priority of an existing extension, which I think turned out to be 30 (which matched the hard coded value), I tried to use 31 as the priority for my extension, and it didn't work as expected.

So the item with priority 31 did not get sorted to be before the item with priority 30? Or did you mistakenly think that 31 comes after 30. I've made that same mistake a couple times. We are used to lower numbers coming before higher numbers. But, by their nature, priorities get sorted with higher numbers coming first. It is an easy mistake to make.

However, if you got it right, I'd like to see more information about this. How did you assign the priority? How did you determine the behavior was different that expected?

Btw, sorting is handled internally. The private `_sort` method is called by any public method before that methods does any operation that requires items to be sorted. However, the code is optimized to call `_sort` as few times as possible. Therefore, any calls to private methods and/or attributes are likely to return unsorted results.

facelessuser:
> Right, there is no public way to get an item's priority. Not sure why we need that.

I understand, and to be honest, I'm fine if we don't have an official "move" API method.  As long as there is a deregister and register, a move can be made (and I don't think we should get rid of deregister 🙂).

> Each item's priority should be a known constant which never changes.

Yes, as each item has a known priority, technically you can, in most cases, just provide a value that should account for other known priorities.

There may be occasions where you want to insert an extension A with priority X in default cases, but with priority Y if extension B is present.  My specific extension can be refactored where I don't need to this, so this is not an immediate concern, but I'll illustrate what my real concern is.

I'm not concerned if we have a "move" command.  I'm more concerned with unexpected behavior with priorities.  What I was trying to illustrate was that when I acquired the priority of an existing extension, which I think turned out to be `30` (which matched the hard coded value), I tried to use `31` as the priority for my extension, and it didn't work as expected.  This might be due to my misunderstanding or improper use of the API, but I just need to confirm what the issue is.  I'm more concerned with whether or not the priority insertion is intuitive.  I'm not sure if there is special logic when you insert at the tail or not.  Or if there are cases where the list is expected to be sorted, but doesn't get sorted. I just need to make sure this is working as expected.

waylan:
> What I was trying to illustrate was that when I acquired the priority of an existing extension, which I think turned out to be 30 (which matched the hard coded value), I tried to use 31 as the priority for my extension, and it didn't work as expected. 

Can you expound upon that? Are you saying you got confused because sort order is reversed (higher numbers sort before lower numbers -- I've make that mistake a couple times), or are you saying that the higher priority item did come after the lower priority item (which would be a bug)?

Btw, sorting is handled internally and optimized to only run when it is actually needed. Therefore if you access any private methods or attributes, you are likely to break the sorting unless you fully replicate the internal sorting API. In other words, private methods and attributes are actually private and shouldn't be used.

As an aside, I also fixed the bug were `__setitem__` was not preserving location when a new item was assigned to an existing name. In other words, `md.inlinePatterns['autolink'] = auto_link_pattern` should work correctly now. Interestingly, this is the only place where we preserve an existing priority, but the expectation is that the extension author should update that code to:

```python
if util.MD3:
    md.inlinePatterns.register(auto_link_pattern, 'autolink', 120)
else:
    md.inlinePatterns['autolink'] = auto_link_pattern
```

The support for the old way is only to not break extensions until extension authors have an opportunity to make such an update. Of course, the more complex extensions will be broken, but that is unavoidable unfortunately.

facelessuser:
Understood.  I'm evaluating in my free time.  Hopefully I'll have a better understanding in the next day or so.

I've addressed most of the 3.0dev issues, but there are still some more.

facelessuser:
I had the priority backwards.  That **is** a little confusing.  I didn't realize that smaller was after and bigger is before.

facelessuser:
On a side note, looks like the 3.0dev branch doesn't recognize `style` as a block element.

facelessuser:
I think I'm done evaluating the 3.0dev branch.  Your recent fixes resolved most of my other issues.  I'll create an issue that tracks `style` being mistaken for an inline element.

waylan:
> I had the priority backwards. That is a little confusing. I didn't realize that smaller was after and bigger is before.

Yeah, I've made the same mistake a few times myself. But that is how these sorts of libs typically work. A "higher priority" means a "higher number". I just keep that expressions in mind: "higher priority" and "lower priority" and then it makes sense.

PR #695, opened 2018-07-31
Remove lazy_ol keyword. Use sane_lists extension instead.

waylan:
This was adapted from 11408e50 of the md3 branch.


mDuo13:
My code ([Dactyl](https://github.com/ripple/dactyl/)) used `lazy_ol=False` and was silently broken in a small way for almost 2 years because this change was—as far as I can tell—not documented anywhere outside of this PR. Please make an effort to mention things like this in the release notes in the future, and maybe raise an error when providing a removed option.

Issue #660, opened 2018-05-10
Footnote ref number in TOC


mustafa0x:
```python
In [13]: t = '''
    ...: # Header with footnote[^1]
    ...: 
    ...: Lorem Ipsum
    ...: 
    ...: [^1]: footnote text
    ...: '''

In [14]: print(markdown.markdown('[TOC]\n\n' + t, extensions=['markdown.extensions.toc', 'markdown.extensions.footnotes']))
```

```html
<div class="toc">
<ul>
<li><a href="#header-with-footnote1">Header with footnote1</a></li> <!-- Note the '1' in the header id and more importantly the link text -->
</ul>
</div>
<h1 id="header-with-footnote1">Header with footnote<sup id="fnref:1"><a class="footnote-ref" href="#fn:1" rel="footnote">1</a></sup></h1>
<p>Lorem Ipsum</p>
<div class="footnote">
<hr />
<ol>
<li id="fn:1">
<p>footnote text&#160;<a class="footnote-backref" href="#fnref:1" rev="footnote" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

waylan:
So the [code][1] which sanitizes the text for use in the TOC is pretty simple. It simply pulls the text from the HTML elements. It could be significantly more complex to exclude footnote refs. And I find it odd that we would need to only do this for a non-standard add-on syntax. Additionally, the fact that this is only being reported now suggests that this is an unusual edge case that not many users will encounter.

That said, it is clearly not what one would expect and should probably be fixed. Of course, pull requests are welcome.

[1]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/toc.py#L237

Issue #699, opened 2018-08-15
is there a way to render math in toc?


Ir1d:
Hi, I was trying to include math formula in toc, and I found that this feature was not included.

For example: `### $a_i$` would result in `a_ia_i` in the toc.

I opened a issue [here](https://github.com/squidfunk/mkdocs-material/issues/849) and realised that toc also doesn't support bold or italicize.

Wondering if this feature can be supported or is there a way to hack into this and include math in toc on my own?

Thx in advance.

waylan:
This is directly related to #660. Our sanitizer is pretty dumb about how and what it removes from TOC items. It strips out everything except plain text.

There are some things we definitely want to have stripped (links for example), but it is not always clear. For example, I might want italics in my header, but not in the TOC item for that header. Whereas another person might want it preserved in both. And something like math is only supported through third party extensions. I would not expect the TOC extension to know how to handle math unless it provided a hook for the math extension to inform it. And I'm not sure what that hook might look like.

We could have a whitelist of elements to leave in. Or a blacklist of elements to strip. Ether way, in most cases when we strip an element, we preserve the text content of the element. However, in the case of #660, we not only need to strip the element tags, but also the content. The trick there is that we shouldn't just strip all `<sup>` elements, but only the ones which point to a footnote. That is a little more complicated than a simple list of elements regardless of whether it is a whitelist or a blacklist.

waylan:
IIRC, MultiMarkdown provides a more sophisticated way of handling this stuff, and they support math. Or maybe its Pandoc that I'm thinking of. Perhaps we should investigate their behaviors.

waylan:
@Ir1d would #700 be a reasonable solution for you? You can find the documentation [here](https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes).

I realize it is more of a workaround for you as you would need to manually define the text of the TOC item, but there are valid use cases for that outside of yours. Specifically, it solves the problem I have, which is demonstrated in the example in the docs.

The one limitation is that it only accepts plain text (Markdown syntax does not bet parsed). However, I believe that would not be a problem for you as your math should get passed through unaltered to be processed by a JS lib in the browser.

```md
### $a_i$ { #a_i data-toc-label="$a_i$" }
```

would result in the following TOC item:

```html
<li><a href="#a_i">$a_i$</a></li>
```

facelessuser:
[...]

PR #700, opened 2018-08-17
Support custom labels in TOC.

waylan:
When the Table of Contents Extension is used in conjunction with the Attribute Lists Extension  and a `data-toc-label` [attribute][1] is defined on a header, the content of the  `data-toc-label` attribute is now used as the content of the Table of Contents item for that header.

[1]: https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes

### DIFF
--- a/docs/change_log/release-3.0.md
+++ b/docs/change_log/release-3.0.md
@@ -202,6 +202,12 @@ The following new features have been included in the release:
 * A new `toc_depth` parameter has been added to the
   [Table of Contents Extension](../extensions/toc.md).
 
+* A new `toc_tokens` attribute has been added to the Markdown class by the
+  [Table of Contents Extension](../extensions/toc.md), which contains the raw
+  tokens used to build the Table of Contents. Users can use this to build their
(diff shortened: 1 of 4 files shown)


facelessuser:
Interesting workaround for sane tocs with complex headers.

waylan:
Yeah, I've had the idea bouncing around for a while, but was thinking of using the `title` attribute. I wasn't crazy about it though as there is a legitimate use for titles and they are potentially visible to end users (and screen readers). Today it occurred to me that we could use a `data-*` attribute. And then I had it working with only one line of code. Writing the documentation was the hardest part.

We should probably still properly fix #660, although this would serve as a workaround in the meantime. Note sure if there is anything else we could do for #699 though.

waylan:
The last commit (94b9c31) just added a another (unrelated) feature to TOC:

A new `toc_tokens` attribute has been added to the Markdown class, which contains the raw tokens used to build the Table of Contents. Users can use this to build their own custom Table of Contents rather than needing to parse the HTML available on the `toc` attribute of the Markdown class.

PR #705, opened 2018-08-31
Pulling apart markdown.core.convert into two functions.

patjenk:
Breaking the part of markdown.core.convert that creates an xml.etree.ElementTree into its own method. This will allow the library to function more like a compiler front end and expose access to the intermediate representation and hopefully enable developers to augment or manipulate the markdown in intelligent ways.

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -247,31 +247,7 @@ class Markdown(object):
         5. The output is written to a string.
 
         """
-
-        # Fixup the source text
-        if not source.strip():
-            return ''  # a blank unicode string
-
-        try:
-            source = util.text_type(source)
-        except UnicodeDecodeError as e:  # pragma: no cover
-            # Customise error message while maintaining original trackback
-            e.reason += '. -- Note: Markdown only accepts unicode input!'
-            raise
-
-        # Split into lines and run the line preprocessors.
-        self.lines = source.split("\n")
-        for prep in self.preprocessors:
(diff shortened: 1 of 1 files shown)


waylan:
Thank you for your contribution, but I see no good reason to make this change at this time. We already have the plugin API which gives users complete access to the ElementTree object. What additional functionality does this provide?

My assumption is that you are trying to replicate the `[source] => parser => [AST] => renderer => [html]` model. The problem is that the ElementTree object is not a standalone, complete representation of the document (like an AST is). The document will be broken without the postprocessors which only run after the ElementTree object is serialized to a string (for an example see the HtmlStash). In other words, the entire library is a single unit and the individual parts cannot stand alone.

The first thing that would need to happen is that all of the preprocessors should to be eliminated (mostly by rewriting their functionality as blockprocessors). Once that happens, most of the postprocessors will be unnecessary and the few remaining can be easily reworked in the document tree. With all of the parsing happening in the core (blockprocessors, treeprocessors and inlineprocessors), we can finally consider the sort of change proposed here. However, until then it would just create an extra support burden to make this change as users will assume they can use one or the other of the two parts and will be frustrated when things are broken/incomplete. Sure, it might be sufficient for your use case, but it won't work for everyone and I'm not willing to take on that extra support burden.

To be clear, I'm not opposed to the end goal, its just that this change needs to be the last step in working toward that goal, not the first.

PR #709, opened 2018-09-22
Bring back the old version variable

ghost:
To fix an AttributeError for backward-compatibility:

```
if markdown.version <= '2.2':
AttributeError: module 'markdown' has no attribute 'version'
```

In this case from django-rest-framework.

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -56,4 +56,6 @@ def _get_version():  # pragma: no cover
 
 
 __version__ = _get_version()
+
+# Also support `version` for backward-compatabillity with <3.0 versions
 version = __version__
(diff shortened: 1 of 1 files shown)


waylan:
I'm really annoyed by this. I only want to have the `__version__` variable, but a version variable name that is not consistent across versions is not very useful. Sigh.

Thanks for the patch.

ghost:
Yeah, I feel your pain. My first reaction was to submit a PR to django-restframework, but then realized this will be a common problem with other code that uses Python-Markdown. The backwards-compatibility woes of a popular library. ;-)

mitya57:
Note that Python-Markdown 2.6.11 had a `version_info` attribute too (as opposed to `__version_info__` in 3.0). There is (at least some) third-party code that uses it too:

https://codesearch.debian.net/search?q=markdown.version_info

PR #713, opened 2018-09-25
Make ENTITY_RE support hexadecimal entities

### LINKED ISSUE #712: Hexadecimal HTML entities are not rendered
According to standard Markdown, HTML entities are to be recognized and rendered correctly (https://daringfireball.net/projects/markdown/syntax#autoescape). Up to v2.6.11, Python-Markdown has been able to do so, but with v3 it does not work any longer.

A hex HTML entity is not recognized. The leading ampersand is rendered as "`&amp;`", and in HTML, the entity consequently reads as the entity itself, like, for example `&#x25a0;` instead of the intended ■.

mitya57:
Fixes #712.

### DIFF
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -155,8 +155,8 @@ AUTOMAIL_RE = r'<([^> \!]*@[^> ]*)>'
 # <...>
 HTML_RE = r'(\<([a-zA-Z/][^\>]*?|\!--.*?--)\>)'
 
-# &amp;
-ENTITY_RE = r'(&(?:\#[0-9]+|[a-zA-Z0-9]+);)'
+# "&#38;" (decimal) or "&#x26;" (hex) or "&amp;" (named)
+ENTITY_RE = r'(&(?:\#[0-9]+|\#x[0-9a-fA-F]+|[a-zA-Z0-9]+);)'
 
 # two spaces at end of line
 LINE_BREAK_RE = r'  \n'
--- a/markdown/serializers.py
+++ b/markdown/serializers.py
@@ -53,7 +53,7 @@ __all__ = ['to_html_string', 'to_xhtml_string']
 
 HTML_EMPTY = ("area", "base", "basefont", "br", "col", "frame", "hr",
               "img", "input", "isindex", "link", "meta", "param")
-RE_AMP = re.compile(r'&(?!(?:\#[0-9]+|[0-9a-z]+);)', re.I)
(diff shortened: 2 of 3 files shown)


facelessuser:
This should be fixed in the serializer as well.

mitya57:
@facelessuser Can you point where? My change is enough to make the added tests pass.

facelessuser:
Here https://github.com/Python-Markdown/markdown/blob/master/markdown/serializers.py#L56

facelessuser:
Basically we want to avoid covering already converted hex entities. It may get there via a different path.

mitya57:
Fixed. If you give me a simple example of a “different path”, I will add a test for it.

facelessuser:
Sure, I'll give you one later today. I'm currently on my phone while getting my kids ready for school :smile:.

facelessuser:
@mitya57, I found a similar case here: https://github.com/Python-Markdown/markdown/pull/670/files#diff-c12f46c56418674a5a9b8566477cab40R122.  Basically, we are trying to prevent entities from getting double escaped, so things like link source go through the serializer, and we just check that we aren't escaping a valid entity.

mitya57:
Nice, I have amended that test now. If you approve this PR, I will squash and merge it.

facelessuser:
Looks like it is already merge, but looks good :smile: .

mitya57:
Oops, I mistakenly pushed to master instead of this branch, without squashing :(

Anyway it fixes the bug, and there are tests, so I hope it's fine.

waylan:
I've noticed that a few PRs have been created by pushing to branches on the Python-Markdown/markdown repo. Going forward, let's create branches on the our private forks and create PRs from those. I have been doing that myself without problems for some time on the mkdocs project and it helps avoid mistakes like this.

Note that I still occasionally need to push something upstream (like tags), but git allows configuration of multiple "remotes". The thing is, you have to specify any remote except the default and as long as the default remote is your personal fork, there's less chance of messing up the upstream repo. Rebasing, squashing, etc on your personal fork should be less of an issue as you're not likely to have people forking from it.

mitya57:
I have now done this:
```
$ git remote add my git@github.com:mitya57/Python-Markdown.git
$ git fetch my master
$ git branch --set-upstream-to=my/master master
```

Now `git push` on master branch without arguments will push to my fork. Thanks @waylan for the advice!

waylan:
That works. Personally, I use the workflow described in <https://help.github.com/articles/working-with-forks/>.

PR #726, opened 2018-10-07
Fix issue with < and > characters (among others) being double-escaped.

### LINKED ISSUE #725: codehilite extension double escapes HTML
Here's an example:

    import markdown
    
    example_content = """# Test 
      
        >>> print('hi') 
        hi 
    
    The above is valid MarkDown."""
    
    output = markdown.markdown(example_content, extensions=['markdown.extensions.codehilite'])
    print(output)

As you can see, my example file includes a Markdown example with > characters. Unfortunately, with Codehilite enabled, it produces doubly-escaped content. Here's the output of that script:

```
<h1>Test</h1>
<div class="codehilite"><pre><span></span>&amp;gt;&amp;gt;&amp;gt; print(&#39;hi&#39;) 
hi
</pre></div>


<p>The above is valid MarkDown.</p>
```

See those `&amp;&gt;` parts? That means your HTML output includes the > characters escaped twice. It should be just `&gt;&gt;&gt;`, not `&amp;gt;&amp;gt;&amp;gt;`.

kkinder:
This should fix #725.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -76,6 +76,10 @@ class CodeHilite(object):
     def __init__(self, src=None, linenums=None, guess_lang=True,
                  css_class="codehilite", lang=None, style='default',
                  noclasses=False, tab_length=4, hl_lines=None, use_pygments=True):
+
+        # Kind of dirty, but un-escape HTML since we re-escape it anyway.
+        src = src.replace('&amp;', '&').replace('&lt;', '<').replace('&gt;', '>').replace('&quot;', '"')
+
         self.src = src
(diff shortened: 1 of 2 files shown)


facelessuser:
I suspect the unescaping should happen in the treeprocessor, not in the codehilite object. Extensions like fenced code will not have escaped the content before calling, while the code block processors will have escaped the content.

facelessuser:
@kkinder, I went ahead and created a pull that properly addresses the issue without breaking `fenced_code`. I will close this issue in favor of #727.  Thanks for pointing out this issue! We didn't actually have tests in place for this, so it got missed.

PR #727, opened 2018-10-07
Fix double escaping of block code

facelessuser:
Ref #725. When post processing code blocks handled by the `code` step, unescape previously escaped entities. Make sure to test both `fenced_code` and `codehilte` to ensure the output of both are the same in relation to entities.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -200,13 +200,20 @@ class CodeHilite(object):
 class HiliteTreeprocessor(Treeprocessor):
     """ Hilight source code in code blocks. """
 
+    def code_unescape(self, text):
+        """Unescape code."""
+        text = text.replace("&amp;", "&")
+        text = text.replace("&lt;", "<")
+        text = text.replace("&gt;", ">")
+        return text
+
     def run(self, root):
         """ Find code blocks and store in htmlStash. """
         blocks = root.iter('pre')
         for block in blocks:
             if len(block) == 1 and block[0].tag == 'code':
                 code = CodeHilite(
-                    block[0].text,
+                    self.code_unescape(block[0].text),
                     linenums=self.config['linenums'],
                     guess_lang=self.config['guess_lang'],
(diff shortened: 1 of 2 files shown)


markdown/extensions/codehilite.py
```diff
@@ -200,13 +200,20 @@ def _parseHeader(self):
...
+        return text
+
     def run(self, root):
         """ Find code blocks and store in htmlStash. """
         blocks = root.iter('pre')
         for block in blocks:
             if len(block) == 1 and block[0].tag == 'code':
                 code = CodeHilite(
-                    block[0].text,
+                    self.code_unescape(block[0].text),
```
waylan, line 216: Are we sure this doesn't unescape code that the user manually escaped? For example a code block with HTML entities in it:

    Example:

        &
        &amp;
        &amp;amp;

The above code block should some out the other end correctly.

waylan: We might want to add a test for this.

facelessuser: Because this is code, it will unescape all `&amp;` `&lt;` and `&gt;` because in the `code` step, we escape all `&`, `<`, and `>`.  It is done in the code step because we want literal representations of anything in the code region. We do not preserve entities, but the characters. The serializer was updated to avoid already escaped entities, code does *not* do this which is why we do it in the code step.

We only do this to code captured via the `code` step which is why we have this occurring in the `HiliteTreeprocessor` and not in the `CodeHilite` class itself.  Fenced code will capture the code and process it immediately, so no unescaping is needed.  Only indented blocks are handled in two phases requiring an unescape.

With all of that said, I will gladly add the requested test below.  We can expect that each `&` will be expanded to `&amp;` in the final output regardless of whether it is part of an entity or not as what we are looking for is preserving the characters.

```
    &
    &amp;
    &amp;amp;
```

pquentin:
Thank you! This fixed a bug I had on my pelican blog.

Issue #723, opened 2018-10-04
Footnotes separator logic needs revision for 3.0


avaris:
Since `output_format` is now limited to `html` or `xhtml`, [this `if` branch](https://github.com/Python-Markdown/markdown/blob/3.0.1/markdown/extensions/footnotes.py#L143-L146) in footnotes extension will never be True. Therefore a different output is generated compared to the previous versions if you had `html5` or `xhtml5`, and that is causing test errors (see getpelican/pelican#2414). 

It is not clear to me which of the separator options is preferable now, so I'm hesitant to adjust the test outputs in pelican prematurely.

waylan:
The alternate separator behavior seems to have been introduced as a solution for #180 in 41cc0555. I had completely forgot about this when reducing the output formats (in #633). For the record, after 41cc0555 any formats prior to HTML5 continued to use the old format (a colon as a separator in the footnote ids), while HTML5 switched to using a hyphen as the separator.

Of course, with the 3.0 release, we reduced the output formats to `html` and `xhtml` with the understanding that both formats output HTML5 (in the HTML or XHTML flavors respectively). Therefore, the footnote ids should presumably be using the hyphen as a separator.

The problem is that that is not backward compatible for users who either intentionally chose pre-HTML5 formats or who previously were not picky and used the non-version specific format (which used to default to a pre-HTML5 format). In other words, whichever way we go, we will break things for half of our users. So the question is: which half do we break things for?

waylan:
Thinking about this further I am reminded that the footnotes extension had always been intended as a clone of PHP Markdown Extra's footnote behavior. In fact, the one notable difference was that we used a hyphen rather than a colon as a separator in the footnote ids for HTML5 output. As PHP Markdown does not have an HTML5 specific output format, that has never been an option for them. Additionally, there is nothing in the HTML5 spec which prevents a colon from being part of an id. Therefore, I am inclined to leave the behavior as-is, that is, only use a colon as the separator.

However, it is reasonable to expect that some users will want the hyphen rather than the colon in their output. It is even possible that a few users have specifically chosen one of the HTML5 specific output formats for this reason alone. Therefore, it makes sense to add a new configuration option which allows the user to define their own separator (the default being a colon).

avaris:
:+1: 
I think a configuration option is a good solution.

facelessuser:
I was looking to take care of this and I saw this in the footnote code:

```python
[...]

PR #728, opened 2018-10-09
foot note adjustments

facelessuser:
Add a config to set the footnote separator. Also remove rel/rev as they aren't really compatible with HTML5 and we already have classes set for refs and backrefs. Ref #723.


tests/test_extensions.py
```diff
@@ -1121,14 +1121,35 @@ def testBacklinkText(self):
...
             '<li id="fn:1">\n'
-            '<p>A Footnote&#160;<a class="footnote-backref" href="#fnref:1" rev="footnote"'
+            '<p>A Footnote&#160;<a class="footnote-backref" href="#fnref:1"'
             ' title="Jump back to footnote 1 in the text">back</a></p>\n'
             '</li>\n'
             '</ol>\n'
             '</div>'
         )
+
+    def testFootnoteSeparator(self):
```
waylan: Let's add this new test using the new testing framework. Ideally the existing tests should be moved as well. But I'll settle for the new one only for now.

facelessuser: I'll most likely move all footnote tests in this file to the new format.

facelessuser:
Requested changes have been made.

minchinweb:
Out of curiosity, is there an ETA for when a release containing this fix will be pushed out?

facelessuser:
I am hoping soon, not specifically for this issue, but for the code escape issue: https://github.com/Python-Markdown/markdown/issues/725.

waylan:
I expect a release once #732 is finalized (which includes some maintenance to bring the repo inline with the guidelines).

Note that this adds a feature, so this cannot be part of a bugfix release, but will be a point release (3.1). Therefore, it is a little slower moving.

justinmayer:
@waylan: Any chance that a v3.0.2 release could be published sooner that contains the aforementioned fixes? (i.e., without new features)

waylan:
@justinmayer that is a valid question. However, the effort to backport the bugfixes and the effort to finalize 3.1 are probably about the same. Therefore, I think it makers more sense to move forward with 3.1. It's just the matter of finding the time.

Issue #730, opened 2018-10-09
Move defined block list to init


facelessuser:
I believe we should move the block HTML list to the class init https://github.com/Python-Markdown/markdown/blob/master/markdown/core.py#L55.

I believe much like the escaped character list, if plugins take advantage of this, it will modify the base class instead of the class instance.

I'm opening this issue to allow disagreement with this before I do the work.

waylan:
> I believe much like the escaped character list, if plugins take advantage of this, it will modify the base class instead of the class instance.

You are correct. This could be a problem in multi-treadled environments, etc.

PR #731, opened 2018-10-09
Ensure block elements are defined per instance

facelessuser:
Block level elements should be defined per instance, not as base class variables. Ref #730

### DIFF
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -942,6 +942,18 @@ class TestEscapeAppend(unittest.TestCase):
         self.assertEqual('|' not in md2.ESCAPED_CHARS, True)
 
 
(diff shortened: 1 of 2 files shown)


tests/test_apis.py
```diff
@@ -942,6 +942,18 @@ def testAppend(self):
         self.assertEqual('|' not in md2.ESCAPED_CHARS, True)
 
 
+class TestBlockAppend(unittest.TestCase):
```
facelessuser, line 945: @waylan , wasn't sure if you have a special place for **new** API tests, so I put this with the old.

waylan: This isn't actually testing rendered output, so this is were it belongs.

PR #732, opened 2018-10-10
Add contributing guide.

waylan:
A first draft. It's still a little rough and incomplete.

For the record I found inspiration from

* [encode/django-rest-framework], 
* [atom/atom] and 
* [jessesquires/HowToContribute].

[encode/django-rest-framework]: https://github.com/encode/django-rest-framework/blob/master/CONTRIBUTING.md
[atom/atom]: https://github.com/atom/atom/blob/master/CONTRIBUTING.md
[jessesquires/HowToContribute]: https://github.com/jessesquires/HowToContribute

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -68,13 +68,13 @@ necessarily need to be the final, finished submission. In fact, if you discover
 an issue and intend to provide a fix for it, there is no need to open an issue
 first. You can report the issue and provide the fix together in a pull request.
 
-All pull requests should be made from your personal clone of the library hosted
+All pull requests should be made from your personal fork of the library hosted
 in your personal GitHub account. Do not create branches on the
 [Python-Markdown/markdown] project for pull requests. All pull requests should
 be implemented in a new branch with a unique name. Remember that if you have an
 outstanding pull request, pushing new commits to the related branch of your
-GitHub repository will also automatically update the pull request. If may help
-to review GitHub's documentation on [Using Pull Requests].
+GitHub repository will also automatically update the pull request. It may help
+to review GitHub's documentation on [Creating a pull request from a fork].
 
 If you are providing a fix for a previously reported issue, you must reference
 the issue in your commit message. Be sure to prefix the reference with one of
@@ -242,6 +242,102 @@ should be found. References to issues and pull requests should only provide the
 context in which a choice was made. However, the commit should be able to stand
 on its own.
 
+## Development Environment
+
+To start developing on Python-Markdown is it best to create a [fork] of the
+project on GitHub. After [cloning your fork] to your local system, you will want
+to [configure a remote] that points to the upstream repository so that you can
+[sync changes] made in the original repository with your fork.
+
+It is recommended that all development be done from within a Python [virtual
+environment], which isolates any experimental code from the general system. To
+create a virtual environment, use the following command from the root of the
+local working copy of your GitHub fork:
+
+```sh
+virtualenv venv
+```
+
+That creates a virtual environment which is contained in the `venv` directory
+within your local working copy. Note that the repository is configured so that
+git will ignore any files within a directory named `venv` or `ENV` for this
+very reason.
+
+On Posix systems (Linux, BSD, MacOS, etc.), use the following command to
+activate the environment:
+
+```sh
+source venv/bin/activate
+```
+
+On Windows, use this command instead:
+
+```sh
+venv/Scripts/activate
+```
+
+See the [User Guide] for more information on using virtual environments.
+
+To be able to run the Markdown library directly while working on it, install the
+working copy into the environment in [Development Mode] after activating the
+virtual environment for the first time:
+
+```sh
+python setup.py develop
+```
+
+Now any saved changes will immediately be available within the virtual
+environment.
+
+You can run the command line script with the following command:
+
+```sh
+python -m markdown
+```
+
+And you can directly run the tests with:
+
+```sh
+python -m unittest discover tests
+```
+
+!!! note
+
+    Some tests require the [PyTidyLib] library, which depends on the [HTML Tidy]
+    library. If you do not have PyTidyLib installed, the tests which depend upon
+    it will be skipped. Given the difficulty in installing the HTML Tidy library
+    on many systems, you may choose to leave both libraries uninstalled and
+    depend on the Travis server to run those tests when you submit a pull
+    request.
+
+The above setup will only run tests against the code in one version of Python.
+However,  Python-Markdown supports multiple versions of Python. Therefore, a
+[tox] configuration is included in the repository, which includes test
+environments for all supported Python versions, a [Flake8] test environment, and
+a spellchecker for the documentation. While it is generally fine to leave those
+tests for the Travis server to run when a pull request is submitted, for more
+advanced changes, you may want to run those tests locally. To do so, simply
+install tox:
+
+```sh
+pip install tox
+```
+
+Then, to run all configured test environments, simply call the command `tox`
+with no arguments. See help (`tox -h`) for more options.
+
+!!! note
+
+    The tox environments expect that some dependencies are already installed on
+    your system. For example, by default, any Python version specific
+    environment will fail if that version of Python is not installed.
+    Additionally, the tox environments assume that the [HTML Tidy] library is
+    installed and may fail when attempting to install [PyTidyLib] if it is not.
+    Finally, the `spellchecker` environment requires [aspell] and the
+    `aspell-en` dictionary to be installed. Unfortunately, installing those
+    dependancies may differ significantly from system to system and is outside
+    the scope of this guide.
+
 ## Versions
 
 Python-Markdown follows [Semantic Versioning] and uses the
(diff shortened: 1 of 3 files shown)


docs/contributing.md
```diff
@@ -0,0 +1,151 @@
...
+Some tips on good issue reporting:
+
+* When describing issues try to phrase your ticket in terms of the behavior you
+  think needs to change rather than the code you think needs to change.
+* Make sure you're running the latest version of Python-Markdown before
+  reporting an issue.
+* Search the issue list first for related items. Be sure to check closed issues
+  and pull requests. GitHub's search only checks open issues by default.
+* You may want to check the [syntax rules] and/or [Babelmark] to confirm that
+  you expectations align with the rules and/or other implementations of
```
mitya57: you expectations → you**r** expectations

docs/contributing.md
```diff
@@ -0,0 +1,151 @@
...
+outstanding pull request, pushing new commits to the related branch of your
+GitHub repo will also automatically update the pull request. If may help to
+review GitHub's documentation on [Using Pull Requests].
+
+If you are providing a fix for a previously reported issue, you must reference
+the issue in your commit message. Be sure to prefix the reference with one of
+GitHub's [action words] which will automatically close the issue when the pull
+request is merged. For example, `fixes #42` and `closes #42` would be
+acceptable, whereas `ref #42` would not. Of course, if merging a pull request
+should not cause an an issue to be closed, then the action word should not be
```
mitya57: an an → an

docs/contributing.md
```diff
@@ -0,0 +1,151 @@
...
+`master` branch when the MAJOR and/or MINOR version is in `dev` status.
+
+A separate commit to the `master` branch should be made to bump up the MAJOR
+and/or MINOR version and set the `dev` status. Only then will any PRs
+implementing new features or backward incompatible changes be accepted.
+
+If a bug fix is deemed to be important and the `master` branch is in `dev`
+status, a backport of the fix should be committed to a version branch. If the
+appropriate version branch does not exist, then it should be created and a PR
+back-porting the fix made against that branch.The version branch should be named
```
mitya57: Missing space between two sentences.

docs/contributing.md
```diff
@@ -0,0 +1,151 @@
...
+
+
+[Python-Markdown Organization]: https://github.com/Python-Markdown
+[Python-Markdown Code of Conduct]: https://github.com/Python-Markdown/markdown/blob/master/CODE_OF_CONDUCT.md
+[Python-Markdown/markdown]: https://github.com/Python-Markdown/markdown
+[issue tracker]: https://github.com/Python-Markdown/markdown/issues
+[syntax rules]: http://daringfireball.net/projects/markdown/syntax
+[Babelmark]: http://johnmacfarlane.net/babelmark2/
+[Using Pull Requests]: https://help.github.com/articles/using-pull-requests
+[action words]:
```
mitya57: This should point to https://help.github.com/articles/closing-issues-using-keywords/.

mitya57 (dismissed):
I suggest some minor corrections.

docs/contributing.md
waylan: Thanks. I hadn't gotten around to digging that URL up yet and ran out of time yesterday.

docs/contributing.md
```diff
@@ -278,6 +278,42 @@ be named with the most recently released MINOR version. For example, if the
...
+   the date of the new release.
+
+3. Update the version defined in [`markdown/__init__.py`][markdown/__init__.py]
+   as well as in the `theme.release` option of the [`mkdocs.yml`][mkdocs.yml]
+   configuration file.
+
+4. Build a local copy of the documentation, browse through the pages and
+   confirm that no obvious issues exist with the documentation.
+
+4. Create a pull request with a commit message in the following format:
```
mitya57: There are two items `4.`. I know this one will render as `5.`, but it would be nice to keep the numbering consistent.

docs/contributing.md
```diff
@@ -99,12 +102,141 @@ are running as expected.
 
 ## Style Guides
 
+In an effort to maintain consistency, Python-Markdown follows the following
```
mitya57: I am not a native speaker, but “follows the following” does not sound nice to me. Maybe “adheres to the following”?

docs/contributing.md
```diff
@@ -0,0 +1,340 @@
...
+
+Use the present tense ("Add feature" not "Added feature").
+
+Use the imperative mood ("Move item to..." not "Moves item to...").
+
+Limit the first line to 72 characters or less.
+
+Reference issues and pull requests liberally after the first line. Include a
+summary of the changes/additions made without replicating the content of the
+documentation or release notes. This is were an explanation of the choices made
```
mitya57: were → where

mitya57 (dismissed):

docs/contributing.md
waylan: Good catch. I didn't notice that after multiple readings.

docs/contributing.md
```diff
@@ -0,0 +1,347 @@
...
+context in which a choice was made. However, the commit should be able to stand
+on its own.
+
+## Versions
+
+Python-Markdown follows [Semantic Versioning] and uses the
+`MAJOR.MINOR.POINT(PRERELEASE/DEV)` format for specifying releases. The status
+of the `master` branch should always be identified in the `__version_info__`
+variable defined in [`markdown/__init__.py`][markdown/__init__.py] and should
+conform to [PEP 440].
```
waylan: Note that we currently conform to PEP 386, which had been superseded by 440. We will need to make a minor adjustment before we conform to this guide.

waylan: Just a note that [pypa/packaging] officially implements PEP 404. And as of pypa/setuptools@9382fa0c, (version 8.0) `setuptools` and `pkg_resources` vendor and use `pypa/packaging` to do all version parsing. Finally, since 52b8da8e (version 3.0) we are using `setuptools` for installation.

Therefore, the differences between `pkg_resources` and PEP 440 listed in the PEP are presumably outdated and we should be able to use `pkg_resources` for all version handling without any added dependencies. In other words, we can remove our own version parsing code when conforming to PEP 440.

[pypa/packaging]: https://github.com/pypa/packaging/

docs/contributing.md
```diff
@@ -0,0 +1,451 @@
...
+* Branch `master` should generally be stable and release-ready at all times.
+* Version branches should be used for bug-fixes back-ported to the most recent
+  PATCH release.
+* No other branches should be created. Any other branches which exist are
+  preserved for historical reasons only.
+
+## Issues
+
+Please address issues on the correct channel. Usage questions should be directed
+to the discussion group. Feature requests, bug reports and other issues should
```
waylan: My understanding is that the old sourceforge mailing list (linked in the readme) isn't working any more. The archives still exist, but no new messages are being accepted.

Unless we create some new discussion list somewhere, this comment should be removed. If we do create something, then we should link to it here. Anyone have any suggestions?

waylan: Some obvious solutions would be Google Groups or Slack. Although, I believe Slack doesn't offer public archives. 

[Groups.io] looks interesting (even supports Markdown and has GitHub integration), but I don't know much about it.

[Groups.io]: https://groups.io/

facelessuser: I don't know much about Groups.io, but I've seen people use [Discord](https://discordapp.com/) for this kind of thing. It's marketed for gamers, but I've seen it used for all sorts of things.

Gitter is popluar, but I honestly don't like their mobile chat app, it's just a web view of their web interface.

waylan: Slack, Gitter, Discord, etc are all "chat" solutions, not mailing list solutions. AFAICT, a common feature is that the archives are private to members only. We want something with public archives, which generally means a mailing list. Groups.io appears to be a modern take on mailing lists, but the feature I think we need is the ability to participate outside of an email client. I'm not sure if groups.io supports that. Google Groups certainly does. And its not unusual for people to already have a google account which eliminates the need for people to create another account just to get help.

It is possible to move an old mailman list archive to google groups, but I don't have admin access to the group as it was created before I got involved. I reached out to @yuri when we got notice that SourceForge was shutting the list down, but never heard back.

Another option would be to just use GitHub issues for support questions in addition to bug reports and feature requests. We would need to redefine the `labels` to properly categorize things, but given the amount of questions we get, that is probably sufficient.

facelessuser: I would agree that GitHub Issues is probably sufficient.  If we were getting a ton of activity, then I think it would make more sense to have a separate area to chat.

waylan:
I just pushed a change which now allows usage questions to be asked as GitHub Issues. I also included a list of [labels] and a description of each. Note that many new labels have been introduced. Please provide any feedback you may have before I update GitHub with the new labels.

facelessuser:
I have no complaints.

waylan:
I have updated the labels here in GitHub. I used a color scheme for the status labels which is reminiscent of traffic lights:

green: various approved states (expect for `someday-maybe` which is gray).
yellow: waiting on something
red: various rejected states

All other labels (types and categories) arbitrarily use blue, orange, purple, brown... not shades of green/yellow/red.

mitya57 (approved):
The latest changes look good to me. And I really like the proposed issue/PR labels! :+1:

facelessuser:
Yeah, I may steal the label convention or at least adapt it for some other projects.  I've always had problems coming up with a label system I was happy with.

waylan:
I think this is done. I did a final review and only found a couple minor tweaks.

I also color coded the labels in the docs. Any tweaks to the CSS are welcome:

screenshot_2018-10-24 contributing to python-markdown python-markdown 3 0 1 documentation

Issue #42, opened 2011-09-12
Placeholder slips to HTML markup


earshinov:
When in Markdown source an HTML comment contains Markdown's link markup, the resulting HTML contains link placeholder in place of the comment's text:

Example.md:

<pre><code>
Please see &lt;!-- [Example][1] --&gt;

<span>[</span>1]: http://example.com/
</code></pre>

Example.html (rendered with python-markdown 2.0.3):

<pre><code>
&lt;p&gt;Please see &lt;!-- klzzwxh:0000 --&gt;&lt;/p&gt;
</code></pre>

Actually there are STX and EXT chars in the output, they are just not shown.

waylan:
This has already been fixed in the current code and will be available in the next release (currently in alpha - see [downloads](https://github.com/waylan/Python-Markdown/downloads)).

``` python
>>> markdown.version
'2.1.0.alpha'
>>> markdown.markdown("Please see <!-- [Example][1] -->")
u'<p>Please see <!-- [Example][1] --></p>'
```

earshinov:
OK, sorry that I didn't test the latest version.

PR #740, opened 2018-10-25
Deprecate version and version_info

### LINKED ISSUE #739: Possible old version deprecation path
This is just a proposal.  I think it is a neat idea and could lead us to a path to deprecate old versions. Is this something worth doing? I'll leave that up to the team to decide, I just thought this was interesting.

So let me start by saying, this idea is actually from something that I read Guido said that they purposely allowed in the import system: https://mail.python.org/pipermail/python-ideas/2012-May/014969.html.  More elaborate implementations have been implemented in libraries such as Twisted, but for our purposes, a simple implementation based off Guido's remark would be all we would need.

The reason why deprecating attributes is difficult is because Python doesn't support `__getattributes__` on modules (before Python 3.7). So to get this functionality you'd have to wrap the module in a class.  Now I know what your thinking, that sounds like hackery, but see Guido's link above. So here's the idea, in `util` add the following class (I've also modified the decorator wrap to allow specifying the stacklevel as we have an additional layer of abstraction).

Notice below that we wrap the class, and first look in the class for attributes we desire, if we fail, we just get it from the module.

util.py

```py
def deprecated(message, stacklevel=2):  # pragma: no cover
    """
    Raise a `DeprecationWarning` when wrapped function/method is called.

    Borrowed from https://stackoverflow.com/a/48632082/866026
    """

    def deprecated_decorator(func):
        """Deprecation decorator."""

        @wraps(func)
        def deprecated_func(*args, **kwargs):
            """Display deprecation warning."""

            warnings.warn(
                "'{}' is deprecated. {}".format(func.__name__, message),
                category=DeprecationWarning,
                stacklevel=stacklevel
            )
            return func(*args, **kwargs)
        return deprecated_func
    return deprecated_decorator


class ModuleWrap(object):
    """
    Provided so that we can deprecate old version methodology.

    https://mail.python.org/pipermail/python-ideas/2012-May/014969.html. 
    """

    def __init__(self, module):
        """Initialize."""

        self._module = sys.modules[module]
        sys.modules[module] = self

    def __getattribute__(self, name):
        """Get the class attribute first, and fallback to the module."""

        try:
            return super(ModuleWrap, self).__getattribute__(name)
        except AttributeError:
            return getattr(super(ModuleWrap, self).__getattribute__('_module'), name)
```

Next we add this in `markdon/__init__.py`

```py3
from .util import ModuleWrap, deprecated

# Also support `version` for backward-compatabillity with <3.0 versions
class _ModuleWrap(ModuleWrap):
    """
    Wrap module with special deprecator to deprecate global objects.

    https://mail.python.org/pipermail/python-ideas/2012-May/014969.html
    """

    @property
    @deprecated("Use '__version__' instead.", stacklevel=3)
    def version(self):
        """Get deprecated version."""

        return super(_ModuleWrap, self).__getattribute__('_module').__version__

    @property
    @deprecated("Use '__version_info__' instead.", stacklevel=3)
    def version_info(self):
        """Get deprecated version info."""

        return super(_ModuleWrap, self).__getattribute__('_module').__version_info__


_ModuleWrap(__name__)
```

So does it work?

```py3
>>> import markdown
>>> import warnings
>>> warnings.simplefilter('always')
>>> markdown.version
__main__:1: DeprecationWarning: 'version' is deprecated. Use '__version__' instead.
'3.0.1'
>>> markdown.__version__
'3.0.1'
```

No problem with other imports either:

```py3
>>> from markdown import extensions
>>> from markdown import *
>>> Markdown().convert('# text')
'<h1>text</h1>'
```

I know that originally the idea was to have the more standard `__version__` etc., but we had to add the old stuff in for compatibility, but this is probably the best idea I can think of to allow a path forward and kill the old.

Had we been including `version` in things in `__all__ = ['Markdown', 'markdown', 'markdownFromFile']`, I'm not sure if it would have worked as well, but since we don't, this is a possible path.

Let me know if this intrigues or horrifies you 🙂 .

facelessuser:
This issue fixes #739. This essentially implements the closest we can get to PEP562 which allows for modules to control `__dir__` and `__getattr__` in order to deprecate attributes. Here we provide a wrapper class for the module in `util`. If a module has attributes that need to deprecated, we derive from the wrapper class and define the attributes as functions with the `property` decorator and the provided `deprecated` decorator. The class is instantiated with the module's `__name__` attribute and the class will properly replace the module with the wrapped module.  When accessing the depracted attributes, a warning will raised.

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -196,7 +196,7 @@ class ModuleWrap(object):
         # Include the module's attributes and the the attributes that have been added.
         attr = (
             set(dir(super(ModuleWrap, self).__getattribute__('_module'))) |
-            (set(super(ModuleWrap, self).__dir__()) - set(dir(ModuleWrap)))
+            (set(self.__class__.__dict__.keys()) - set(ModuleWrap.__dict__.keys()))
         )
 
         return sorted(list(attr))
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -178,7 +178,12 @@ MISC AUXILIARY CLASSES
 
 
 class ModuleWrap(object):
-    """Provided so that we can deprecate old version methodology."""
+    """
+    Provided so that we can deprecate old version methodology.
+
+    See comments from Guido: <https://mail.python.org/pipermail/python-ideas/2012-May/014969.html>
+    and see PEP 562 which this is essentially a backport of: <https://www.python.org/dev/peps/pep-0562/>.
+    """
 
     def __init__(self, module):
         """Initialize."""
@@ -190,10 +195,10 @@ class ModuleWrap(object):
         """
         Implement the `dir` command.
 
-        Return module's `dir` and any attributes that are not private.
+        Return module's results for the `dir` command along with any
+        attributes that have been added to the class.
         """
 
-        # Include the module's attributes and the the attributes that have been added.
         attr = (
             set(dir(super(ModuleWrap, self).__getattribute__('_module'))) |
             (set(self.__class__.__dict__.keys()) - set(ModuleWrap.__dict__.keys()))
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -202,12 +202,12 @@ class ModuleWrap(object):
         return sorted(list(attr))
 
     def __getattribute__(self, name):
-        """Get the module attribute first and fallback to the module if not available."""
+        """Get the class attribute first and fallback to the module if not available."""
 
         try:
-            return getattr(super(ModuleWrap, self).__getattribute__('_module'), name)
-        except AttributeError:
             return super(ModuleWrap, self).__getattribute__(name)
+        except AttributeError:
+            return getattr(super(ModuleWrap, self).__getattribute__('_module'), name)
 
 
 class AtomicString(text_type):
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -113,7 +113,7 @@ AUXILIARY GLOBAL FUNCTIONS
 """
 
 
-def deprecated(message):
+def deprecated(message, stacklevel=2):
     """
     Raise a DeprecationWarning when wrapped function/method is called.
 
@@ -125,7 +125,7 @@ def deprecated(message):
             warnings.warn(
                 "'{}' is deprecated. {}".format(func.__name__, message),
                 category=DeprecationWarning,
-                stacklevel=2
+                stacklevel=stacklevel
             )
             return func(*args, **kwargs)
         return deprecated_func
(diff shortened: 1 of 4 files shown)


facelessuser:
Turns out this works fine with `__all__` even if those items are deprecated in `__all__`. We don't do that, but it is nice to know you get the warning on `from module import *`. 

Let me know if we see improvements that can be made.

markdown/util.py
```diff
@@ -177,6 +177,39 @@ def code_escape(text):
...
+        """Initialize."""
+
+        self._module = sys.modules[module]
+        sys.modules[module] = self
+
+    def __dir__(self):
+        """
+        Implement the `dir` command.
+
+        Return module's `dir` and any attributes that are not private.
```
waylan: Not sure we should include `dir`. If we don't, then the deprecated vars won't get listed in the result. But, given they are deprecated, why would we want to make them discoverable?

On the other hand, including `dir` here provides feature parity with Python 3.7 behavior. Guess I'm indifferent about it. I wouldn't have added it, but now that it's here, I'm okay leaving it.

facelessuser: I think not having prevents it from working properly when the items are listed in "all", but we don't have ours there, so it doesn't matter. I can exclude then from dir if you prefer, but I'm also indifferent.

facelessuser:
Guess I need a test to cover dir as well.

waylan:
And now that it is documented, to address a pet peeve of mine... 😉 ... the [Contributing Guide][1] states:

> If you are providing a fix for a previously reported issue, you must reference the issue in your commit message. Be sure to prefix the reference with one of GitHub's [action words] which will automatically close the issue when the pull request is merged. For example, `fixes #42` and `closes #42` would be acceptable, whereas `ref #42` would not. Of course, if merging a pull request should not cause an issue to be closed, then the action word should not be included when referencing that issue.

[1]: https://github.com/Python-Markdown/markdown/pull/732/files#diff-05b7a8e562e0a482b29460c121776963R78
[action words]: https://help.github.com/articles/closing-issues-using-keywords/

facelessuser:
Guess I should read the guide again :smile:.

facelessuser:
All tests are passing.  I got hung up on Python 2 and Python 3 differences when implementing `__dir__`, but the results are now consistent across all Python versions.  I can't wait till EOL of Python 2.  I also updated the issues message properly refrencing the issue and providing a detailed description.

waylan:
Looks good. Just need to update the release notes and this will be good to go.

PR #743, opened 2018-10-27
Fix for bad id whitespace collapsing in references

### LINKED ISSUE #742: Line break in link references inside lists
Hey!

Consider the following code:

```
Two things:

 - I would like to tell you about the [code of
   conduct][] we are using in this project.
 - Only one in fact.

[code of conduct]: https://github.com/Python-Markdown/markdown/blob/master/CODE_OF_CONDUCT.md
```

CommonMark is rendering the link correctly. GitHub as well (see below). PHP Markdown does render correctly too. However, Python-Markdown doesn't because of the space at the beginning of the line. I don't know if space at the beginning of a line should just be ignored inside a link or if only the exact amount of space used for list indent should be ignored.

---
Two things:

 - I would like to tell you about the [code of
   conduct][] we are using in this project.
 - Only one in fact.

[code of conduct]: https://github.com/Python-Markdown/markdown/blob/master/CODE_OF_CONDUCT.md

facelessuser:
This fixes #742 by generally collapsing all whitespace down to a single space in reference ids opposed to the previous logic of collapsing only newlines preceeded by whitespace.

### DIFF
--- a/docs/change_log/release-3.1.md
+++ b/docs/change_log/release-3.1.md
@@ -35,3 +35,4 @@ The following bug fixes are included in the 3.1 release:
 * Block level elements are defined per instance, not as class attributes
   (#731).
 * Double escaping of block code has been eliminated (#725).
+* Problems with newlines in references has been fixed (#742).
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -617,7 +617,7 @@ class ImageInlineProcessor(LinkInlineProcessor):
 
 class ReferenceInlineProcessor(LinkInlineProcessor):
     """ Match to a stored reference and return link element. """
-    NEWLINE_CLEANUP_RE = re.compile(r'[ ]?\n', re.MULTILINE)
+    NEWLINE_CLEANUP_RE = re.compile(r'\s+', re.MULTILINE)
 
     RE_LINK = re.compile(r'\s?\[([^\]]*)\]', re.DOTALL | re.UNICODE)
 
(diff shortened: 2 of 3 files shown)


vincentbernat:
Works fine for me and it doesn't change anything on my existing set of markdown documents where I avoided to run into this case. Thanks!

waylan:
I'm thinking we need another test. I doubt this would be a problem, but lets make sure this doesn't break things were we shouldn't get a link. For example, a blank line between the two lines:

```md
I would like to tell you about the [code of

conduct][] we are using in this project.
```

That specific example shouldn't be a problem as the two halves are in different paragraphs (and therefore not in the same block of text), but it illustrates my concern: If the new regex was run against that block of text, it would strip out the blank line in addition to any other extra whitespace. And that would be an error.

I'm just not coming up with an example that might actually occur. I want to make sure we have given it some thought and, at the very least, confirm that the above example [works as expected][1].

[1]: http://johnmacfarlane.net/babelmark2/?normalize=1&text=I+would+like+to+tell+you+about+the+%5Bcode+of%0A%0Aconduct%5D%5B%5D+we+are+using+in+this+project.

facelessuser:
I imagine the stated example will work as expected. I'm happy to put it in.

I'm also having difficulty coming up with something desired that this could break. Currently, no existing tests fail, so that is a plus. This may be something, that in time, someone will have a contrived example we must account for, but currently I cannot come up with one.

waylan:
I agree that any read-world failure would likely be from a rather contrived edge case that we are not likely to think up ahead of time. That said, I think we need to add some test to show that we at least considered it.

facelessuser:
> That said, I think we need to add some test to show that we at least considered it.

I agree. Hopefully I'll get to it later today.

facelessuser:
Requested check is in and is passing as expected.

waylan:
I just realized this needs a comment in the release notes. A one-liner under bug fixes should suffice.

facelessuser:
Added.

PR #744, opened 2018-10-28
Add a method splitChunks

frodo821:
Fixed a bug that Markdown blocks aren't properly divided.

I think the document
````Markdown
# Test Document

```Python
def main():
    print("Hello")

if __name__ == "__main__":
    main()
```

<div>
This is div block.

FooBar
</div>
````
is splitted into ````['# Test Document', '```Python\ndef main():\n    print("Hello")\n\nif __name__ == "__main__":\n    main()\n```'...]````, but actual is ````['# Test Document', '```Python\ndef main():\n    print("Hello")', 'if __name__ == "__main__":\n    main()\n```'...]````.

I think the current behaviour (the latter) is a bug. What don't you?
Thanks!

### DIFF
--- a/markdown/blockparser.py
+++ b/markdown/blockparser.py
@@ -22,6 +22,10 @@ License: BSD (see LICENSE.md for details).
 
 from __future__ import unicode_literals
 from __future__ import absolute_import
+
+from collections import deque
+from re import search, MULTILINE, DOTALL
+
 from . import util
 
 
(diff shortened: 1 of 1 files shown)


facelessuser:
If you want fenced blocks to be recognized, then you have to use the fenced_code extension, or some other 3rd party extension. Fenced code is not part of the Markdown spec, which is why it is not handled by default.

frodo821:
Ok, I've got it.

waylan:
Thank you for your contribution. However, there is absolutely no way that that "HTML" parser is sufficient to accurately find all valid posable opening and closing HTML tags. As anyone who has tried quickly comes to understand, HTML parsing is "complicated".

And if we were going to implement fenced code and HTML parsing into the core, we would not be doing it that way. Each should be their own blockprocessor. And yes, the solution would be complicated. That's why both are currently implemented as preprocessors. Of course, the raw HTML processor is on by default, but as @facelessuser explained, fenced code blocks are non-standard and off by default.

In any event, this solution is not something we can use and therefore is being closed.

Issue #751, opened 2018-11-06
Prettier support


lydell:
I know about #451 and about the following:

> https://daringfireball.net/projects/markdown/syntax#list
>
> List items may consist of multiple paragraphs. Each subsequent paragraph in a list item must be indented by either 4 spaces or one tab

But not even the original markdown implementation follows that. Try pasting the following:

```md
- one

  two
```

…in https://daringfireball.net/projects/markdown/dingus. Result:

```html
<ul>
<li><p>one</p>

<p>two</p></li>
</ul>
```

And since then, [CommonMark](https://commonmark.org/) has gained popularity, and it allows fewer spaces:

> https://spec.commonmark.org/0.28/#example-219
>
> ```md
> - one
>
>   two
> ```
>
> ```html
> <ul>
> <li>
> <p>one</p>
> <p>two</p>
> </li>
> </ul>
> ```

And here’s the thing: [Prettier](https://prettier.io/) can auto-format markdown, and it uses CommonMark spec. Which unfortunately means that it turns more 4 spaces into the more good-looking 2 spaces:

**Prettier 1.14.3**
[Playground link](https://prettier.io/playground/#N4Igxg9gdgLgprEAuEBaABNOAdKv0HowDuEIANCBAA4wCW0AzsqAIYBO7ExAChwsxSsAbhDoATCiABG7VmADWcGAGVq8ulADmyGOwCucSpsZx2MHnK0BbVsgBmrADanKAK0YAPAEJzFylVZrOAAZTTgHZ1cQdXZTdmQQW3YFcW4oKWp2TRgAdQkYAAtkAA4ABkosiFNcuWpErLh44QjKdjgAR306dstWGzskRxcjEFNrOl0DUcZNLSc4AEV9CHhIkcoYVml88SLkACZNuTonOYBhCGtbRKgsKX1TABVtwWHTAF8PoA)
```sh
--parser markdown
```

**Input:**
```markdown
- one

    two
```

**Output:**
```markdown
- one

  two

```

Which means that I cannot use Prettier to format markdown that will be parsed using this package :(

Would you be willing to reconsider this?

- Not even the original implementation follows the 4 space rule.
- Most markdown implementations don’t either.
- Being able to use Prettier would be awesome!

EDIT: This fixes the issue: https://github.com/radude/mdx_truly_sane_lists

waylan:
> Not even the original implementation follows the 4 space rule.

I consider this a bug and will never be convinced otherwise.

[...]

Issue #442, opened 2015-11-10
List handling not consistent with Markdown.pl


kitt-cat:
Given this input markdown

```
Some text

* **Example**

 This should be part of the list element

* **Other example**

 But it's not in to python-markdown :/

* **list element**

 Instead it's a separate list!

End text
```

I expect it to render like this:

> Some text
> - **Example**
>   
>   This should be part of the list element
> - **Other example**
>   
>   But it's not in to python-markdown :/
> - **list element**
>   
>   Instead it's a separate list!
> 
> End text

But python-markdown gives this to me:

```
<p>Some text</p>
<ul>
<li><strong>Example</strong></li>
</ul>
<p>This should be part of the list element</p>
<ul>
<li><strong>Other example</strong></li>
</ul>
<p>But it's not in to python-markdown :/</p>
<ul>
<li><strong>list element</strong></li>
</ul>
<p>Instead it's a separate list!</p>
<p>End text</p>
```

Every list element is put in its own list!

---

`Markdown.pl` from https://daringfireball.net/projects/markdown/ behaves as expected:

```
<p>Some text</p>

<ul>
<li><p><strong>Example</strong></p>

<p>This should be part of the list element</p></li>
<li><p><strong>Other example</strong></p>

<p>But it's not in to python-markdown :/</p></li>
<li><p><strong>list element</strong></p>

<p>Instead it's a separate list!</p></li>
</ul>

<p>End text</p>
```

mitya57:
You need to use four spaces for indentation. The “Indentation/Tab Length” section in the [documentation](http://pythonhosted.org/Markdown/#differences) explains why it is so and how to change it.

Let me quote it here:

> The [syntax rules](http://daringfireball.net/projects/markdown/syntax#list) clearly state that when a list item consists of multiple paragraphs, “each subsequent paragraph in a list item **must** be indented by either 4 spaces or one tab” (emphasis added). However, many implementations do not enforce this rule and allow less than 4 spaces of indentation. The implementers of Python-Markdown consider it a bug to not enforce this rule.
> 
> This applies to any block level elements nested in a list, including paragraphs, sub-lists, blockquotes, code blocks, etc. They **must** always be indented by at least four spaces (or one tab) for each level of nesting.
> 
> In the event that one would prefer different behavior, [tab_length](http://pythonhosted.org/Markdown/reference.html#tab_length) can be set to whatever length is desired. Be warned however, as this will affect indentation for all aspects of the syntax (including root level code blocks).

Issue #313, opened 2014-05-29
early list termination with multi-paragrah list items containing a blank line


minektur:
```
#!/usr/local/bin/python2.7
from markdown import markdown
testinput = """* a b
  c"""
output = markdown(testinput)
print output
```

Yields

```
<ul>
<li>a b</li>
</ul>
<p>c</p>
```

When it should yield

```
<ul>
<li><p>foo</p>

<p>bar</p></li>
</ul>
```

at least according to http://daringfireball.net/projects/markdown/dingus

data may be available at: http://pastebin.com/fiCx80P4

minektur:
hm... seems if I make 4 spaces instead of 2 on the last line after the break, It works.  Is that the spec?  where is the spec? :)

waylan:
I would suggest reading the [Differences](https://pythonhosted.org/Markdown/index.html#differences) section of the docs, specifically the second list item: "Indentation/Tab Length". As stated there, this is not a bug, but the intended behaviour. The bug is in other implementations.

Issue #125, opened 2012-07-24
Wrong sublist parsing


Scorpil:
Python-Markdown is not supporting subsists in the way that Mardown.pl does.

For example, if we have this markdown input:

```
 * List
   * Sublist
      * Subsublist
 * List again
 * List once more
```

We have different outputs. Python version:

```
$ markdown_py input 
<ul>
<li>List</li>
<li>Sublist<ul>
<li>Subsublist</li>
</ul>
</li>
<li>List again</li>
<li>List once more</li>
</ul>
```

And Markdown.pl creates this code:

```
$ perl Markdown.pl input 
<ul>
<li>List
<ul>
<li>Sublist
<ul>
<li>Subsublist</li>
</ul></li>
</ul></li>
<li>List again</li>
<li>List once more</li>
</ul>
```

For me this is kind of important issue - i use sublists a lot. I'll try to write some code to change this behavior once i have more spare time.

waylan:
This is a known difference and a request that has been rejected many times in the past. Python-Markdown consistently requires 4 spaces of indent per level or it ignores the indent (markdown.pl does this in some cases not in others) If you want a different amount of indent, then set the the desired amount with the [tab_length](http://packages.python.org/Markdown/reference.html#tab_length) keyword. Of course, this will consistently change the indent for the entire document.

Alternatively, you can always use the [Extension API](http://packages.python.org/Markdown/extensions/api.html) to write your own extension which implements whatever behavior you want.

Personally, I have no interest in implementing what I consider to be a bug and incomprehensible inconsistency in markdown.pl.

PR #752, opened 2018-11-06
Update docs about list indentation

lydell:


docs/index.md
```diff
@@ -91,7 +92,9 @@ are summarized below:
     In the event that one would prefer different behavior,
     [tab_length](reference.md#tab_length) can be set to whatever length is
     desired. Be warned however, as this will affect indentation for all aspects
-    of the syntax (including root level code blocks).
+    of the syntax (including root level code blocks). Alternatively, you could
+    use the [mdx_truly_sane_lists](https://github.com/radude/mdx_truly_sane_lists)
```
waylan: While I understand the spirit of this change, I'd rather not favor any one third-party extension over another. Instead, this link should point to the [wiki page][1] where we keep a list of all known third party extensions. Perhaps one of them will meet the users needs.

And that list is publicly editable, so if you see an extension is missing, feel free to add it yourself.

[1]: https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions

docs/index.md
```diff
@@ -80,7 +80,8 @@ are summarized below:
     The [syntax rules](http://daringfireball.net/projects/markdown/syntax#list)
     clearly state that when a list item consists of multiple paragraphs, "each
     subsequent paragraph in a list item **must** be indented by either 4 spaces
-    or one tab" (emphasis added). However, many implementations do not enforce
+    or one tab" (emphasis added). However, many implementations – including the
+    original markdown.pl implementation itself – do not enforce
```
waylan: I see no value in adding this. Lets leave it out.

waylan (changes requested):
Thanks for your contribution. I have noted some suggestions below. If you make the requested changes, I'll merge your PR.

docs/index.md
lydell: #125, #313, #442 and #751 mention the original implementation as a reason for Python-Markdown to also implement this, so I figured it'd be helpful to mention that we already know that markdown.pl already behaves the way it does to avoid duplicate issues.

waylan: People should be searching old issues before opening a new issue, in which case they would know that. I realize GitHub doesn't search closed issues by default, but we call this out in our (new) [contributing guide][1]. If people follow the guide, then this is a non-issue.

[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/contributing.md#issues

lydell: People don't. I know, because I didn't (until afterwards).

waylan (approved):

lydell:
It was worth a try! Thanks for your time! :tada:

PR #753, opened 2018-11-13
add: use stacklevel=2 with warning

blueyed:
Ref: https://github.com/encode/django-rest-framework/issues/6317

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -423,7 +423,9 @@ class Registry(object):
         if isinstance(key, string_type):
             warnings.warn(
                 'Using setitem to register a processor or pattern is deprecated. '
-                'Use the `register` method instead.', DeprecationWarning
+                'Use the `register` method instead.',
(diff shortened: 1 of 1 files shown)


mitya57:
This may be useful in other places where we issue warnings, too.

blueyed:
Some places where using it already, but looking again it is missing here: https://github.com/blueyed/markdown/blob/1052b333472a9702badf26e08841e030ed0d4b18/markdown/util.py#L421-L451

Not sure about dunder-methods and warnings though, so would need more investigation.

waylan:
> looking again it is missing here: https://github.com/blueyed/markdown/blob/1052b333472a9702badf26e08841e030ed0d4b18/markdown/util.py#L421-L451

As those are all related to the same feature, they should all behave the same. Could you make the same change there as well?

blueyed:
Updated.
Please test/verfify.

PR #754, opened 2018-11-14
Improve CI / testing

blueyed:
This centralized testing requirements and allows for
`pip install -e .[testing]` etc.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+dist: xenial
 sudo: false
 language: python
 
@@ -7,18 +8,19 @@ matrix:
       env: TOXENV=py27
     - python: '3.4'
       env: TOXENV=py34
+      dist: trusty
     - python: '3.5'
       env: TOXENV=py35
     - python: '3.6'
       env: TOXENV=py36
     - python: '3.7'
       env: TOXENV=py37
-      dist: xenial
-      sudo: true
     - python: 'pypy'
       env: TOXENV=pypy
+      dist: trusty
     - python: 'pypy3'
       env: TOXENV=pypy3
+      dist: trusty
     - env: TOXENV=flake8
     - env: TOXENV=checkspelling
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,15 +30,14 @@ addons:
     - aspell-en
 
 install:
-  - pip install 'setuptools>=36'
-  - pip install tox
-  - pip install 'coverage<4.0'
-  - pip install codecov
+  # NOTE: setuptools needs to be installed explicitly for py34 (trusty).
+  - pip install 'setuptools>=36' tox
 
 script:
   - tox
 
 after_success:
+  - pip install codecov
   - codecov
 
 branches:
--- a/.travis.yml
+++ b/.travis.yml
@@ -23,13 +23,16 @@ matrix:
       dist: trusty
     - env: TOXENV=flake8
     - env: TOXENV=checkspelling
+      addons:
+        apt:
+          packages:
+          - aspell
+          - aspell-en
 
 addons:
   apt:
     packages:
     - libtidy-0.99-0
-    - aspell
-    - aspell-en
 
 install:
   # NOTE: setuptools needs to be installed explicitly for py34 (trusty).
--- a/.travis.yml
+++ b/.travis.yml
@@ -40,7 +40,7 @@ script:
 
 after_success:
   - pip install codecov
-  - codecov
+  - codecov --flags "$TOXENV"
 
 branches:
   only:
--- a/setup.py
+++ b/setup.py
@@ -78,6 +78,12 @@ setup(
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
     install_requires=['setuptools >= 36'],
+    extras_require={
+        'testing': [
+            'coverage<4.0',
+            'pyyaml',
+        ],
+    },
     entry_points={
         'console_scripts': [
             '%s = markdown.__main__:run' % SCRIPT_NAME,
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
     install_requires=['setuptools >= 36'],
     extras_require={
         'testing': [
-            'coverage<4.0',
+            'coverage<5.0',
             'pyyaml',
         ],
     },
--- a/setup.py
+++ b/setup.py
@@ -80,7 +80,7 @@ setup(
     install_requires=['setuptools >= 36'],
     extras_require={
         'testing': [
-            'coverage<5.0',
+            'coverage',
             'pyyaml',
         ],
     },
--- a/test-requirements.txt
+++ b/test-requirements.txt
@@ -1,5 +1,0 @@
-coverage<4.0
-pyyaml
-pytidylib
-mkdocs>=1.0
-mkdocs-nature
--- a/tox.ini
+++ b/tox.ini
@@ -4,10 +4,8 @@ requires = setuptools>=36
 isolated_build = True
 
 [testenv]
-deps =
-    coverage<4.0
-    pyyaml
-    pytidylib
+extras = testing
+deps = pytidylib
 commands = coverage run --source=markdown -m unittest discover {toxinidir}/tests
            coverage report --show-missing
 


.travis.yml
```diff
@@ -21,18 +21,15 @@ matrix:
       env: TOXENV=pypy3
     - env: TOXENV=flake8
     - env: TOXENV=checkspelling
```
blueyed, line 25: What do you think about running flake8 and checkspelling in a single build job?
(`TOXENV=flake8,checkspelling`)

blueyed: Done, but I can take it out again, of course.

.travis.yml
```diff
@@ -7,39 +8,41 @@ matrix:
...
-    packages:
-    - libtidy-0.99-0
-    - aspell
-    - aspell-en
+      dist: trusty
+    - env: TOXENV=flake8,checkspelling
+      addons:
+        apt:
+          packages:
+          - libtidy-0.99-0
```
waylan: `libtidy` needs to be available for the syntax tests, not the spellchecking tests. In fact, we must have that as those tests are likely not run locally. As the [Contributing Guide][1] explains:

> Some tests require the PyTidyLib library, which depends on the HTML Tidy
library. If you do not have PyTidyLib installed, the tests which depend upon
it will be skipped. Given the difficulty in installing the HTML Tidy library
on many systems, you may choose to leave both libraries uninstalled and
depend on the Travis server to run those tests when you submit a pull
request.

[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/contributing.md#development-environment

setup.py
```diff
@@ -78,6 +78,13 @@
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
     install_requires=['setuptools >= 36'],
+    extras_require={
+        'testing': [
+            'coverage<5.0',
+            'pyyaml',
+            'pytidylib',
```
waylan: Let's not have `pytidylib` here. Most users (and contributors) don't have the underlying c lib installed and the install of the Python lib will fail for them, which would prevent the tests from even beginning to run for them. `pytidylib` must be optional for everyone. We only enforce it's use on Travis.

.travis.yml
```diff
@@ -7,39 +8,41 @@ matrix:
...
   - tox
 
 after_success:
-  - codecov
+  # Upload coverage information for py* tox environments.
+  - |
+    if [ "${TOXENV#py}" != "$TOXENV" ]; then
+      .tox/$TOXENV/bin/coverage xml
+      bash <(curl -s https://codecov.io/bash) -Z -X gcov -X coveragepy -X search -X xcode -X fix -f coverage.xml -F "${TOXENV//-/,}"
+    fi
```
waylan: Sorry, I'm not a codecov expert, but how does this make thing better?

.travis.yml
```diff
@@ -7,39 +8,41 @@ matrix:
...
-    - env: TOXENV=checkspelling
-
-addons:
-  apt:
-    packages:
-    - libtidy-0.99-0
-    - aspell
-    - aspell-en
+      dist: trusty
+    - env: TOXENV=flake8,checkspelling
```
waylan: I definitely do ***not*** want these combined. I like the granularity of each item being a single thing.

waylan (changes requested):
I have comments on a number of things below. For the most part, I don't see how any of this improves anything. To me it mostly seems like changes for changes sake. Some explanations of the reasoning might help.

Regardless, the things I call out need to be fixed.

.travis.yml
facelessuser: I think the idea is to only call codecov upload on cases that actually have coverage data. With that said, nothing will upload on cases without coverage data anyways, and I don't see the need to use the codecov bash script to do this.

blueyed: 1. codecov-bash is more lightweight and meant to be the go-to implementation (less pip-install etc)
2. it will retry on upload failures
3. the condition could be remozed, but the `-Z` will fail the job on failure (which is good I think)

.travis.yml
blueyed: Sure, I've kept it in a separate commit for that reason.

setup.py
blueyed: Haven't rechecked, but initially on Arch Linux without `tidy` being installed only some tests errored because of this, but installation (in tox) went fine.

.travis.yml
blueyed: Ok, this should be installed then globally again.

.travis.yml
facelessuser: From what I understand, the PyPI codecov package also retries on failure and fails the build.  The only other argument is lightweight.  I don't think the python package is heavy either, but yes, you have to install it, but that isn't really a big deal .  I don't really think we are gaining anything with this change.

blueyed: > From what I understand, the PyPI codecov package also retries on failure

Not a few days ago when I looked into it.

> and fails the build

Yes, there is also an option for it.

But OK, it's not critical to have this change.

It's just from my box of things I like to do with Travis configs.

facelessuser: > Not a few days ago when I looked into it.

See here: https://github.com/codecov/codecov-python/blob/master/codecov/__init__.py#L765. Not sure how well it works, but it's technically there.  I guess my point is I've never personally felt the `codecov` PyPI package wasn't fitting my needs, but maybe @waylan feels differently.

blueyed: See also above: https://github.com/codecov/codecov-python/blob/0743daa83647f12ff31b84d07113d2c24c27b924/codecov/__init__.py#L759-L763

*Edit*: that looks good after all, but I certainly have seen it failing (maybe for some 40x then, that wasn't a validation error)?!

setup.py
waylan: Yes, but what about users who run tests on Windows? They will fail every time if an attempt to install `pytidylib` happens. My point is that auto-installing `pytidylib` should only ever happen on Travis. If a user wants to run those tests locally, then they **must** manually install `pytidylib` as documented in the Contributing Guide.

.travis.yml
waylan: Let's revert this. It is much more complicated and offers no additional value AFAICT.

.travis.yml
waylan: I mean `flake8` and `checkspelling` should be separate Travis jobs.

setup.py
blueyed: Well, it was previously installed always already (in tox).
I will remove it from "testing", but keep it in tox then.

waylan (approved):

setup.py
```diff
@@ -78,6 +78,12 @@
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*',
     install_requires=['setuptools >= 36'],
+    extras_require={
+        'testing': [
+            'coverage<5.0',
```
waylan: We previously were using `coverage<4.0` but this is `coverage<5.0`. We added the previous restriction (in 20b17c29) because  Coverage 4.0 dropped support for PY3.2.  However, we have also dropped support for PY3.2 (in d7764996). That being the case, we don't need that restriction. Currently, Coverage is at version 4.5.2, so there is no 5.0. Maybe we should remove the version restriction altogether?

waylan: I also note that Coverage currently supports PY2.6 and Py3.3, neither of which we support. That being the case, I see no need to have a version restriction here.

blueyed: Cool, I've just added/kept it since coverage.py is in alpha already and quite a bit different (but should be compatible).  So this is good to remove.

.travis.yml
blueyed: Just for info: it failed here: https://ci.appveyor.com/project/pytestbot/pytest/builds/20372997/job/vku9wlsxtbjwwk0j#L423

```
      _____          _
     / ____|        | |
    | |     ___   __| | ___  ___ _____   __
    | |    / _ \ / _  |/ _ \/ __/ _ \ \ / /
    | |___| (_) | (_| |  __/ (_| (_) \ V /
     \_____\___/ \____|\___|\___\___/ \_/
                                    v2.0.15
==> Detecting CI provider
    AppVeyor Detected
    Fixing merge commit SHA
==> Preparing upload
    Error running `cd 'C:\projects\pytest' && git ls-files`: None
XX> Skip processing gcov
==> Collecting reports
XX> Searching for reports disabled
    Targeting specific files
    + coverage.xml bytes=1071184
==> Appending environment variables
    + TOXENV
==> Uploading
    .url https://codecov.io
    .query commit=3acfa3abdc7577f97bb40b9790c10d43c0e486e4&branch=master&job=pytestbot%2Fpytest%2F1.0.5975&pr=4409&service=appveyor&build=vku9wlsxtbjwwk0j&slug=pytest-dev%2Fpytest&flags=py27%2Cwindows&package=py2.0.15
    Pinging Codecov...
Error: ('Connection aborted.', RemoteDisconnected('Remote end closed connection without response',))
Tip: See all example repositories: https://github.com/codecov?query=example
Support channels:
  Email:   hello@codecov.io
  IRC:     #codecov
  Gitter:  https://gitter.im/codecov/support
  Twitter: @codecov
```
Ref: https://github.com/codecov/codecov-python/issues/158

PR #757, opened 2018-11-21
Use a PEP562 implementation for deprecating attributes

facelessuser:
Use a vendored Pep562 backport to simulate Python 3.7's new PEP 562 feature. For versions greater than Python3.7, default to the official implementation.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -3,3 +3,4 @@ omit=
     *site-packages*
     tests/*
     markdown/test_tools.py
+    markdown/pep562.py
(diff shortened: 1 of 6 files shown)


facelessuser:
Vendored library is excluded from tests and coverage. Outside the lib, there is no new code that is untested. Coverage is simply failing due to the removal of code from files which changed the percentage of coverage since there is now less code in the file.

facelessuser:
Went ahead and removed the __dir__ override.  We'll just let it return the module __dir__. We aren't gaining anything by having the override currently.

PR #769, opened 2018-12-28
Correctly handle overlapping raw HTML matches

### LINKED ISSUE #458: Placeholder in output for fenced code blocks nested in raw HTML blocks
When using the `markdown.extensions.extra`, syntax looking like a code fence inside of HTML block elements gets rendered as gibberish instead of text, unless the HTML block has `markdown=1` applied.

This test case demonstrates what I mean:
[test.md.txt](https://github.com/waylan/Python-Markdown/files/166209/test.md.txt)

The output I get from parsing that page includes a section like this:

```
<div>

Save this file to `test.md` and then try this:


wzxhzdk:0


</div>
```

The exact contents of the gibberish change, but it typically seems to start with wz... which suggests to me that it's rendering something that's not meant to be a string as a string.

Same behavior on Python 3.5.1, using Python-Markdown 2.6.2 and Python-Markdown 2.6.5, as well as Python 2.7.11 using Python-Markdown 2.6.5. (On two different Arch Linux machines.)

PhilipTrauner:
Should sufficiently fix https://github.com/Python-Markdown/markdown/issues/458.
Instead of rewriting the raw HTML substitution logic I chose to run it recursively until the output is equal to the previous input.

With kinds regards,
Philip Trauner

(and let me know if I missed something 😅 )

### DIFF
--- a/markdown/postprocessors.py
+++ b/markdown/postprocessors.py
@@ -81,9 +81,12 @@ class RawHtmlPostprocessor(Postprocessor):
 
         if replacements:
             pattern = re.compile("|".join(re.escape(k) for k in replacements))
-            text = pattern.sub(lambda m: replacements[m.group(0)], text)
+            processed_text = pattern.sub(lambda m: replacements[m.group(0)], text)
 
-        return text
+        if processed_text == text:
+            return processed_text
+        else:
+            return self.run(processed_text)
 
     def isblocklevel(self, html):
         m = re.match(r'^\<\/?([^ >]+)', html)


waylan (changes requested):
That is one way to fix this. We generally try to avoid recursive code as longer documents have been known to reach Python recursion limit in the past. However, In that case, that seems unlikely.

In addition to actually passing the [tests][0], as per the [Contributing Guide][1], this needs a test case and a comment in the release notes.

[0]: https://travis-ci.org/Python-Markdown/markdown/builds/473152099?utm_source=github_status&utm_medium=notification
[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/contributing.md#pull-requests

PhilipTrauner:
Yep, kinda messed up on the test front and especially didn't consider the "no replacement" case. Will submit a fix tomorrow.

With kind regards,
Philip Trauner

PR #770, opened 2018-12-29
Correctly handle overlapping raw HTML matches (2)

PhilipTrauner:
(Recreated because #769 didn't respect the Contributing Guide)

Instead of rewriting the raw HTML substitution logic I chose to run it recursively until the output is equal to the previous input.

With kinds regards,
Philip Trauner

### DIFF
--- a/docs/change_log/release-3.1.md
+++ b/docs/change_log/release-3.1.md
@@ -31,6 +31,7 @@ The following new features have been included in the release:
 
 The following bug fixes are included in the 3.1 release:
 
+* Overlapping raw HTML matches no longer leave placeholders behind (#458).
 * Emphasis patterns now recognize newline characters as whitespace (#783).
 * Version format had been updated to be PEP 440 compliant (#736).
 * Block level elements are defined per instance, not as class attributes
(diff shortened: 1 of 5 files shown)


waylan (changes requested):
Thanks for the update.

However, I note that you added the test to the [Legacy Tests][1] rather than the new [Markdown TestCase][2]. We would prefer all new tests to use the new framework. Ideally, a new file would be created at [tests/test_syntax/extensions/][3]`fenced_code.py`, which included a `markdown.test_tools.TestCase` subclass with a method to test the relevant behavior.

As explained in #622 the goal is to eventually have all tests implemented in the new framework. Adding tests to the old only doubles the work so we'd prefer all new tests to be int he new framework.

Finally, please don't create a new PR. You may simply push any changes to your branch and GitHub will add them to this PR. We will squash the commits when we merge to master.

[1]: https://python-markdown.github.io/test_tools/#markdowntest_toolslegacytestcase
[2]: https://python-markdown.github.io/test_tools/#markdowntest_toolstestcase
[3]: https://github.com/Python-Markdown/markdown/tree/master/tests/test_syntax/extensions

PR #771, opened 2018-12-30
allow html as permalink

ggueret:
I wanted to use a Font-Awesome icon as permalink for a project. So I defined TOC permalink as follows: `<i class="fas fa-link"></i>`. Currently, `TocTreeprocessor.add_permalink` treats permalink content as an `etree.Element` text, so the html tag will be escaped.

This patch aim to allows both text and html.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -186,9 +186,13 @@ class TocTreeprocessor(Treeprocessor):
 
     def add_permalink(self, c, elem_id):
         permalink = etree.Element("a")
-        permalink.text = ("%spara;" % AMP_SUBSTITUTE
-                          if self.use_permalinks is True
-                          else self.use_permalinks)
+        content = ("%spara;" % AMP_SUBSTITUTE
+                   if self.use_permalinks is True
+                   else self.use_permalinks)
+        if content.startswith("<") and content.endswith(">"):
+            permalink.append(etree.fromstring(content))
+        else:
+            permalink.text = content
         permalink.attrib["href"] = "#" + elem_id
         permalink.attrib["class"] = "headerlink"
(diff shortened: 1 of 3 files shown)


waylan (changes requested):
Interesting idea. I'm not 100% sure we want to do this, but I can see the usefulness of it. What if we allowed custom classes to be assigned to the link itself instead?

Either way, this needs a test and a comment in the release notes before it can be accepted.

waylan:
Interesting change. By accepting a ElementTree object you eliminate the need to parse HTML, which can be problematic if invalid HTML is provided. However, I like this even less from a user's perspective. I still think the proper approach would be to simply add a `permalink_class` setting which defined a HTML class on the `<a>` tag for the permalink.

Then you could easily do `permalink_class="fas fa-link"`, which would give you:

    '<a class="fas fa-link" href="#header" title="Permanent link">&para;</a>'

I believe that would work fine with Font-Awesome's CSS.

Of course the default value of `permalink_class` would be `headerlink` to match the existing behavior.

markdown/extensions/toc.py
```diff
@@ -260,7 +259,7 @@ def run(self, doc):
 
                 if self.use_anchors:
                     self.add_anchor(el, el.attrib["id"])
-                if self.use_permalinks:
+                if self.use_permalinks or isinstance(self.use_permalinks, etree.Element):
```
ggueret, line 262: Needed extra check, because a `bool(etree.Element("a"))` will return False

ggueret:
Thanks @waylan, yeah, it makes a lot more sense to accept an `etree.Element` instance as a permalink value. I added a test and updated the toc doc.

ggueret:
Closed because of irrelevant changes, continued in #776.

PR #772, opened 2019-01-06
CodeHilite: add 'default_lang' option

p-ouellette:
Useful for highlighting a file that does not have fenced code blocks or any of the other ways of specifying a language, and uses a single language. I can document it if there is approval.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -56,10 +56,12 @@ class CodeHilite(object):
     * src: Source string or any object with a .readline attribute.
 
     * linenums: (Boolean) Set line numbering to 'on' (True),
-      'off' (False) or 'auto'(None). Set to 'auto' by default.
+      'off' (False) or 'auto' (None). Set to 'auto' by default.
 
-    * guess_lang: (Boolean) Turn language auto-detection
-      'on' or 'off' (on by default).
+    * default_lang: The language to use if it cannot be determined otherwise.
+
+    * guess_lang: (Boolean) Turn language auto-detection 'on' or 'off' (on by
+      default). Not used if 'default_lang' is set.
 
     * css_class: Set class name of wrapper div ('codehilite' by default).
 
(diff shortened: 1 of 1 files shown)


waylan:
This seems like a bad idea to me. We should never assume a default language. That said, the concern about being able to identify a language is valid when not using fenced code blocks. As it turns out, the codehilite extension already provides two ways to address that:

1. If no language is defined, it will use Pygments' guessing algorithm (unless `guess_lang` is set to `False`). No need for the user to do anything. Of course, for very small snippets, the accuracy if less than desirable, but with a few lines of code, it is usually pretty good in my experience.

2. The document author can define the language of the code using a shebang line as [documented][1] in the codehilite extension documentation. Of course, snippets often do not have shebang lines, so we also support [shebang lines without paths][2] (which get removed from the output) and [three colons][3] (which get removed and do not use line numbers) to give the author more control over the output.

[1]: https://python-markdown.github.io/extensions/code_hilite/#syntax
[2]: https://python-markdown.github.io/extensions/code_hilite/#shebang-no-path
[3]: https://python-markdown.github.io/extensions/code_hilite/#colons

p-ouellette:
Not sure what you mean by 'We should never assume a default language'. The option is unset by default and is only used for code blocks that have no shebang/colons/fences. It allows the user to highlight a file that uses only regular indented code blocks without any shebangs/colons. I'd rather use fenced code blocks than shebangs or the three colons since fenced code blocks are more standard with different markdown implementations. But ideally I'd use neither for the file I'm working with in order to keep it cleaner for reading in text form.

I understand if the option is too specialized but don't see how it's a bad idea.

waylan:
It occurs to me that perhaps you are looking at this from the perspective of a single document where the document author and the library user are the same person. In that case it might make sense. However, I'm looking at this as a setting which is set once for an entire site where the library user and the document author are different people. In that scenario, such a setting would be a bad idea as it could be too much of a surprise to the document author.

We consider Python-Markdown to be a library first and a command-line (or end user) tool second. So any decisions about features are made from that perspective.

In fact, some have argued that the feature of guessing the language suffers from the same problem and should be removed. However, as it was an existing feature, we left it in and added the option to turn it off. However, if it had not already existed, we might have never added it. The same goes for this.

On a personal note, I rarely have ever had a document of my own which contains all one language in the code blocks. But then, they mostly contain a combination of Markdown, HTML and Python which makes sense given the fact I maintain a Python implementation of Markdown. So, while I can't relate to your use case personally, I suppose it could be possible that your documents all contain the same language in all of the code blocks. Regardless, given the lack of requests for the feature, I'm inclined to think that your situation is not the norm.

So, from the 'library first' perspective, it is a bad idea and from the end user tool perspective it is too specialized.

PR #774, opened 2019-01-09
Added options: USE_LETTERS and SHOW_BACKLINKS

wilhelmer:
* USE_LETTERS - Use letters (a-z) instead of numbers as footnote markers.
* SHOW_BACKLINKS - Enable or disable backlinks from the footnote to the reader's place.

### DIFF
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -24,6 +24,7 @@ from .. import util
 from collections import OrderedDict
 import re
 import copy
+import string
 
 FN_BACKLINK_TEXT = util.STX + "zz1337820767766393qq" + util.ETX
 NBSP_PLACEHOLDER = util.STX + "qq3936677670287331zz" + util.ETX
@@ -57,7 +58,15 @@ class FootnoteExtension(Extension):
                  "footnote number."],
             "SEPARATOR":
                 [":",
-                 "Footnote separator."]
+                 "Footnote separator."],
+            'USE_LETTERS':
+                [False,
+                 "Use letters (a-z) instead of numbers"
+                 "as footnote markers."],
+            'SHOW_BACKLINKS':
+                [True,
+                 "Enable or disable backlinks"
+                 "from the footnote to the reader's place."]
         }
         super(FootnoteExtension, self).__init__(**kwargs)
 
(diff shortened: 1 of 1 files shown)


wilhelmer:
Short explanation:
* Using letters is useful, e.g., if you have footnotes on tables with numeric data. With letter symbols, the markers are better separated from numeric data.
* Disabling backlinks is useful, e.g., if you reuse the same footnote marker 15 times and the extension would otherwise add 15 backlink symbols on the same footnote, which would look confusing.

waylan:
My inclination is to not add these features. However, I am still undecided. Note that these are features that have never been requested before, so there is not a high demand for them. And, I like to avoid feature creep. Regardless, I have some issues with each of the specific features as well.

1. Why should the references be limited to ASCII characters? For example, within our tests we have test documents in Russian and Arabic. ASCII characters would never make sense there. 
2. If you are referencing the same footnote 15 different times, I agree that 15 backlink symbols is probably not good. But I don't think disabling all backlinks is the answer either. Perhaps we should only add a backlink for the first reference and leave all others off. I don't think that would even require a setting.

As a final point, this extension primarily is a clone of PHP Markdown Extra's footnote implementation. How does that extension address these matters (I haven't checked)? I'm inclined to do the same here.

wilhelmer:
1. True, a better way to implement this may be some kind of LIST_STYLE_TYPE option, similar to the [css property](https://www.w3schools.com/cssref/playit.asp?filename=playcss_ol_list-style-type), with only a few possible values (lower and upper alpha?) to get started. If desired, other list styles could be added later.
(AFAIK, the PHP Markdown Extra spec does not make any specifications regarding numbering. It defaults to `<ol>` tags for output and states that "a future version may provide a programming interface to allow different markup to be generated.")

2. Leaving all references but the first off would work, too. But doing that without a setting may be unfavorable for users who are referencing the same footnote only 2-3 times and like the additional backlink symbols.

waylan:
> 1. True, a better way to implement this may be some kind of LIST_STYLE_TYPE option, similar to the css property, with only a few possible values (lower and upper alpha?) to get started. If desired, other list styles could be added later.

On the one hand, that feels like more feature creep. On the other, I like the idea. Of course, the problem is that we can't just assign the CSS style to the list, we also need to have the matching characters int he reference, so we need to explicitly add support for each style. And this is where the feature creep comes in. I'm not interested in maintaining a list of all characters in every one of the values supported by CSS. Therefore, my inclination is to say no to this feature in general.

> Leaving all references but the first off would work, too. But doing that without a setting may be unfavorable for users who are referencing the same footnote only 2-3 times and like the additional backlink symbols.

Fair point. Which is (in part) why I'm inclined to leave things as-is.

That fact is, the entire footnotes feature is an extension. Anyone is free to fork the extension and implement whatever changes in behavior they want. Your specific feature requests are certainly edge cases which most users do not need. And you already have working code which meets your needs. 

Given the above, we will not be implement these features at this time. If either of these become common requests in the future. then we may reconsider.

wilhelmer:
Fair enough, I agree and keep my fork :-) Keep up the great work!

PR #776, opened 2019-01-10
allow a custom css class to be applied on anchorlinks and permalinks

ggueret:
This pull request follows on #771 that I closed because it is no longer relevant.

The goal is to allow a custom css class to be applied on permalinks as suggested by @waylan in the original pull request (I unfortunately did not see your answer during my last changes).

It perfectly fulfills my need with font-awesome. Indeed, there is no need for an extra tag and it seems semantically better to apply `fa-` classes to a link that has a title attribute.
So I added the same keyword logic "_class" for the anchorlinks, for that matter.

I fixed the case that if the permalink is an empty string, it is treated as `False` and thus disabled. I guess this is the expected behavior when I look at [this line](https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/toc.py#L135), could you confirm ?

I hope this pull request can be merged as is, I remain available if changes need to be made.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -257,7 +257,7 @@ class TocTreeprocessor(Treeprocessor):
 
                 if self.use_anchors:
                     self.add_anchor(el, el.attrib["id"])
-                if self.use_permalinks:
+                if self.use_permalinks is not False:
                     self.add_permalink(el, el.attrib["id"])
 
         toc_tokens = nest_toc_tokens(toc_tokens)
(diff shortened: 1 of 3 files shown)


markdown/extensions/toc.py
```diff
@@ -256,7 +258,7 @@ def run(self, doc):
 
                 if self.use_anchors:
                     self.add_anchor(el, el.attrib["id"])
-                if self.use_permalinks:
+                if self.use_permalinks is not False:
```
waylan, line 261: I'm not sure I'm comfortable with this change from a backward compatible perspective. It could easily break existing users stuff. As far as I can tell, this change is not necessary for the other changes here. Therefore, let's back it out.

waylan (changes requested):
In addition to the one change requested below, a comment should be added to the release notes. Otherwise this looks good.

markdown/extensions/toc.py
waylan: Upon further review. I now realize that this is necessary to allow an empty permalink, which is desirable for using icon fonts form CSS.

waylan:
This is being closed on favor of #885, which uses this as a base.

PR #778, opened 2019-01-16
Add .keys() method to Registry.

aero31aero:
Adds a `.keys()` method. This would be useful in cases like:

1. Removing all elements from a registry *except* some. example: https://github.com/zulip/zulip/commit/b7c5ae7bca3990421971e93fbe0ce85f1699f347#diff-237f2d7c30a5fdd38c7f1e2d8f6734eeL1751
2. Potentially reordering a registry object from another module without having to read that code to find out which element is at what priority.
3. Helpful in debugging.

### DIFF
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -399,6 +399,16 @@ class RegistryTests(unittest.TestCase):
         self.assertEqual(len(r), 2)
         self.assertEqual(list(r), ['b2', 'a'])
 
+    def testGetKeys(self):
+        r = markdown.util.Registry()
+        r.register(Item('a'), 'a', 20)
+        r.register(Item('c'), 'c', 30)
+        r.register(Item('b'), 'b', 10)
+        r.register(Item('d'), 'd', 40)
+        self.assertEqual(r.keys(), ['d', 'c', 'a', 'b'])
+        r.deregister('c')
+        self.assertEqual(r.keys(), ['d', 'a', 'b'])
+
     def testRegistryDeprecatedAdd(self):
         with warnings.catch_warnings(record=True) as w:
             warnings.simplefilter("always")
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -417,6 +417,11 @@ class Registry(object):
         else:
             raise TypeError
 
+    def keys(self):
+        """ Return all stored keys, sorted by decreasing priority. """
+        self._sort()
+        return [x.name for x in self._priority]
+
     def add(self, key, value, location):
         """ Register a key by location. """
         if len(self) == 0:


tests/test_apis.py
```diff
@@ -399,6 +399,19 @@ def testRegisterDupplicate(self):
         self.assertEqual(len(r), 2)
         self.assertEqual(list(r), ['b2', 'a'])
 
+    def testGetKeys(self):
+        r = markdown.util.Registry()
+        r.register(Item('a'), 'a', 10)
+        r.register(Item('b1'), 'b', 20)
+        r.register(Item('c'), 'c', 30)
```
showell, line 405: I would suggest adding one more key here, and make the calls to register out of order (to prevent this from passing by accidental sorting order).  So add maybe c, d, a, b or something.

showell:
I think you want `keys` to either guarantee a sorting order (in which case you can tighten up the tests a bit) or just return a set.  I think the former makes more sense here?

aero31aero:
Yeah, I just checked the code again, and pushed the unsorted version of the tests. Since dicts aren't guaranteed to keep the order intact, I expect some tests to fail based on implementation (python 3.6 onwards should pass).

waylan:
> 
> 1. Removing all elements from a registry except some. example: zulip/zulip@b7c5ae7#diff-237f2d7c30a5fdd38c7f1e2d8f6734eeL1751

I'm not sure I understand how this would be helpful. The old API required you to confirm that a "name" actually existed before deleting it. However, the new API includes the `strict` argument on the `deregister` method. With `strict=False` you won't get and error if the `name` doesn't exist. Therefore, you don't need to confirm that a processor has previously been registered. Just do:

```python
for k in [my_list_of_processors_to_remove]:
    md.inlinePatterns.deregister(k, strict=False)
```

There is no need to confirm that `name` actually exists. Therefore, there is no need to get a list of keys.

> 2. Potentially reordering a registry object from another module without having to read that code to find out which element is at what priority.

Our documentation suggests that extensions should be documenting the priorities they use. I realize we are not yet following our own advise here, but I intentionally designed the API this way. You should never need to obtain the priority of existing items programmatically. If you think otherwise, then you are doing it wrong.

> 3. Helpful in debugging.

See my comment above. Or did you have something else in mind? Could you provide an example where this would be helpful.

aero31aero:
1. That code is actually `for k not in [my_list_of_processors_to_keep]:`. In that particular instance, we need to disable most of the formatting.

2. Lets say I know that we get an InlineProcessor `x` and `y` such that `x` > `y` in priority. I need to add my own `z` which should be `x` > `z` > `y`. For this, I need to know the priorities of `x` and `y` so that I insert `z` in the correct order. If I'm missing some way around this, do tell.

aero31aero:
@waylan I would like to present our usecase in more detail. We at Zulip try to optimize markdown for chat instead of documents, and have been happily using py-markdown as a base to build upon for all our customizations.

These include removing a lot of the default InlineProcessors, changing the regexes for some of them (example: we disable _ for bold and italics), rendering nested fencing.

Furthermore, our markdown implementation also computes user mentions and alert words among other message properties, storing them alongside the rendered html.

We love py-markdown and the ease of extending it to our needs. Some of us were having a discussion about our short term needs and that discussion turned into a broader conversation on how maybe we use markdown differently from a lot of other projects. We would love to have your inputs and share our thoughts with you (outside of the scope of just this PR)! :D

If you could take some time out, do join us at https://chat.zulip.org/#narrow/stream/68-markdown/topic/migration.20to.20priority.20numbers

Or if you prefer to chat elsewhere, or just continue the conversation here, that would be great, too. And we understand you're busy, so if you respectfully decline, that's fine. Thanks again!

showell:
Hi Waylan, I also work on the Zulip codebase with Rohitt and others.  We've been using the Python-Markdown engine since about 2013 or so, and I'll echo that we do some fairly exotic things on top of what the core engine does.  So I would also welcome the chance to run some things by you that we do in our customization, if only to just share some information.  (I'm also on the chat that Rohitt linked to.)  Thanks for all the great work!

waylan:
> 1. That code is actually `for k not in [my_list_of_processors_to_remove]:`. In that particular instance, we need to disable most of the formatting.

As a potential user of your extension I would take issue with this. Suppose I am using your extension along with my private extension. Of course, you would not include my private extension in your list, making it impossible for the two to co-exist. You should only be deregistering items you know about. In other words, your argument for adding this feature is to allow bad behavior, which I am not inclined to do.

As an aside, I realize that in my example, the two extensions might still be able to co-exist if they are loaded in the correct order (yours first). However, I have always tries to avoid any such (ordering) requirements. In fact, the API change was significantly motivated by a desire to further eliminate such extension loading order restrictions. I'm not inclined to support something that reintroduces such a restriction.

> 2. Lets say I know that we get an InlineProcessor `x` and `y` such that `x` > `y` in priority. I need to add my own `z` which should be `x` > `z` > `y`. For this, I need to know the priorities of `x` and `y` so that I insert `z` in the correct order. If I'm missing some way around this, do tell.

Then get the priority of `x` and `y` from the documentation for `x` and `y` and hardcode your priority for `z`. I never intended for the priority to be determined at run time. Yes, not everyone currently is documenting that, but they should be. The current situation is encouragement to do so.

I can't help but notice that you repeatedly suggest that you are overriding *most* of the processors, which suggests that you are rebuilding nearly all of the parser.  In that case I would suggest subclassing the `markdown.Markdown` class and overriding the `build_parser` method (rather than using an extension). You can build up the parser from scratch with your own set of processors. No need to build the default, tear it down and then rebuild your custom parser. That should improve performance significantly (for first time runs) as well. In that case you then get to assign your own priorities to everything, so no need to determine what priorities are used elsewhere.

showell:
> I can't help but notice that you repeatedly suggest that you are overriding _most_ of the processors, which suggests that you are rebuilding nearly all of the parser. In that case I would suggest subclassing the `markdown.Markdown` class and overriding the `build_parser` method (rather than using an extension). You can build up the parser from scratch with your own set of processors. No need to build the default, tear it down and then rebuild your custom parser. That should improve performance significantly (for first time runs) as well. In that case you then get to assign your own priorities to everything, so no need to determine what priorities are used elsewhere.

Indeed, that does describe our use case.  We were actually independently considering moving to a `build_parser` kind of strategy, so it's good to hear that you endorse it.  (In older versions we were indeed seeing some performance impact of all the dict re-ordering when we launched the engine and did all the inserts/deletes.)

aero31aero:
Closing this. Thanks a lot for the detailed response! We'll give the `build_parser` strategy a go.

PR #784, opened 2019-02-06
autodirection extension 

aliva:
This will add `dir=auto` attribute to each paragraph, it will tell browser to set the direction of paragraph based on the content, this extension will be helpful for users who write right to left text within markdown file

input:

    Text

output:

    <p dir="auto">Text</p>

note that it's just an initial commit, just wanted to know if it's acceptable to be added into this repo, and be informed of other necessary changes

### DIFF
--- a/markdown/extensions/autodirection.py
+++ b/markdown/extensions/autodirection.py
@@ -1,0 +1,37 @@
+"""
+AutoDirection Extension for Python-Markdown
+===========================================
+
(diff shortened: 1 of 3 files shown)


waylan:
Interesting extension. However, I see no reason why this cannot exist as a third party extension. There are many third party extensions listed in the [wiki]. Feel free to add this to the list.

Note, that I am more inclined to remove the extensions we have now from being included with Markdown than I am to include more.

Issue #786, opened 2019-02-07
toc: toc_depth as range


klml:
Currently [toc_depth](https://python-markdown.github.io/extensions/toc/) *defines up to which section level “n” ( \<h1> to <hn>, where 1 <= n <= 6) to include in the Table of Contents.*

Would be nice to have it not only defining *up to which* but *from and up to which* section level to include in the Table of Contents.
So you could include section level only from range ```h2``` to ```h6``` to the Table of Contents.

I could pass in a PR, should I change *toc_depth* to an array or introduce a new config value (e.g. *toc_height*)?

## use case
I want to use ```h1``` as sole [document title](https://www.w3.org/QA/Tips/Use_h1_for_Title) and so it should not be part of the ToC.

waylan:
This may be a duplicate of #637. However that discussion took a lot of turns so it is unclear. Regardless, the [final comment][1] on that issue is relevant:

> I use the meta extension and include the title of the document in the meta-data. My templates then insert that Title into the HTML separate from the body of my document which comes from the Markdown. Therefore, I never get the document title included in my TOC.
> 
> I work on Python-Markdown as a volunteer in my spare time and I don't have any reason to use up my spare time on features I have no use for. Therefore, if this is something you care about, you need to do the work. If someone submits a PR, I'll take the time to review it, and if it passes my review, I'll merge it. Other than that, I have more important things (to me) to work on.

[1]: https://github.com/Python-Markdown/markdown/issues/637#issuecomment-408430235

Issue #637, opened 2018-02-15
Limit the TOC level depth in extensions/toc


the9000:
####  Use case
Take a document, such as an API description, or a document converted from Org mode, with lots of deep-level headers and very short texts under each of them.
A TOC of such a document becomes excessively long and detailed. (My particular case is what mkdocs generates.)

#### A fix
* Add a config parameter that optionally limits the maximum level of headers included in the TOC.
* Filter the levels [where they are generated](https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/toc.py#L242;) according to that parameter.

 I'll gladly make a PR, but would appreciate any advice on the best way to add a config parameter.

waylan:
I'm sure this is a duplicate, but I can't seem to find it, In any event, we have traditionally been resistant to adding more config options. That certainly has been the stated reason for not yet adding this feature in previous discussions (which I can't find now).

Regardless, the TOC level depth is completely controlled by themes from within the templates. Our position in the past has been that you should be able to override this with [theme customization][1]. However, it occurs to me that as themes can now [define their own settings][2] (as of version 0.17), this could be user configurable as a theme setting. I would be willing to consider a PR which added a theme setting to both of the built-in themes.

[1]: http://www.mkdocs.org/user-guide/styling-your-docs/#using-the-theme-custom_dir
[2]: http://www.mkdocs.org/user-guide/custom-themes/#theme-configuration

mitya57:
@waylan Did you realize that this issue is for Python-Markdown, not for mkdocs?

I know that you are a developer of both projects, so this can be quite confusing :)

waylan:
Oh whoops. Thanks for the headsup @mitya57.

waylan:
Interestingly, the OP made this statement:

> (My particular case is what mkdocs generates.)

So, in the end, I did point to the correct way to address this in MkDocs. Although, by default, the builtin MkDocs themes limit the depth to 2 so I'm not sure why you're having a problem with excessively deep TOC. If you are using a third-party theme, then take that up with the devs of that theme. Otherwise, you are free to submit a PR to MkDocs as stated above.

I'm not so sure I want to add that option to the Markdown extension.

ghost:
It's quite common to use markdown headings for structure and viewability, including deeper level headings however in the TOC you may not want to call them all out.  I have a document that has H1-H6, but I only want to show the higher level headings in the TOC, h1-h4.

RoyiAvital:
[...]

PR #787, opened 2019-02-08
adding toc_height to exclude upper headings, proposal for #786 and #637

klml:
I added a new config value toc_height. Using an array for _toc_depth_  seems to me very unlikely. 

## test local

```
~/devel/markdown$ python3 -m unittest discover tests
... 550 tests in 1.259s
OK (skipped=71)

```

### DIFF
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -199,8 +199,8 @@ The following options are provided to configure the output:
 * **`toc_depth`**
     Define the range of section levels to include in the Table of Contents.
     A single integer (b) defines the bottom section level (<h1>..<hb>) only.
-    Two digits separated by a hyphen in between (```2-5```), define the
-    top (t) and the bottom (b) (<ht>..<hb>). Defaults to `6` (bottom).
+    A string consisting of two digits separated by a hyphen in between ("2-5"),
+    define the top (t) and the bottom (b) (<ht>..<hb>). Defaults to `6` (bottom).
 
     When used with conjunction with `baselevel`, this parameter will not
     take the fitted hierarchy from `baselevel` into account. That is, if
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -196,21 +196,16 @@ The following options are provided to configure the output:
 * **`separator`**:
     Word separator. Character which replaces white space in id. Defaults to "`-`".
 
-* **`toc_height`**
-    Define the highest section level "n" (`<hn>` to `<h6>`, where `1 <= n <= 6`)
-    to include in the Table of Contents. Section levels above are omitted.
-    Defaults to `1`.
+* **`toc_depth`**
+    Define the range of section levels to include in the Table of Contents.
+    A single integer (b) defines the bottom section level (<h1>..<hb>) only.
+    Two digits seperated by a hyphen in between (```2-5```), define the
+    top (t) and the bottom (b) (<ht>..<hb>). Defaults to `6` (bottom).
 
     When used with conjunction with `baselevel`, this parameter will not
-    take the fitted hierarchy from `baselevel` into account.
-    If you set `baselevel` to 3 and `toc_height` to 2, the *first* headline
-    will be `<h3>` and so still included in the Table of Contents. To exclude
-    this first level, you have to set `toc_height` to 4.
-
-* **`toc_depth`**
-    Define up to which section level "n" (`<h1>` to `<hn>`, where `1 <= n <= 6`)
-    to include in the Table of Contents. Defaults to `6`.
-
-    When used with conjunction with `baselevel` this parameter will limit the
-    resulting (adjusted) heading. That is, if both `toc_depth` and `baselevel`
-    are 3, then only the highest level will be present in the table.
+    take the fitted hierarchy from `baselevel` into account. That is, if
+    both `toc_depth` and `baselevel` are 3, then only the highest level
+    will be present in the table. If you set `baselevel` to 3 and
+    `toc_depth` to '2-6', the *first* headline will be `<h3>` and so still
+    included in the Table of Contents. To exclude this first level, you
+    have to set `toc_depth` to '4-6'.
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -196,6 +196,14 @@ The following options are provided to configure the output:
 * **`separator`**:
     Word separator. Character which replaces white space in id. Defaults to "`-`".
 
+* **`toc_height`**
+    Define from to which section level "n" (`<hn>` to `<h6>`, where `1 <= n <= 6`)
+    to include in the Table of Contents. Defaults to `1`.
+
+    When used with conjunction with `baselevel` this parameter will not follow.
+    To exclude the first level, you have to add the count of `baselevel` to
+    `toc_height`.
+
 * **`toc_depth`**
     Define up to which section level "n" (`<h1>` to `<hn>`, where `1 <= n <= 6`)
     to include in the Table of Contents. Defaults to `6`.
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -199,7 +199,7 @@ The following options are provided to configure the output:
 * **`toc_depth`**
     Define the range of section levels to include in the Table of Contents.
     A single integer (b) defines the bottom section level (<h1>..<hb>) only.
-    Two digits seperated by a hyphen in between (```2-5```), define the
+    Two digits separated by a hyphen in between (```2-5```), define the
     top (t) and the bottom (b) (<ht>..<hb>). Defaults to `6` (bottom).
 
     When used with conjunction with `baselevel`, this parameter will not
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -197,12 +197,15 @@ The following options are provided to configure the output:
     Word separator. Character which replaces white space in id. Defaults to "`-`".
 
 * **`toc_height`**
-    Define from to which section level "n" (`<hn>` to `<h6>`, where `1 <= n <= 6`)
-    to include in the Table of Contents. Defaults to `1`.
+    Define the highest section level "n" (`<hn>` to `<h6>`, where `1 <= n <= 6`)
+    to include in the Table of Contents. Section levels above are omitted.
+    Defaults to `1`.
 
-    When used with conjunction with `baselevel` this parameter will not follow.
-    To exclude the first level, you have to add the count of `baselevel` to
-    `toc_height`.
+    When used with conjunction with `baselevel`, this parameter will not
+    take the fitted hierarchy from `baselevel` into account.
+    If you set `baselevel` to 3 and `toc_height` to 2, the *first* headline
+    will be `<h3>` and so still included in the Table of Contents. To exclude
+    this first level, you have to set `toc_height` to 4.
 
 * **`toc_depth`**
     Define up to which section level "n" (`<h1>` to `<hn>`, where `1 <= n <= 6`)
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -298,8 +298,8 @@ class TocExtension(Extension):
                         "Defaults to the headerid ext's slugify function."],
             'separator': ['-', 'Word separator. Defaults to "-".'],
             "toc_height": [1,
-                          "Define from to which section level n (<hn>..<h6>) to "
-                          "include in the TOC"],
+                           "Define from to which section level n (<hn>..<h6>) to "
+                           "include in the TOC"],
             "toc_depth": [6,
                           "Define up to which section level n (<h1>..<hn>) to "
                           "include in the TOC"]
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -298,8 +298,8 @@ class TocExtension(Extension):
                         "Defaults to the headerid ext's slugify function."],
             'separator': ['-', 'Word separator. Defaults to "-".'],
             "toc_height": [1,
-                           "Define from to which section level n (<hn>..<h6>) to "
-                           "include in the TOC"],
+                           "Define the highest section level n (<hn>..<h6>) to"
+                           "include in the TOC. Section levels above are omitted."],
             "toc_depth": [6,
                           "Define up to which section level n (<h1>..<hn>) to "
                           "include in the TOC"]
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -135,12 +135,11 @@ class TocTreeprocessor(Treeprocessor):
         if self.use_permalinks is None:
             self.use_permalinks = config["permalink"]
         self.header_rgx = re.compile("[Hh][123456]")
-        self.toc_depth = config["toc_depth"]
-        if type(self.toc_depth) is not int:
-            self.toc_top, self.toc_bottom = self.toc_depth.split('-')
+        if isinstance(config["toc_depth"], basestring) and '-' in config["toc_depth"]:
+            self.toc_top, self.toc_bottom = [int(x) for x in config["toc_depth"].split('-')]
         else:
             self.toc_top = 1
-            self.toc_bottom = self.toc_depth
+            self.toc_bottom = int(config["toc_depth"])
 
     def iterparent(self, node):
         ''' Iterator wrapper to get allowed parent and child all at once. '''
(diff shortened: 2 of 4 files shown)


markdown/extensions/toc.py
```diff
@@ -295,6 +297,9 @@ def __init__(self, **kwargs):
                         "Function to generate anchors based on header text - "
                         "Defaults to the headerid ext's slugify function."],
             'separator': ['-', 'Word separator. Defaults to "-".'],
+            "toc_height": [1,
+                          "Define from to which section level n (<hn>..<h6>) to "
+                          "include in the TOC"],
```
waylan: There must be a typo here. "from to which" can' the correct. I'm not even sure what it is supposed to say.

docs/extensions/toc.md
```diff
@@ -196,6 +196,14 @@ The following options are provided to configure the output:
 * **`separator`**:
     Word separator. Character which replaces white space in id. Defaults to "`-`".
 
+* **`toc_height`**
+    Define from to which section level "n" (`<hn>` to `<h6>`, where `1 <= n <= 6`)
+    to include in the Table of Contents. Defaults to `1`.
```
waylan: There must be a typo here. "from to which" can' the correct. I'm not even sure what it is supposed to say.

docs/extensions/toc.md
```diff
@@ -196,6 +196,14 @@ The following options are provided to configure the output:
...
+
+    When used with conjunction with `baselevel` this parameter will not follow.
+    To exclude the first level, you have to add the count of `baselevel` to
+    `toc_height`.
```
waylan: Sorry, but I'm struggling to understand what this is talking about. What does "will not follow" mean? I could guess, but its not clear.

waylan (changes requested):
Thanks for your contribution.

 In addition to the concerns mentioned below, I'm not sure "height" is the right word to describe this feature. To me a "height" of `4` means there are four levels, but that is not what the example int he tests shows. Perhaps that is why I'm struggling to understand the documentation. I don't actually understand what you are trying to do.

Both examples in the tests simply skip level 1. Perhaps one of them should do something other than that. As  reminder, this is a feature that I find completely unnecessary, so perhaps I'm missing something obvious here. But if I am, then other users could be as well, which suggests the name and/or docs could be improved.

markdown/extensions/toc.py
waylan: Also, you are getting a [flake8 error][1] here. Note that the [Contributing Guidelines][2] define how to format multi-line lists. I realize the other items in the file don't match that style, but they predate the guidelines, and as the guidelines state:

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/490761623
[2]: https://github.com/Python-Markdown/markdown/blob/master/docs/contributing.md#code-style-guide

klml:
@waylan thanks for your quick response.

You are right about the word height. I was unsure about the naming of ```toc_height```. I wanted to name it the [opposite of toc_depth](https://www.wordhippo.com/what-is/the-opposite-of/depth.html), and I was undecided between ```toc_height``` and ```toc_top```.
You mentioned ```skip```, but toc_depth also skips headings (at the bottom). Do you have an idea what word would be meaningful?


Sorry for all these other issues, I will get into it and get back to you with an update on my PR.

klml:
I fixed the flake error and changed the first test to a higher toc_height.
I tried to improve the wording. But I have a little subjective arbitrability, so do you think it is now more understandable?

I am still struggling with the parameters name ```toc_height```. My alternative shortlist:


* `toc_height` is missleading, 'a "height" of `4` means there are four levels'. 
* `toc_skip` toc_depth also skips headings (at the bottom). 
* `toc_top` 
* `toc_edge` 
* `toc_begin` 
* `toc_baselevel` is very long and potentialy mixup with baselevel 
* `toc_startlevel` is very long and potentialy mixup with baselevel 


@waylan I know you find this feature "completely unnecessary", so I am very glad that you are dealing with me;)
But I think you, as maintainer, should decide which phrase fits best in the whole project.

waylan:
Your recent changes are a big improvement. They helped bring the new feature into focus for me. Thank you.

I agree that `toc_height` feels like the wrong name for this. Personally I would favor `toc_top`.

However, adding this feature introduces another problem, in that `toc_depth` is now misleading. A user might reasonably assume that the word "depth" refers to the total number of levels. So for example, she might expect the following settings: `toc_top=2, toc_depth=2` to result in a TOC which included levels `h2` and `h3` for a **depth of 2**. But, in fact, `toc_depth` limits the bottom to `h2` and the "depth" (or number of levels) of the TOC is no longer `2`, but `1`. Previously, when the "top" couldn't be altered, "depth" and "bottom" were always synonymous. But once the "top" is altered, "depth" and "bottom" are not the same any more. Therefore, `toc_depth` is a misleading name for the setting as it currently behaves.

I see a few ways forward:

1. Alter `toc_depth` to be the actual "depth" (number of levels) of the TOC. Perhaps do something like `bottom = top + depth - 1` (so for `toc_top=2, toc_depth=2`, `bottom` would be `3`).  For existing users who are not using `toc_top`, this changes nothing in the actual output. And as `toc_top` is new, the new behavior of `toc_depth` will be taken into consideration for any users of the new `toc_top` setting. 
2. Leave `toc_depth` behavior as-is, but deprecate the name and rename it to `toc_bottom`. This would be the most disruptive to existing users, which makes it my least favorite option. But having the two settings names "top" and "bottom" is attractive. 
3. Implement a range instead. If `toc_depth` is an integer, maintain the old behavior. If `toc_depth` is a range (`'t-b'` where `t` and `b` are integers), then treat the range as top and bottom.

Both for options 2 and 3, we would also need to handle the case where a user assigns values such that `bottom < top`. Of course, that would be an error. But in option 1 that is never a factor. As long as `depth` is any whole number between `1-6`, we will always get `bottom > top`. Of course, as HTML never drops below `h6` we don't actually care if `bottom > 6`.

waylan (approved):

klml:
I like 3. best. With `toc_depth` optional as integer (as it was) or as range.

But ```t-b``` is not actually a data type, but a division. Could I use a tuple (```toc_depth = 2, 6```) instead in this case?

waylan:
As a reminder, users don't always provide config settings in Python. Sometimes they might be using a YAML file or similar. So specifically requiring a Python `tuple` might be a stretch. However, any iterable of length 2 might work.

Personally I was thinking a string with the format `"t-b"` would make the most sense. If an integer is provided, do old behavior. If a string is provided, it must be two integers separated by a hyphen. Anything else would be an error.

docs/extensions/toc.md
```diff
@@ -196,21 +196,16 @@ The following options are provided to configure the output:
...
-    If you set `baselevel` to 3 and `toc_height` to 2, the *first* headline
-    will be `<h3>` and so still included in the Table of Contents. To exclude
-    this first level, you have to set `toc_height` to 4.
-
 * **`toc_depth`**
-    Define up to which section level "n" (`<h1>` to `<hn>`, where `1 <= n <= 6`)
-    to include in the Table of Contents. Defaults to `6`.
+    Define the range of section levels to include in the Table of Contents.
+    A single integer (b) defines the bottom section level (<h1>..<hb>) only.
+    Two digits seperated by a hyphen in between (```2-5```), define the
```
waylan: I'm assuming you are trying to indicate that the range should be wrapped in quotes. However, triple backpacks won't give you that. You need to actually use quotes: `"2-5"`. And as a string is required, the docs should probably mention that the range needs to be a string consisting of "Two digits separated by a hyphen".

Also it is spelled "sep**a**rated", not "seperated."

markdown/extensions/toc.py
```diff
@@ -236,8 +235,12 @@ def run(self, doc):
         for el in doc.iter():
             if isinstance(el.tag, string_type) and self.header_rgx.match(el.tag):
                 self.set_level(el)
-                if int(el.tag[-1]) < int(self.toc_height) or int(el.tag[-1]) > int(self.toc_depth):
-
+                if type( self.toc_depth ) != int:
```
waylan: Note that `self.toc_depth` could reasonably be a string which contains a single integer. Testing for an `int` here will fail in many cases. You probably need to do something like:

```python
try:
    toc_bottom = int(self.toc_depth)
    toc_top = 1
except ValueError:
    toc_top, toc_bottom = [int(t), int(b) for t, b in self.toc_depth.split('-')]
```

And that last line should perhaps be wrapped in its own try/except block to catch and raise an error if a bad range is provided.

markdown/extensions/toc.py
```diff
@@ -297,12 +300,13 @@ def __init__(self, **kwargs):
...
-                           "Define the highest section level n (<hn>..<h6>) to"
-                           "include in the TOC. Section levels above are omitted."],
             "toc_depth": [6,
-                          "Define up to which section level n (<h1>..<hn>) to "
-                          "include in the TOC"]
+                          "Define the range of section levels to include in"
+                          "the Table of Contents. A single integer (b) defines"
+                          "the bottom section level (<h1>..<hb>) only."
+                          "Two digits seperated by a hyphen in between"
+                          " (```2-5```), define the top (t) and the bottom (b)"
```
waylan: See my previous comment about including quotes to indicate the range should be a string.

klml:
I implemented the optional range for toc_depth. I tried to find a more elegant method than two if statements, but I hope this is ok.

waylan (changes requested):
Just a few tweaks and this should be good.

klml:
Using "try" is a good idea, sorry I missed this before. 

* I separated and duplicated the continue if clause, so we don't need to define toc_top = 1 in the integer case. 
* I moved the split from the loop over all headings to outside. 
* I defined the except for a TypeError and a ValueError. Should I use an global except to catch both?

I also thought about using the AttributeError with the integer case:

```
try:
    toc_top, toc_bottom = self.toc_depth.split('-')
    if int(el.tag[-1]) < int(toc_top) or int(el.tag[-1]) > int(toc_bottom):
        continue
except AttributeError:
    if int(el.tag[-1]) > int(self.toc_depth):
        continue
except ValueError:
    pass
```
I am not sure whats better


> And that last line should perhaps be wrapped in its own try/except block to catch and raise an error if a bad range is provided.

At the moment I have no error handling, but a wrong value just gives you the whole Table of Contents and behaves like the toc_depth is not set at all.

I think a warning should be very useful. But as long as I can see, there is no explicit error or warning-handling in Python-Markdown. When I set toc_depth to "foobar" in the existing code I get ```ValueError: invalid literal for int() with base 10: 'foobar'```. Or did I misunderstood you? 
When I have 



Thank you for your patience with this issue;)

klml:
Using "try" is a good idea, sorry I missed this before. 

* I separated and duplicated the continue if clause, so we don't need to define toc_top = 1 in the integer case. 
* I moved the split from the loop over all headings to outside. 
* I defined the except for a TypeError and a ValueError. Should I use an global except to catch both?


I also thougt about using the AttributeError with the integer case:

```
try:
    toc_top, toc_bottom = self.toc_depth.split('-')
    if int(el.tag[-1]) < int(toc_top) or int(el.tag[-1]) > int(toc_bottom):
        continue
except AttributeError:
    if int(el.tag[-1]) > int(self.toc_depth):
        continue
except ValueError:
    pass
```
I am not sure what is better.


> And that last line should perhaps be wrapped in its own try/except block to catch and raise an error if a bad range is provided.

At the moment I have no error handling, but a wrong value just gives you the whole Table of Contents and behaves like the toc_depth is not set at all.

I think a warning should be very useful. But as long as I can see, there is no explicit error or warning-handling in Python-Markdown. When I set toc_depth to "foobar" in the existing code I get ```ValueError: invalid literal for int() with base 10: 'foobar'```. Or did I missunderstood you? 
When I resolve this, I get also these codecov.io errors.

Thank you for your patience with this issue;)

markdown/extensions/toc.py
```diff
@@ -136,6 +136,8 @@ def __init__(self, md, config):
             self.use_permalinks = config["permalink"]
         self.header_rgx = re.compile("[Hh][123456]")
         self.toc_depth = config["toc_depth"]
+        if type(self.toc_depth) is not int:
+            self.toc_depth = self.toc_depth.split('-')
```
waylan: If we're going to split the range here, why not due the type conversion here as well. When we don't need all the try/except blocks later. I realize this is a different suggestion than previously, but previously I envisioned a couple lines at the top of the block of code which did the type conversion, then the actual conditionals wouldn't be so complicated.

Also wherever we do the type conversion we should define `top` and `bottom` as internal vars for later use in the conditionals. That should eliminate the need for try/except blocks in the conditionals.

waylan (changes requested):

markdown/extensions/toc.py
klml: I implemented your suggestions and now this looks much simpler for me.
Without the try/except blocks we lose catching "bad range errors". I am good with that.

markdown/extensions/toc.py
```diff
@@ -136,6 +136,11 @@ def __init__(self, md, config):
             self.use_permalinks = config["permalink"]
         self.header_rgx = re.compile("[Hh][123456]")
         self.toc_depth = config["toc_depth"]
+        if type(self.toc_depth) is not int:
```
waylan: First of all, we don't need to assign `self.toc_depth` as we don't need it anymore. `self.toc_top` and `self.toc_bottom` serve that purpose just fine. In the next few lines we can just reference `config["toc_depth"]`.

Sorry I missed this last time, but `if type(foo) is bar` is not the correct way to do a type test in Python. You should use `if isinstance(foo, bar)`, which allows users to pass in subclasses of the specified type (in other words, support duck-typing).

Thirdly, this does not account for `self.toc_depth` being a single integer within a string (`"2"`). 

There are a couple ways we could address this:

1. We could do what I suggested the first time:

    ```python
    try:
         self.toc_bottom = int(config["toc_depth"])
         self.toc_top = 1
    except ValueError:
         self.toc_top, self.toc_bottom = [int(x) for x in config["toc_depth"].split('-')]
    ```

2. Or we could check for a hyphen:

    ```python
    if isinstance(config["toc_depth"], basestring) and '-' in config["toc_depth"]:
        self.toc_top, self.toc_bottom = [int(x) for x in config["toc_depth"].split('-')]
    else:
        self.toc_top = 1
        self.toc_bottom = int(config["toc_depth"])

Notice that on both solutions, we explicitly force all values to integers. Therefore, we do not need to do that later when we use the values.

As it turns out, I like the second solution better as it keeps the old behavior and simply does a conditional check for the new syntax. Any bad data will still act exactly the same as it would have before the change. That's also why I changed it to check for a `basestring`. All string subclasses will be accepted as a range if they contain a hyphen. Whereas all other types are presumed to be an integer (the old behavior).

waylan: I'd make these changes myself, but you used your `master` branch which makes it difficult for me to push any changes to (and I'd rather not open a new PR). If you apply the following diff exactly as-is, I'll merge this PR:

```diff
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -135,12 +135,11 @@ class TocTreeprocessor(Treeprocessor):
         if self.use_permalinks is None:
             self.use_permalinks = config["permalink"]
         self.header_rgx = re.compile("[Hh][123456]")
-        self.toc_depth = config["toc_depth"]
-        if type(self.toc_depth) is not int:
-            self.toc_top, self.toc_bottom = self.toc_depth.split('-')
+        if isinstance(config["toc_depth"], basestring) and '-' in config["toc_depth"]:
+            self.toc_top, self.toc_bottom = [int(x) for x in config["toc_depth"].split('-')]
         else:
             self.toc_top = 1
-            self.toc_bottom = self.toc_depth
+            self.toc_bottom = int(config["toc_depth"])
 
     def iterparent(self, node):
         ''' Iterator wrapper to get allowed parent and child all at once. '''
@@ -240,7 +239,7 @@ class TocTreeprocessor(Treeprocessor):
         for el in doc.iter():
             if isinstance(el.tag, string_type) and self.header_rgx.match(el.tag):
                 self.set_level(el)
-                if int(el.tag[-1]) < int(self.toc_top) or int(el.tag[-1]) > int(self.toc_bottom):
+                if int(el.tag[-1]) < self.toc_top or int(el.tag[-1]) > self.toc_bottom:
                     continue
                 text = ''.join(el.itertext()).strip()
```

klml: @waylan thanks for your help. I used your suggestions, but it failed on

```
if isinstance(config["toc_depth"], basestring) and '-' in config["toc_depth"]:
```
 with: _NameError: name 'basestring' is not defined_

'basestring' is python2 only, can I use [string_types from six](https://pythonhosted.org/six/#six.string_types) instead?

PS: I don't want to push non working code, so I will wait for you.

waylan: Sorry, I forgot about that. six is why overkill for what we need. You can use `markdown.util.string_type`, which is already imported by the module as `string_type`. So just replace `basestring` with `string_type`.

klml: @waylan Oh i missed looking into markdown/util.py, this is great;) I committed your suggestions.

waylan (approved):

waylan:
This looks good. Thank you. It just needs an item added to the [3.1 release notes][1] under "New Features".

[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/release-3.1.md

PR #801, opened 2019-03-07
have treeprocessors always use the current inlinePatterns

adambyer:
This update forces treeprocessor->InlineProcessor to use the current `inlinePatterns`. Before this, `InlineProcessor.inlinePatterns` would always point to the value set initially, even if `Markdown.inlinePatterns` changed. This allows you to register only the processors you want, instead of having to deregister everything, like so...

```
class MyExtension(Extension):
        md.inlinePatterns = util.Registry()
        md.inlinePatterns.register(SimpleTagInlineProcessor(STRONG_RE, 'strong'), 'strong', 40)
```

### DIFF
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -74,7 +74,6 @@ class InlineProcessor(Treeprocessor):
                                       + len(self.__placeholder_suffix)
         self.__placeholder_re = util.INLINE_PLACEHOLDER_RE
         self.md = md
-        self.inlinePatterns = md.inlinePatterns
         self.ancestors = []
 
     @property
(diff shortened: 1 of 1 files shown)


facelessuser:
What is the use case for this?

waylan:
I am inclined to not accept this change. 

The primary concern is that extensions are expected to work with other extensions. Most extensions assume the default patterns are there. And as it's tedious to remove all patterns, most extensions don't do that. This is more likely to break existing extensions.

That said, it is completely legitimate to want to start over with a fresh new set of patterns. If that is what you want, then you should subclass the `markdown.Markdown` class and override the `build_parser` method. This will also be a clue that many of the existing extensions might not work as expected.

adambyer:
My use case is that I need to only allow a few of the patterns. All others should not be processed.

It seems like a bug to me. If I reset `inlinePatterns` I would expect only the patterns I add to be applied. But, currently all default patterns are still applied, which seems counterintuitive and actually took me awhile to figure out what was going on.

I guess I don't fully understand the reasoning behind having it setup the way it is currently.

Overriding `build_parser` does work. But it just seemed odd to have to reset `treeprocessors` also even though my intention is only to reset `inlinePatterns`...

```
super(MyMarkdown, self).build_parser()
self.inlinePatterns = util.Registry()
self.treeprocessors = build_treeprocessors(self)
```

Issue #780, opened 2019-01-27
Raw HTML is throwing an exception


ikus060:
The following mardown: 
```
<div class="row" markdown="1">
<div class="col-md-6" markdown="1">
**SomeText**
</div>

<div class="col-md-6" markdown="1">

**blod text**  
<small>(<i class="fa fa-arrow-left"></i> small)</small>

<div class="barchart" markdown="1">
* item1
* item2
</div>

more text

</div>
</div>
```
is failing with this:
```
Traceback (most recent call last):
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/test_tools.py", line 117, in test
    output = markdown(input, **kwargs)
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/core.py", line 391, in markdown
    return md.convert(text)
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/core.py", line 268, in convert
    root = self.parser.parseDocument(self.lines).getroot()
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/blockparser.py", line 92, in parseDocument
    self.parseChunk(self.root, '\n'.join(lines))
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/blockparser.py", line 107, in parseChunk
    self.parseBlocks(parent, text.split('\n\n'))
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/blockparser.py", line 125, in parseBlocks
    if processor.run(parent, blocks) is not False:
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/extensions/extra.py", line 127, in run
    block = self._process_nests(element, block)
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/extensions/extra.py", line 95, in _process_nests
    block[nest_index[-1][1]:], True)                      # nest
  File "/home/ikus060/workspace/PDSL/markdown.git/markdown/extensions/extra.py", line 101, in run
    tag = self._tag_data[self.parser.blockprocessors.tag_counter]
IndexError: list index out of range

```

waylan:
Which version of Markdown are you using and what arguments are being passed to `markdown`?

waylan:
I am seeing the same error:

```
>>> import markdown
>>> markdown.__version__
'3.1.dev0'
>>> markdown.markdown(src, extensions=['extra'])
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
[...]

Issue #595, opened 2017-11-15
Additional paragraph when using Markdown in raw HTML


jannschu:
First thank you for this implementation! 👍 

I provide a minimal working example of some unexpected behavior I encountered:

```python
import markdown
md = markdown.Markdown(extensions=['extra'])
html = md.convert("<div markdown><p>Hello _World!_</p></div>")
print(html)
```

The output was
```html
<div>
<p><p>Hello <em>World!</em></p></p>
</div>
```

But I expected
```html
<div>
<p>Hello <em>World!</em></p>
</div>
```

My workaround ~~is to use two line breaks~~ (edit)
```python
html = md.convert("<div markdown>\n\n<p>Hello _World!_</p></div>")
```
does not include the additional paragraph, but the Markdown is not replaced.

waylan:
Actually this is the correct behavior. The problem is that the Markdown-in-raw-HTML behavior is not well documented. 

For example, it may help to understand that a `markdown="1"` attribute does not apply to the nested `<p>` element. It only applies to the `<div>` it is assigned to. Any content inside the `<div>` is now regular Markdown content and must follow the rules of Markdown.

As a reminder, the [rules][1] state:

> The only restrictions are that block-level HTML elements — e.g. `<div>`, `<table>`, `<pre>`, `<p>`, etc. — must be separated from surrounding content by blank lines, and the start and end tags of the block should not be indented with tabs or spaces. Markdown is smart enough not to add extra (unwanted) `<p>` tags around HTML block-level tags.

That rule is strictly enforced. You only avoid the extra (unwanted) `<p>` tags if those three conditions are met (block-level element, blank lines and no indentation). Of course, if you add a `markdown="1"` attribute to a tag, then the rule doesn't apply to that tag, but it still applies to any nested tags.

Therefore, this is what you want as input:

```html
<div markdown="1">

<p  markdown="1">Hello _World!_</p>

</div>
```

[1]: https://daringfireball.net/projects/markdown/syntax#html

jannschu:
I had already suspected that this might be an user error. Thanks for your answer with details about the background.

jannschu:
Is the following expected to work?

```python
html = md.convert("""<div markdown="1">

<div markdown="1">

<p markdown="1">Hello _World!_<p>

</div>

</div>""")
```

I get an `IndexError` running this.

Edit: This is #584 probably.

waylan:
[...]

PR #803, opened 2019-03-07
Refactor HTML Parser

waylan:
This is experimental. More of the HTMLParser methods need to be fleshed out. So far the basic stuff works as long as there is no invalid HTML in the document (which is untested at this point).

## Input:

```md
Some *Markdown* text.

<p>Some *raw* HTML</p>

<span>*inline*</span>

    <p>code block</p>

`<em>code span</em>`

<div>

foo *bar*

* baz bar

blah blah

</div>

More *Markdown*.
```

## Output:

```html
<p>Some <em>Markdown</em> text.</p>
<p>Some *raw* HTML</p>

<p><span><em>inline</em></span></p>
<pre><code>&lt;p&gt;code block&lt;/p&gt;
</code></pre>
<p><code>&lt;em&gt;code span&lt;/em&gt;</code></p>
<div>

foo *bar*

* baz bar

blah blah

</div>

<p>More <em>Markdown</em>.</p>
```

... which exactly matches the existing behavior.

I havn't actually run the tests on this yet, so I'm curious to see what Travis says...


facelessuser:
I was considering rewriting this myself to fix some of this stuff once and for all, so I'm glad to see the work done here. I'll have to take a look at this and see how it handles or see if there are additional things we need to address. It seems you simplified a lot of the code which is a good thing.

waylan:
Added some tests using the new framework. The new tests include comments on things we might want to reevaluate.

Also, one of the new tests is failing and needs fixed. The plan is to continue adding  new tests and fix various edge cases until all the old tests pass At that point, the old tests should be redundant and can be removed.

facelessuser:
Here's a question. Why don't we build `markdown="1"` functionality right into the HTML parser?  I guess you could still have an extension that enables the functionality, but it seems it would be simpler and more straight forward to just build it right in with some kind of toggle.

facelessuser:
Maybe the new `extra` would simply just replace the default HTML parser instance with a new instance that enables `markdown=1` feature.  No need to add a separate plugin that has to intercept blocks and then modify a difficult to understand tracking object.  I think that would be easier to maintain.

waylan:
I've rounded out the tests of valid raw HTML blocks. I have yet to address any of the failing tests. And I haven't even looked at invalid raw HTML.

However, I have questions regarding what the correct behavior should be in comments associated with various tests in the [test file][1]. Any feedback on those would be appreciated.

[1]: https://github.com/Python-Markdown/markdown/pull/803/files#diff-c7451b8f01beebb43e0258539a704255

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+    # TODO: Decide correct behavior. This matches current behavior and Commonmark.
+    # The reference implementation considers this inline not block level:
+    #
+    #   <p><!DOCTYPE html><em>bar</em></p>
+    #
+    # But most implementations do this instead:
+    #
+    #   <p>&lt;!DOCTYPE html&gt;<em>bar</em></p>
+    #
+    # Either makes sense, but the later seems more correct to me.
```
mitya57: Doctype not at the beginning of document is definitely invalid, so I vote for the latter variant.

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+                More <em>Markdown</em> text.</p>
+                """
+            )
+            # The reference implementation does this instead:
+            # self.dedent(
+            #     """
+            #     <p>Some <em>Markdown</em> text.</p>
+            #     <p>*Raw* HTML.</p>
+            #     <p>More <em>Markdown</em> text.</p>
+            #     """
```
mitya57: I think our current behavior is fine.

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+            '<p>A *raw* paragraph.</p>',
+            '<p>A *raw* paragraph.</p>'
+        )
+
+    def test_raw_indent_one_space(self):
+        self.assertMarkdownRenders(
+            ' <p>A *raw* paragraph.</p>',
+            # TODO: reevaluate. This matches strict rules and reference
+            # implementation version 1.0.1 but not 1.0.2b8.
+            '<p><p>A <em>raw</em> paragraph.</p></p>'
```
mitya57: I think it will be better to avoid nested `<p>`s if it won't introduce much complication to our logic.

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+        )
+
+    # TODO: Decide behavior here. Python-Markdown current outputs:
+    #
+    #   <!-- *foo* -->
+    #   <p><em>bar</em></p>
+    #
+    # But the reference implementation outputs:
+    #
+    #   <p><!-- *foo* --><em>bar</em></p>
```
mitya57: These two versions of HTML render the same way. Maybe our version is a bit easier to read…

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+    #
+    # As the raw HTML is not alone on the line, the reference implementation
+    # considers it inline rather than block level. The behavior defined in
+    # the test below is from the CommonMark spec, which we don't follow.
+    def test_raw_comment_one_line_followed_by_text(self):
+        self.assertMarkdownRenders(
+            '<!-- *foo* -->*bar*',
+            '<!-- *foo* -->*bar*'
```
mitya57: …But why is `*bar*` not converted to `<em>bar</em>` here?

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -0,0 +1,783 @@
...
+    # TODO: Decide correct behavior. This matches current behavior and Commonmark.
+    # The reference implementation considers this inline not block level:
+    #
+    #   <p><![CDATA[ document.write(">"); ]]><em>bar</em></p>
+    #
+    # But most implementations do this instead:
+    #
+    #   <p>&lt;[CDATA[ document.write(“&gt;”); ]]&gt;<em>bar</em></p>
+    #
+    # Either makes sense, but the later seems more correct to me.
```
mitya57: Here I also vote for the latter variant.

tests/test_syntax/blocks/test_html_blocks.py
waylan: This is easy to fix. We simply allow 0-3 spaces instead of the current 0.

I agree, not producing invalid HTML is a good argument for the change. I asked because this would be a change in behavior for Python-Markdown. But if this change affects existing documents, the affect would be to make what was otherwise invalid HTML valid, so that's not a bad thing.

tests/test_syntax/blocks/test_html_blocks.py
waylan: Note that with the current behavior, that last line is not wrapped in a `p` tag. There is a closing tag, bu no opening tag. Conversely, the first lien has an opening tag, but no closing tag (although it gets auto-closed by the opening tag of the second line).

tests/test_syntax/blocks/test_html_blocks.py
waylan: Good point.

tests/test_syntax/blocks/test_html_blocks.py
waylan: The problem is that my new, incomplete implementation is stashing the entire string, including `*bar*`. Although, I could be wrong. I haven't dug in yet.

tests/test_syntax/blocks/test_html_blocks.py
waylan: I was going to say it is at the beginning, but that's not true as the opening `p` tag would be at the beginning. Good point. Thanks.

tests/test_syntax/blocks/test_html_blocks.py
waylan: So, with CDATA you suggest treating the tag as inline, but here you suggest treating the tag as block-level. In both cases, the reason for the question is the text after the tag on the same line. It seems to me all tags should exhibit the same behavior in that situation.

So my question is: if an otherwise block-level tag is not alone on a line(s), do we still treat it as block-level or inline? I have the same questiosn on multiple senarios in these tests, and would prefer the same behavior on all (unless there is a good argument to do otherwise).

tests/test_syntax/blocks/test_html_blocks.py
waylan: Your suggestion here seems to differ from all of the other situations were plain text follows and otherwise block-level tag on the same line. Any reason why?

mitya57: Whether CDATA is valid inside the document depends on whether it is HTML or XHTML. As our default output format is HTML (and XHTML is generally considered deprecated nowadays), I think it is better to escape occurrences of CDATA in our output.

tests/test_syntax/blocks/test_html_blocks.py
mitya57: I was not considering this as the same case as CDATA, I did not have in mind that these two cases are linked together.

> if an otherwise block-level tag is not alone on a line(s), do we still treat it as block-level or inline?

The implementations divide into two large groups (and one small group) here: [results for `<div>foo</div> bar`](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Cdiv%3Efoo%3C%2Fdiv%3E+bar).

If there is no clear leader, then maybe we should stick to what the reference implementation (Markdown.pl 1.0.2b8) does? I.e. treat the element as block-level but then process the remaining part of line separately.

tests/test_syntax/blocks/test_html_blocks.py
mitya57: My point was that the test should match the comment.

tests/test_syntax/blocks/test_html_blocks.py
mitya57: Ah, I did not notice that. Then maybe we should do the same thing as reference implementation, as valid output is better than invalid.

tests/test_syntax/blocks/test_html_blocks.py
facelessuser: I would personally treat all tags the same. We shouldn't bother considering if the tag is valid do to location. If just generally treat it as a preprocessor statement and not care whether it is valid or not.. keep things simple, at least that is my personal opinion.

tests/test_syntax/blocks/test_html_blocks.py
facelessuser: Most browsers just ignore CDATA in HTML5. I would personally just treat it as a preprocessor and move on. I personally don't like the idea of treating CDATA as text content.

tests/test_syntax/blocks/test_html_blocks.py
waylan: I agree with @facelessuser here. We shouldn't be special casing different content within the tag. Anything that matches `<!.*>` should all get the same treatment (whether that's a doctype, cdata or anything else). The question is: what treatment? I think it is clear that anything between the angle brackets gets ignored by Markdown (so no escaping), but do we treat the tag as inline or block level? If it is inline, then it gets wraped in `<p>` tags, if block-level it doesn't.

But this issue, like others I raised, is specific to the behavior when the tag is followed by additional text on the same line. Based on other tests, I fully intend for `<!anything>` to be treated as blocklevel. But what about `<!anything>more text` where "more text" is on the same line. That is the only question I am asking here. Should that specific edge case be inline or block level?

I ask about this specific edge case because the existing behavior is odd. It treats the tag as blocklevel, but also skips Markdown parsing on the "more text". Interestingly, that is what the [Commonmark spec][2] requires. But we are not a Commonmark parser. So, should we follow the reference implementation and treat the tag as inline or something else?

Personally I would have expected "more text" to be broken out into a separate Markdown paragraph. But according to [Babelmark][1], almost noone does that (marked being the sole exception).

[1]: http://johnmacfarlane.net/babelmark2/?normalize=1&text=%3C!DOCTYPE+html%3E*bar*
[2]: https://spec.commonmark.org/0.28/#example-144

tests/test_syntax/blocks/test_html_blocks.py
waylan: @mitya57 actually our default output format is XHTML. As a reminder, the reference implementation outputs XHTML and the rules specifically expect XHTML output. We will not be changing our default as long as that is the case.

Of course, in practice that makes no difference for HTML5 as both XHTML and HTML style tags are valid in HTML5. And as we only output fragments, not entire documents, the tag style is the only difference any output format can effect (`<br>` vs. `<br />`).

In any event, if we were going to escape or not escape CDATA based on output format, then that should be determined in the serializer, not in the HTML parser. Personally, I'm inclined to not bother with that. It is the document authors responsibility to ensure his raw HTML is valid, not ours. If including CDATA makes for an invalid document, then it shouldn't be included by the document author. It is not our place to make that determination.

mitya57: > actually our default output format is XHTML.

Ah, indeed.

> Personally, I'm inclined to not bother with that. It is the document authors responsibility to ensure his raw HTML is valid, not ours.

Sounds fine.

tests/test_syntax/blocks/test_html_blocks.py
mitya57: I'm fine with any behavior. If you have no preference then maybe keep the existing one?

waylan:
Sigh. I finally looked at the failing [processing instruction tests][0]. I never thought this would be what breaks the standard lib HTML parser:

```python
>>> class TestParser(HTMLParser):
...     def handle_pi(self, data):
...             print ('PI:', '"{}"'.format(data))
...     def handle_data(self, data):
...             print ('TEXT:', '"{}"'.format(data))
...
>>> parser = TestParser()
>>> parser.feed("<?php echo '>'; ?>")
PI: "php echo '"
TEXT: "'; ?>"
```

Note that the output should be:

```
PI: "php echo '>'; ?"
```

However, the angle bracket in the quote is mistaken for the closing bracket and everything after it is considered text outside of a tag. In fact, a look at the [source code][1] revels that it really is only looking for the next `>` irrespective of the context. I could file a bug, but this would need to get fixed and backported to all supported versions before it would be useful.

And then there is [this comment][2] in another section of the code (for finding the closing bracket of an end tag):

>     # consume and ignore other stuff between the name and the >
>     # Note: this is not 100% correct, since we might have things like
>     # </tag attr=">">, but looking for > after tha name should cover
>     # most of the cases and is much simpler

I would guess that Python-Markdown doesn't currently handle that case either, but is seems that these sorts of edge cases are not handled very well. In the processing instruction case, that is a very reasonable sort of thing to expect.

Although, I suppose it would be weird for PHP code to be embedded in a Markdown file which is not in a code block. The only purpose for that would be to use PHP an a template engine for generating Markdown. But in that case, the document would be run through PHP first, before being run through the Markdown parser and the PHP would no longer be present. That said, it *should* work correctly.

I suppose we could override the `parse_pi` method and fix it ourselves. Technically, the method is not private (it does not start with an underscore), although it is labeled "internal" in a comment. However, I'm not really interested in fixing underlying bugs in `HTMLParser`.

[0]: https://github.com/Python-Markdown/markdown/pull/803/files#diff-c7451b8f01beebb43e0258539a704255R585
[1]: https://github.com/python/cpython/blob/3.5/Lib/html/parser.py#L292
[2]: https://github.com/python/cpython/blob/3.5/Lib/html/parser.py#L407-L410

waylan:
All things considered, regardless of whether we use the HTML parsing implementation in this PR, I created a pretty good set of tests which we can use regardless of how HTML parsing is implemented.

waylan:
I did some research on processing instructions and it seems that, as [Wikipedia][1] summarizes:

> An SGML processing instruction is enclosed within `<?` and `>`.
> 
> An XML processing instruction is enclosed within `<?` and `?>`.

Wondering which applies to HTML, I checked the HTML5 spec, and under [Dependencies][2], it specifically states that the [Associating Style Sheets with XML documents specification][3] is where processing instructions are defined.

From that it seems clear that valid processing instructions in HTML5 should be enclosed within `<?` and `?>`. Of course, XHTML is XML, so the same applies there as well. That being the case, it shouldn't be difficult to monkeypatch HTMLParser to search for a closing `?>`, which will mostly resolve the issue. Any thoughts?

[1]: https://en.wikipedia.org/wiki/Processing_Instruction#Syntax
[2]: https://www.w3.org/TR/html5/single-page.html#conformance-requirements-dependencies
[3]: https://www.w3.org/TR/xml-stylesheet/#dt-xml-stylesheet

waylan:
And now we have a bigger problem with CDATA. The HTMLParser [only permits][1] CDATA within `<script>` and `<style>` tags. Interestingly, the [HTML5 spec][2] states:

> CDATA sections can only be used in foreign content (MathML or SVG).

So the fact that CDATA can only be used in specific situations is a valid concept. There are a few problems though:

1. The parser and the spec consider a different set of elements as acceptable. However, this could be rectified with a monkey patch.
2. More importantly, if CDATA is found outside of an acceptable tag, it is stripped completely from the document with no way to recover it.

I would be okay with DATA in an invalid location being escaped, but being completely removed is a little extreme. I don't know that there is a way around this one.

[1]: https://github.com/python/cpython/blob/3.5/Lib/html/parser.py#L85
[2]: https://www.w3.org/TR/html5/single-page.html#cdata-sections

waylan:
As of today, this has been rebased and updated to be PY3 only.

I also took a first attempt as handling `markdown=1` in 9fe2473, which is pretty simple. I simply did what I thought I remembered the old implementation doing: remove the `markdown` attribute, but otherwise pass the HTML block through (don't stash it) and let the inline HTML processor handle the start and end tags. But it doesn't quite work:

```python
>>> markdown.markdown('<div markdown="1">*text*</div>', extensions=['md_in_html'])
'<p><div><em>text</em></div></p>'
```

The `<div>` should not be wrapped in a `<p>` element. I'll have to dig into the old code to see how we handled that. Then decide if we want to follow the same approach or start over from scratch with a new approach.

waylan:
More progress with `markdown=1` in e4a8796.

Still some cleanup to do in the postprocessor. For example we need to better handle nested `<p>` tags. Also needs tests.

However, in the common case, this works well. All `markdown=1` blocks are normalized to the same insignificant whitespace by the preprocessor:

    start-tag-placeholder

    content

    end-tag-placeholder

Passing that to the blockprocessor results in the start and end tags each in a `<p>` tags by itself. The postprocessor then correctly removes the wrapping `<p>` tags. And the content gets wrapped in its own `<p>`.

Note that PHP Markdown Extra follows this behavior:

    <div markdown="1">

    *text*

    </div>

    <div markdown="1">
    *text*
    </div>

    <div markdown="1">*text*</div>

Each results in the same normalized output:

    <div>
    <p><em>text</em></p>
    </div>
    <div>
    <p><em>text</em></p>
    </div>
    <div>
    <p><em>text</em></p>
    </div>

waylan:
Well, this is discouraging. Finally looked as the extra long files in the tests which are failing. For example. with [tests/basic/markdown-syntax.txt][1] as the source (a copy of the rules from daringfireball), we only get five HTML blocks. The first four are correct. However, the fifth starts with with the `<h3>` on [line 122][2] and contains the entire rest of the document. 

For reference here is the truncated content of `md.htmlStash.rawHtmlBlocks` after parsing that file:

```python
[
    '<ul id="ProjectSubmenu">\n    <li><a href="/projects/markdown/" title="Markdown Project Page">Main</a></li>\n    <li><a href="/projects/markdown/basics" title="Markdown Basics">Basics</a></li>\n    <li><a class="selected" title="Markdown Syntax Documentation">Syntax</a></li>\n    <li><a href="/projects/markdown/license" title="Pricing and License Information">License</a></li>\n    <li><a href="/projects/markdown/dingus" title="Online Markdown Web Form">Dingus</a></li>\n</ul>',
    '<h2 id="overview">Overview</h2>',
    '<h3 id="philosophy">Philosophy</h3>',
    '<h3 id="html">Inline HTML</h3>',
    '<h3 id="autoescape">Automatic Escaping for Special Characters</h3>\n\nThe entire rest of document...'
]
```

The problem appears to be [this paragraph][3], which appears in the preceding section (titled `Inline HTML` ):

```text
The only restrictions are that block-level HTML elements -- e.g. `<div>`,
`<table>`, `<pre>`, `<p>`, etc. -- must be separated from surrounding
content by blank lines, and the start and end tags of the block should
not be indented with tabs or spaces. Markdown is smart enough not
to add extra (unwanted) `<p>` tags around HTML block-level tags.
```

That paragraph includes a number of opening tags which are never closed within code spans. The HTML parser doesn't recognize that it should ignore them, so once the next raw block is opened (with the `<h3>`) it fails to end the raw block at the closing `</h3>` because it recognizes that some additional tags are still unclosed. Apparently, we need to track which tag opens the raw block, and then use the matching closing tag to close the raw block.

When I say that the "HTML parser doesn't recognize that it should ignore them" I mean that is doesn't ignore them completely. It does recognize that they don't start a raw block as they are not at the start of a line. However, it still parses them and adds them the the HTMLParser's internal stack. While the parser recognizes the `<h3>` is closed, as the stack is not empty, it does not close the raw block. So we need to close the raw block based on the fact that the opening tag of the block has been closed, not that all tags have been closed.

This is easy to miss in simple tests if they use `<p>` tags as `<p>` tags autoclose at the next block level tag. But that doesn't happen with `<div>`,`<table>`,  or `<pre>`.

[1]: https://github.com/Python-Markdown/markdown/blob/be7ba7bfa17e7cb0c795e101662d1e490334ec84/tests/basic/markdown-syntax.txt
[2]: https://github.com/Python-Markdown/markdown/blob/be7ba7bfa17e7cb0c795e101662d1e490334ec84/tests/basic/markdown-syntax.txt#L122
[3]: https://github.com/Python-Markdown/markdown/blob/be7ba7bfa17e7cb0c795e101662d1e490334ec84/tests/basic/markdown-syntax.txt#L112-L116

facelessuser:
Hmm, I may be able to take a look at this sometime soon to see if I can provide any help. I'm currently not familiar enough to understand how the new methodology works, but if it is easier to understand than the original implementation, I can hopefully help provide a fix.

facelessuser:
At the very least, HTML tags should be recognized only at the start of a line. I assume we are missing logic for this.

waylan:
Just added a [simple test][1] which demonstrates the problem.

```python
def test_raw_block_preceded_by_markdown_code_span_with_unclosed_block_tag(self):
    self.assertMarkdownRenders(
        self.dedent(
            """
            A paragraph with a block-level code span: `<div>`.

            <p>*not markdown*</p>

            This is *markdown*
            """
        ),
        self.dedent(
            """
            <p>A paragraph with a block-level code span: <code>&lt;div&gt;</code>.</p>
            <p>*not markdown*</p>
            <p>This is <em>markdown</em></p>
            """
        )
    )
```

The actual output is:

```text
<p>A paragraph with a block-level code span: <code>&lt;div&gt;</code>.</p>
<p>*not markdown*</p>

This is *markdown*
```

Notice that it fails to parse the last line as Markdown.

[1]: https://github.com/Python-Markdown/markdown/pull/803/commits/6b4b3519b8d9d415e498b95f4376ca5905deb265#diff-c7451b8f01beebb43e0258539a704255R250

facelessuser:
That's interesting. It seems to get locked into ignoring Markdown. Thanks for adding the test. It will make it more clear what I'm looking to solve when I allocate some time to do so.

waylan:
Fixed the issue in c0194f3. Turns out it was pretty simple once I worked out the problem. It was one of the those things where explaining the problem helped me work it out.

There are still other issues. but those are mostly white space related.

facelessuser:
> It was one of the those things where explaining the problem helped me work it out.

Yup, been there.

waylan:
Looks like a ran into a weird bug in Python's  `html.parser.HTMLParser`. If I feed it:

```text
AT&T
<div>anything</div>
```

It breaks that up into two 'data' "tokens" (yes, I know the `HTMLParser` doesn't actually output a list of tuples, but it is an easy way to represent the output here):

```
[
    ('data', 'AT'),
    ('data', '&T\n<div>anything</div>')
]
```

The problem is that everything after the standalone `&` is consumed with it and not parsed as HTML. I haven't dug into the parser code yet, but AFAICT, the bug is in the parser itself. Note that a proper entity works fine (for example `&amp;` works fine).

I would expect the parser to break the above up into the following tokens:

```
[
    ('data', 'AT&T\n'),
    ('starttag', '<div>'),
    ('data', 'anything'),
    ('endtag', '</div>')
]
```

Or at least something similar (perhaps giving allowance for some weirdness with `AT&T`). However, I would not expect it to cause the entire rest of the document to be seen as plain text (not HTML).

I first encountered the issue while debugging an issue with the `tests/basic/markdown-syntax.txt` test. Specifically, somewhere within the following section all parsing of HTML stopped so that nothing after that point was recognized as block level raw HTML. At first it seemed to be a random location. I assumed it might be related to the standalone `<`, but it handled that just fine. Eventually I narrowed it down to the above simple case.

https://github.com/Python-Markdown/markdown/blob/be7ba7bfa17e7cb0c795e101662d1e490334ec84/tests/basic/markdown-syntax.txt#L122-L181

waylan:
This is weird, the lack of white space around the `&` makes a difference. `AT & T` and `AT& T` do not exhibit the bug, but `AT &T` and `AT&T` do. For that matter so does `&T`. So long as the `&` is immediately followed by `[a-zA-Z]`. I tried punctuation, numbers, and new lines and nothing else seems to trip it up.

As I understand it, the parser is supposed to recognize HTML entities which to not include the semicolon (`;`), but entities do not include non-ASCII letters, so the parser should at a minimum stop at the first non a-z character. Additionally, the parser is not returning it as an entity, but as `data` (plain text content), so it is properly recognizing that it is not an entity. Without looking at the code, my guess is that a regex should be non-greedy where it is greedy. If so, that should be a relatively easy fix, but that requires us to monkeypatch the parser until such time as the issue is addressed upstream. And standard lib issues generally don't get fixed as quickly as third party libs.

waylan:
Ironically, this problem was caught because John Gruber didn't follow his own advice when writing the rules (see [original source][1]). Note this sentence:

> ```
>  If you want to write about 'AT&T', you need to write '`AT&amp;T`'.
> ```

He should have either written `'AT&amp;T'` (using an entity outside of a code span) or `` '`AT&T`' `` (no entity in a code span) for the first instance. Of course, the parser would still trip up over the second of those options, so we need to address it anyway.

[1]: https://daringfireball.net/projects/markdown/syntax.text

facelessuser:
Ugh, that is kind of an annoying issue.

waylan:
Sigh. I forgot about my change in 49c187dfc which is messing up the handling of `&`. The good news is that it's not a bug in the upstream library.

markdown/htmlparser.py
```diff
@@ -26,6 +26,10 @@
 parser.piclose = re.compile(r'\?>')
 # Monkeypatch HTMLParser to only recognize entity references with a closing semicolon.
 parser.entityref = re.compile(r'&([a-zA-Z][-.a-zA-Z0-9]*);')
+# Monkeypatch HTMLParser to no longer support partial entities. We are always feeding a complete block,
```
facelessuser, line 38: Is monkey patching the only way to address this? I'm worried that there may be side effects if using markdown with other libraries that rely on `html.parser`. I'm curious if there is some cleaner way to do this without altering the builtin library..

facelessuser: So, I was looking into this. What if we modify a copy of the module instead of the original?

```py
>>> import importlib
>>> import sys
>>> spec = importlib.util.find_spec('html.parser')
>>> monkey = importlib.util.module_from_spec(spec)
>>> spec.loader.exec_module(monkey)
>>> sys.modules['monkey'] = monkey
>>> monkey.entityref = 'whatever'
>>> from html import parser
>>> parser.entityref
re.compile('&([a-zA-Z][-.a-zA-Z0-9]*)[^a-zA-Z0-9]')
>>> monkey.entityref
'whatever'
```

This way we could use our monkey patched version, but not modify the original. What do you think?

waylan: > I'm worried that there may be side effects if using markdown with other libraries that rely on `html.parser`.

That's a valid concern. I had been avoiding the issue because I didn't want to maintain a copy of the lib. However...

> we could use our monkey patched version, but not modify the original.

That works around my concern quite well. I like it.

waylan: Change has been made in efa36c8.

waylan:
Looks like I'm done with the default behavior. The legacy tests are all passing without modification (I had modified a couple in an earlier commit, but reverted them in f02b427). 

Just need to finish up the last few issues with the `markdown=1` logic.

markdown/htmlparser.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.inraw:
+            self._cache.append(text)
+            if tag in self.stack:
+                # Remove tag from stack
+                while self.stack:
+                    if self.stack.pop() == tag:
+                        break
+            if len(self.stack) == 0:
+                # End of raw block.
+                if self.rawdata[self.line_offset + self.offset + len(text):].startswith('\n\n'):
```
waylan: I think we need to discuss this. All this does is get the existing legacy tests to pass. However, it leaves a lot of unaddressed edge cases. To summarize, without this, All blank lines before and after a raw HTML block would be removed in the final output. And that is all fine as the extra white space is insignificant and does not affect a browser's rendering.

However, the reference implementation retains the blank lines, as did Python-Markdown in the past. Therefore, our old tests all include the blank lines. This change was the last to get all the old tests to pass.

That said, the behavior is inconsistent in practice. The old tests are simply incomplete in that they only include the common cases, but not the edge cases. For example, there could be spaces before or between the two newlines immediately following the closing tag of the HTML block. In those cases, the blank line would not be preserved.

I see three ways forward:

1. Leave things as is. The inconsistency is okay as it is all insignificant white space anyway.
2. Revert the last change and update the tests to not include the blank lines in the expected output.
3. Fix this properly by only closing out a raw block at the first newline following the closing tag. This would require the 'end of block' code to go in the `handle_data` method rather than the `handle_endtag` method. Then, where appropriate, the data would be split with everything up to and including the first newline being added to the raw block, with any remaining data being outside of the raw block.

Note that option 3 would also change the behavior of a number of tests which are not related exclusively to white space. For example, consider these tests:

https://github.com/Python-Markdown/markdown/blob/efa36c8a03d29783d930d595cf341f34a080aa7d/tests/test_syntax/blocks/test_html_blocks.py#L703-L722

https://github.com/Python-Markdown/markdown/blob/efa36c8a03d29783d930d595cf341f34a080aa7d/tests/test_syntax/blocks/test_html_blocks.py#L983-L994

https://github.com/Python-Markdown/markdown/blob/efa36c8a03d29783d930d595cf341f34a080aa7d/tests/test_syntax/blocks/test_html_blocks.py#L1064-L1075

https://github.com/Python-Markdown/markdown/blob/efa36c8a03d29783d930d595cf341f34a080aa7d/tests/test_syntax/blocks/test_html_blocks.py#L1101-L1111

Implementing option 3 would bring all of those inline with the reference implementation.

The weird one is this:

https://github.com/Python-Markdown/markdown/blob/efa36c8a03d29783d930d595cf341f34a080aa7d/tests/test_syntax/blocks/test_html_blocks.py#L135-L146

Option 3 would restore Python-Markdown's behavior, but would not be inline with the reference implementation. Although, I think it would be reasonable to conclude that this is a bug in the reference implementation.

Personally, I'm fine with any of the options (preferring option 1 as it requires no work). I like option 2 because it is what I would have expected had I not checked the behavior of the reference implementation. Of course, option 3 actually most closely matches the reference implementation.

I will point out that I'm not all that concerned about changes in behavior here as they all involve content after the closing tag on the same line. I contend that users should not expect that to be allowed and any reliance on that within their documents is a hack.

Any feedback?

waylan: I took a stab at option 3 and things started to get weird. So I took a less extreme approach. Mostly things stay as-is, except for two cases. The last last test case mentioned in previous comment is now:

```python
    def test_multiple_raw_single_line(self):
        self.assertMarkdownRenders(
            '<p>*foo*</p><div>*bar*</div>',
            self.dedent(
                """
                <p>*foo*</p>
                <div>*bar*</div>
                """
            )
        )
```

Also, the first test case in previous comment is now:

```python
    def test_raw_comment_one_line_followed_by_html(self):
        self.assertMarkdownRenders(
            '<!-- *foo* --><p>*bar*</p>',
            self.dedent(
                """
                <!-- *foo* -->
                <p>*bar*</p>
                """
            )
        )
```

The other three maintain their behavior, which I think makes more sense anyway. It is certainly what users would expect if they formatted their documents that way.

waylan: Basically, I added a special case for a raw block which does not start at the beginning of a line, but immediately follows the end of the closing tag of a raw block on the same line as that closing tag.

waylan:
I would like some feedback on how to proceed here.

As mentioned in my previous comment, the default behavior is working good now. However, the `md_in_html` extension needs to be completed (which adds support for `markdown=1`). There is a simple working version in this PR as described in [this comment](https://github.com/Python-Markdown/markdown/pull/803#issuecomment-653160670). However, the previous implementation also supported `markdown="block"` and `markdown="span"` with `markdown=1` defaulting to `span` on some tags and `block` on others (see [docs](https://github.com/Python-Markdown/markdown/blob/370e17b3923d7d8f4709652bc24b695271a0210d/docs/extensions/md_in_html.md#usage)). In other words, we had a (near) clone of Kramdown's HTML handling. To maintain backward compatibility the simple version we have so far won't cut it. We need a more sophisticated solution.

As an aside, the existing simple solution might make for a basis to implement Commonmark's HTML handling with a few simple modification. Or it could be used for the basis of an extension which copied PHP Markdown Extra's simple behavior. But I digress.

I have made two separate attempts at a solution for this and each works about the same. However, the implementations are very different. Each is summarized below.

# Separate Parsers

In [waylan:mdinhtml](https://github.com/waylan/markdown/blob/mdinhtml/markdown/extensions/md_in_html.py) I (mostly) left the HTML parser unmodified. The only change to the default parser was to include some extra meta-data in the HtmlStash about each stashed block (the parent tag and any attributes on that tag).

Then I created a blockprocessor which matches a block (paragraph) which only contains a placeholder. Using the new meta-data, if that placeholder contains an HTML block with Markdown content that should be parsed, the content of the stash is passed through a second HTML parser which converts to the HTML to an ElementTree object. Then the content of the ElementTree object is recursively parsed as Markdown taking into account the tag and `markdown` attribute. Finally, the element is inserted in to the document as an ElementTree object.

The pros of this implementation are that the default parser is completely undisturbed. We are ensured consistent parsing of raw HTML regardless of any extensions. The only change in behavior is that the content of some tags can now be parsed as Markdown.

The cons are that the HTML is parsed twice, which nearly doubles the parse time for the entire document in my limited testing.

# Combined Parser

In [waylan:mdinhtml2](https://github.com/waylan/markdown/blob/mdinhtml2/markdown/extensions/md_in_html.py) I subclass the HTML parser in the extension. The sub-classed parser will check the attributes of a tag and change its behavior accordingly. If a normal raw HTML block, it stores the block in the HtmlStash as before. However, it id identifies a `markdown` attribute, then it instead builds an ElementTree object and stores that object in the HtmlStash.

This implementation also includes a blockprocessor which matches a paragraph containing only a placeholder. If that placeholder points to an ElementTree object in the stash, then it parses the content of the element as Markdown according to the tag and attribute and inserts that element into the document.

The pros of this implementation is that we only parse the HTML once. And that makes for a faster parse time.

The cons are that I am finding more edge cases which now need to be handled twice in the HTML parser. Once for the default behavior, and a second time for the new behavior. It becomes much more difficult to maintain consistent parsing behavior between the extension and the default.

Regardless, either solution is drastically simpler than the preexisting solution. I don't expect I'll dread working on it in the future. So that's something. Any feedback on the two solutions is welcome. I should note that the combined parser is less developed as I have spent less time on it. I simply got far enough to prove that the concept works.

Based on my explanation above, it seems clear to me that we should go with the combined parser solution. However, I really like the separation of behaviors (and maintainability)  in the separate parsers solution.

One final note I should make: Both of these implementations are parsing the raw HTML and converting it to ElementTree objects. In the past we have always been clear that raw HTML is passed through unaltered. That is no longer true for raw HTML with `markdown` attributes (it remains true for regular raw HTML). Personally, I am okay with that. But I thought it should get pointed out.

facelessuser:
So, questions, the separate parser states:

> The only change in behavior is that the content of some tags can not be parsed as Markdown.

What does this mean exactly? What content cannot be parsed?

waylan:
>So, questions, the separate parser states: ...

@facelessuser sorry, that is supposed to say “now” instead of “not”.

facelessuser:
I'm not sure I fully understand this statement:

> The cons are that I am finding more edge cases which now need to be handled twice in the HTML parser. Once for the default behavior, and a second time for the new behavior. It becomes much more difficult to maintain consistent parsing behavior between the extension and the default.

Is it possible for you to elaborate? Is this stating that with `md_in_html`, the output of the HTML is different than when it is not enabled? In general, are they insignificant differences?

waylan:
> Is it possible for you to elaborate? Is this stating that with `md_in_html`, the output of the HTML is different than when it is not enabled? In general, are they insignificant differences?

In some edge cases yes. There are two contributing factors:

1. The HTML it converted to an ElementTree object. This may reformat the HTML in versions ways. For example, raw unclosed `<p>` tags (which are valid) would get closing tags added to them. That doesn't happen when we simply store the text of raw HTML in a stash. Another example is that insignificant whitespace may be altered when the `PrettifyTreeprocessor` is run.

2. When parsing raw HTML with a `markdown` attribute, the code take a different path and unless all of the same logic is implemented in each path, there will be differences. 

I should note that the first factor applies to both implementations. However, I am not concerned about that. It is the second factor which I was discussing above. 

As an example, in the 'shared parser' solution, the two code paths (for regular raw HTML and raw HTML with `markdown` attribute) handle tails of raw HTML blocks differently (for example: `<div>foo</div> *this part*`) when the tail is on the same line as the closing tag. That logic can be added to the alternate code path, but it is an additional maintenance burden. Each time some new edge case is discovered, the fix will need to be applied twice.

By contrast, in the "separate parsers" solution, this is a non-issue as the HTML is always parsed by the same code path the first time through. That first parse breaks the raw HTML up into blocks as Markdown sees them. Therefore we don't need to be concerned about things like tails when handling `markdown` attributes.

facelessuser:
Ah, that makes sense. it's a shame that more of that logic can't be shared. I haven't looked close enough at this implementation to see if there is anyway besides having redundant handling. It'd be nice if there was some way to unify that logic in on place and reuse it though.

I know that generally, people will prefer the speed. In general, the maintenance burden is probably already much lower than it was compared to the old way. I usually don't mind jumping in and fixing bugs, but I dreaded HTML parsing bugs.

There's also a part of me that likes the idea of easier maintenance by parsing it twice. I have no concept as I am writing this what that time burden feels like right now. I think I'll have to take a look at the new code and get a feel for it. Hopefully this weekend I can evaluate it more.

waylan:
Here are some parse times:

```
$ git checkout master
$ python time_it.py
0.003215398999999999

$ git checkout mdinhtml
$ python time_it.py
0.0045843939999999994

$ git checkout mdinhtml2
$ python time_it.py
0.0026239879999999994
```

As you can see, the `mdinhtml` branch (combined parser) is approximately half the time of the `mdinhtml2` branch (separate parsers). What is interesting to me is that the `master` branch (pre-existing implementation) falls right between the two. So the combined parser solution actually gives us a notable performance boost while the separate parsers solution takes us a step backwards by approximately the same amount.

The choice for users seems pretty clear. And yes, the maintenance burden for either implementation is much less than the pre-existing implementation. The fact is, the code in the `mdinhtml2` branch was thrown together in a couple of hours (with multiple interruptions) as a proof-of-concept. I suspect I could refactor the code in the base class, so that it would be easier to override only the relevant sections in the extension's subclass.

Truth be told, the fact that I was able to create a mostly working implementation with so little effort is a testament to how much we have reduced the maintenance burden. While it is true that I borrowed some code from the slower implementation, overall I suspect that it will still be such an improvement that we won't mind the few inconveniences. Besides, maintaining two HTML parsers could bring its own burden as they both would need to handle the same source in the same manner. Perhaps in the end, there isn't really any maintenance savings from one to the other.

For reference, here is the code I used for the above tests (content of `time_it.py`):

```python
import markdown
import timeit

with open('tests/extensions/extra/raw-html.txt') as f:
    src = f.read()

num, dur = timeit.Timer(lambda: markdown.markdown(src, extensions=['md_in_html'])).autorange()
print(dur / num)
```

facelessuser:
Yeah, seems I'm leaning towards combined right now, based on what I'm seeing.

facelessuser:
The good news is if you did go with the separate parser, it isn't double what's on master, just double the improved combined parser. And to be honest, it's not like we don't have other places we have to deal with things in a redundant manner. I just had to fix essentially the same bug in `details` and `admontions`, albeit the fix was a little different, but the same kind of thing.

facelessuser:
But generally, the combined seems like the better choice.

waylan:
After a month hiatus, I have resumed working on this. I currently have all but one one issue resolved in the `waylan/markdown:mdinhtml2` branch (see a [comparison to master][1] for all changes). That branch is based off of this PR with additional changes to the `md_in_html` extension (see a [comparison to this PR][2] for only the `md_in_html` changes). While I created new tests using the newer test framework, I did not immediately delete the old legacy tests. I worked through them and only deleted them after ensuring they all passed (with a few exceptions relating to insignificant white space). 

I currently have one outstanding issue with failing legacy tests. Specifically, the tests added in #585 are all failing as references (link references, abbreviation references, and footnote references) are not being extracted from raw HTML. Prior to #585 they were not either and the bug report related to that fix (#584) was not specifically requesting that we support that. However, due to the implementation of the extension at that time, including references wrapped in raw HTML would cause errors. The solution we chose happens to add support. However, I am inclined to remove support going forward. 

There is no reason to define a reference within raw HTML as the reference is removed during parsing. The only reason a document author might do so is to keep the reference definition closer to the item is is a reference for. However, in standard Markdown, where no Markdown processing happens inside raw HTML, references would be ignored inside raw HTML. The issue only arises when using `markdown="1"` with the extension. However, `markdown="1"` enables block parsing and references are not "blocks" so I would argue that they should not be parsed there.

From an implementation perspective, references are not being parsed because, in the new implementation the extension's preprocessor extracts all raw HTML (and its content) from the document and converts the raw HTML to ETree Elements held in the HTMLStash. Therefore, any additional preprocessors never see the content of those raw HTML blocks. Later, a block processor identifies the placeholders and replaces them with the ETree elements, parsing the content as appropriate. To enable support for references within raw HTML,I see three possible solutions:

1. Run the reference preprocessors before the raw HTML preprocessor. However, this would only apply when the `md_in_html` extension is enabled. The current order of operations would apply without the extension. This also wouldn't differentiate between `markdown="span"` and `markdown="block"` so probably a no-go.
2. Run the preprocessors on the contents of each raw block. This seems messy from an API perspective. How would we determine which preprocessors to run? How would we deal with unknown third-party extensions? Some might be appropriate to include while others would not.
3. We could refactor all of the reference finding preprocessors to be blockprocessors, but that would be a separate PR from this one. Not something I really want to spend the time on right now.

Any thoughts on this?

[1]: https://github.com/Python-Markdown/markdown/compare/master...waylan:mdinhtml2
[2]: https://github.com/waylan/markdown/compare/html...waylan:mdinhtml2

waylan:
I just checked [Babelmark][1] and I see that all other implementations which support parsing block level Markdown within raw HTML include support for link references being defined within the raw HTML. In addition to all Commonmark implementations (using the blank line technique), PHP Markdown Extra, MultiMarkdown, Kramdown, and markdown.pl 1.0.2b8 all exhibit the same behavior with `markdown="1"`. I suppose that means we should too.

[1]: https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Cdiv+markdown%3D%221%22%3E%0A%0A%5Blink%5D%3A+http%3A%2F%2Fexample.com%0A%0A%3Cdiv+markdown%3D%221%22%3E%0A%0A%5Blink%5D%5Blink%5D%0A%0A%3C%2Fdiv%3E%0A%0A%3C%2Fdiv%3E

waylan:
> I just checked [Babelmark](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Cdiv+markdown%3D%221%22%3E%0A%0A%5Blink%5D%3A+http%3A%2F%2Fexample.com%0A%0A%3Cdiv+markdown%3D%221%22%3E%0A%0A%5Blink%5D%5Blink%5D%0A%0A%3C%2Fdiv%3E%0A%0A%3C%2Fdiv%3E) and I see that all other implementations which support parsing block level Markdown within raw HTML include support for link references being defined within the raw HTML. In addition to all Commonmark implementations (using the blank line technique), PHP Markdown Extra, MultiMarkdown, Kramdown, and markdown.pl 1.0.2b8 all exhibit the same behavior with `markdown="1"`. I suppose that means we should too.

It was easy enough to move reference link and abbr parsing to the block parser, but footnotes are a whole other animal as they can contain a number of nested block-level content. Therefore, before proceeding, I checked [other implementations][1]. It appears that Kramdown and MultiMarkdown 6.3 do not support footnote references being defined within raw HTML blocks. What is most interesting is that MultiMarkdown 5.1 did. I'm not sure if that is a bug which was introduced in version 6 or an intentional change. Although, looking at 5.1's output, it is clear there are some bugs there. For [comparison][2], MultiMarkdown 6.3 works just fine outside of raw HTML blocks, while 5.1 retains the same bugs. In other words, 5.1's bugs are not specific to the reference being wrapped in raw HTML. Instead, it appears that in 5.1 footnotes were not completely working yet. However, in version 6 when footnotes were fully supported, they did not include support for the references being defined within raw HTML.

That said, PHP Markdown Extra and Pandoc both clearly support defining footnote references within raw HTML. Therefore we have 2 for and 2 against including support. And when we consider that Python-Markdown previously supported then as well, it seems like we should probably continue to. I'm just not sure how that will work with the new implementation. Suggestions are welcome.

[1]: https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3Cdiv+markdown%3D%221%22%3E%0Afoo%5B%5E1%5D%0A%3Cdiv+markdown%3D%221%22%3E%0A%5B%5E1%5D%3A+bar%0A%3C%2Fdiv%3E%0A%3C%2Fdiv%3E
[2]: https://johnmacfarlane.net/babelmark2/?normalize=1&text=foo%5B%5E1%5D%0A%0A%5B%5E1%5D%3A+bar

waylan:
I worked out the footnotes issue. Turned out to not be as difficult as I anticipated. 

I have since finished resolving all tests and merged by work back into this PR. And then I discovered that various features of [`xml.etree.ElementTree.TreeBuilder`][1] were added in Python 3.8 which I am relying on. Thus, the failing tests in `<py38`. That's what I get for developing in the latest supported version rather than the earliest. Unfortunately, the only way to address this is to vendor the PY38 version of `TreeBuilder`. Otherwise, any HTML comments or Processing Instructions within `markdown="1"` blocks would be lost.

[1]: https://docs.python.org/3/library/xml.etree.elementtree.html#treebuilder-objects

waylan:
> Unfortunately, the only way to address this is to vendor the PY38 version of `TreeBuilder`. Otherwise, any HTML comments or Processing Instructions within `markdown="1"` blocks would be lost.

Actually we only ever convert block level elements to etree elements if they have `markdown="1"` defined. That will never be the case for comments and PIs so we can remove the code for them from `TreeBuilder`.

waylan:
As a sanity check, could others verify that the following defaults make sense when `markdown="1"`? I based the behavior on the preexisting behavior and what Kramdown (which is is the closest to our behavior) [does][1] by default for each element. Of course, the default for each element can be overridden with either  `markdown="block"` or `markdown="span"` so long as the tag is an element listed in `span_tags` or `block_tags`, respectively.

What I'm asking here is if the elements are in the correct lists? And are there any elements which I missed, or that should not be included?

[1]: https://github.com/gettalong/kramdown/blob/master/doc/syntax.page#L994

https://github.com/Python-Markdown/markdown/blob/42299a84b6545d18523a8d43e93799606e92eab4/markdown/extensions/md_in_html.py#L25-L36

waylan:
Documentation of the `md_in_html` extension has been completely rewritten and the documented behavior of all examples has been verified. For anyone reviewing this who wasn't familiar with the current behavior, the updated documentation might be a good place to start. 

The only failing tests now are related to coverage issues. I need to add a few more tests to [cover some edge cases][1] and this should be ready to go. 😄 

[1]: https://github.com/Python-Markdown/markdown/pull/803/checks?check_run_id=1093028489

facelessuser:
Unfortunately, I haven't really had the time to test this. A lot of stuff going on, a death in the family, and other things. With all of that said though, still very excited for this 🙂.

waylan:
I added the following to the `md_in_html` tests in 2d8ce54:

https://github.com/Python-Markdown/markdown/blob/2d8ce544766abf532896d6c9f7700ac4feb4d70d/tests/test_syntax/extensions/test_md_in_html.py#L24-L30

It works well to ensure that the extension doesn't break the default behavior. The problem is that `unittest discover` sees the `TestHTMLBlocks` class which is imported and runs it as well.  However, it is also run from its original location. Therefore, all of the testcases get run 3 times, once in `TestDefaultwMdInHTML` (with the extension enabled) and twice in `TestHTMLBlocks` (without the extension enabled). Any idea how to avoid the extra, unnecessary run of those tests?

While running the tests an extra time makes the test run slightly longer, that's only a minor inconvenience. The real problem is that it is confusing that the same test failure gets reported three times, two of them being the exact same test. It doesn't even report the two failures as being from different locations. It reports both as being in `tests.test_syntax.blocks.test_html_blocks.TestHTMLBlocks`.

waylan:
We now have 100% patch coverage with all tests passing. 💯 🎉 

We just need to address the following items and this should be ready to go. 🚀 

- [X] 1. <del>The issue raised in my [previous comment][1].</del>
- [X] 2. <del>Release notes.</del>
- [ ] 3. A final review (preferably by someone other than myself).

[1]: https://github.com/Python-Markdown/markdown/pull/803#issuecomment-692719006

mitya57:
> Any idea how to avoid the extra, unnecessary run of those tests?

One option would be adding a `load_tests` function to the module, like described here: <https://docs.python.org/3/library/unittest.html#load-tests-protocol>.

Maybe it will also work (not tested) if you split `TestHTMLBlocks` into two classes: a base class that will have the actual tests and a subclass that will inherit from TestCase, like this:
```python
class HTMLBlocks:
    # Test functions go here.

class TestHTMLBlocks(HTMLBlocks, TestCase):
    # No functions here, just deriving from HTMLBlocks and TestCase.
```

and in the other file:
```python
class TestDefaultwMdInHTML(HTMLBlocks, TestCase):
    default_kwargs = {'extensions': ['md_in_html']}
```

waylan:
I just confirmed that this PR fixes all of the following open issues: #595, #780, #830, and #1012. It appears that we have addressed all open issues related to HTML parsing. 👍 🎉

waylan:
> One option would be adding a `load_tests` function to the module

@mitya57 that did the trick (1eb9fd3 ). Thanks.

facelessuser:
This is fantastic news.

waylan:
@facelessuser @mitya57 I have done a final read-through of this and am ready to merge. However, if either of you expect to have time to take a look soon, I'll happily wait. Let me know and I'll proceed accordingly.

facelessuser:
I'd like an opportunity to try it out this weekend, but if I don't get to it by then, if move ahead without me.

mitya57:
Sorry, I didn't have time to review this properly. But from a quick look it is good. Thanks for your work!

mbarkhau:
x-ref: https://gitlab.com/mbarkhau/markdown-katex/-/issues/14

Just in case anybody else runs into this. Since 3.3 the new intermediate processing of html may change attributes with single quotes to double quotes. This broke some extensions that used naive string replacement rather than html parsing.

Regardless, thanks for the continued work.

Issue #792, opened 2019-02-27
Bold/Italic bug


Dave-ts:
I think I'm running up against another bold/italics bug. I did some quick searches and it looks like the other issues were considered resolved, sorry if I'm re-reporting on something already fixed that hasn't made it upstream yet. 

Installed from pip current Python-Markdown version 3.0.1

The raw markdown line that breaks is:
```
This is text **bold *italic bold*** with more text
```

The output I'm getting is as follows:

```
<p>This is text <strong>bold *italic bold</strong>* with more text</p>
```

However, the following format does seem to work correctly.

```
This is text ***bold italic** italic* more text
```
The output is

```
<p>This is text <em><strong>bold italic</strong> italic</em> more text</p>
```

waylan:
I can confirm I am seeing the same behavior in 3.0.1. However, this behavior does not exist in the latest code here on GitHub and will be available in the next release.

```python
>>> import markdown
>>> markdown.version
'3.1.dev0'
>>> markdown.markdown('This is text ***bold italic** italic* more text')
'<p>This is text <em><strong>bold italic</strong> italic</em> more text</p>'
```

facelessuser:
@waylan, I don't think you tested is case properly. It wasn't `This is text ***bold italic** italic* more text` that failed, but `This is text **bold *italic bold*** with more text`.

facelessuser:
But I will explain a little bit why this happens.

In the pattern `This is text **bold *italic bold*** with more text`, the parser Sees `**bold *italic bold**` and makes it `<strong>bold *italic bold</strong>`. This leaves two `*` in separate tags, and they cannot be resolved or you'd end up with some weird broken HTML.

The 3rd party  [`pmdownx.betterem`](https://facelessuser.github.io/pymdown-extensions/extensions/betterem/) can handle this, but only when `smart_enable` is set to handle the `star` syntax (it only does "smart" for underscore by default).

```py
>>> markdown.Markdown(extensions=["pymdownx.betterem"], extension_configs={"pymdownx.betterem": {"smart_enable": "all"}}).convert("This is text **bold *italic bold*** with more text")
'<p>This is text <strong>bold <em>italic bold</em></strong> with more text</p>'
```

Without smart enabled for `*`, it does what Python Markdown does by default:

```py
>>> markdown.Markdown(extensions=["pymdownx.betterem"]).convert("This is text **bold *italic bold*** with more text")
[...]

PR #805, opened 2019-03-13
Refactor em strong to consolidate code and fix issue #792

facelessuser:
This refactors em strong logic into two processors: one for asterisks and one for underscores.  All logic for the respective emphasis convention is consolidated, and a new pattern is added to handle `**strong*em***` and `__strong_em___` respectively.

### DIFF
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -69,7 +69,6 @@ try:  # pragma: no cover
     from html import entities
 except ImportError:  # pragma: no cover
     import htmlentitydefs as entities
-from collections import OrderedDict
 
 
 def build_inlinepatterns(md, **kwargs):
@@ -576,7 +575,7 @@ class AsteriskProcessor(InlineProcessor):
         return el, start, end
 
 
-class UnderscoreProcessor(InlineProcessor):
+class UnderscoreProcessor(AsteriskProcessor):
     """Emphasis processor for handling strong and em matches inside underscores."""
 
     PATTERNS = [
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -64,6 +64,7 @@ So, we apply the expressions in the following order:
 from __future__ import absolute_import
 from __future__ import unicode_literals
 from . import util
+from collections import namedtuple
 import re
 try:  # pragma: no cover
     from html import entities
@@ -177,6 +178,10 @@ def dequote(string):
         return string[1:-1]
     else:
         return string
+
+
+class EmStrongItem(namedtuple('EmStrongItem', ['pattern', 'builder', 'tags'])):
+    """Emphasis/strong pattern item."""
 
 
 """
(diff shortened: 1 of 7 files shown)


facelessuser:
Tests are still needed for the new, fixed case.

facelessuser:
Current tests are now passing, but will have to address coverage as well.

markdown/inlinepatterns.py
```diff
@@ -431,6 +439,154 @@ def get_stash(m):
...
+        el = None
+        start = None
+        end = None
+
+        for index, pattern in enumerate(self.PATTERNS):
+            m1 = pattern[0].match(data, m.start(0))
+            if m1:
+                start = m1.start(0)
+                end = m1.end(0)
+                el = self.build_element(m1, pattern[1], pattern[2], index)
```
waylan: It took me a minute to realize what `pattern[0]`, `pattern[1]`, and `pattern[2]` were referring to. I wonder if a named tuple would make sense here. That way we would have `pattern.re` (or some better name; `pattern.pattern` seems weird), `pattern.builder`, and `pattern.tags`. That would also clear up what all those strings mean where `PATTERNS` is defined.

In all other respects this looks good, pending complete test coverage.

I hesitated for a minute as we don't get the granularity to override specific parts with extensions with this implementation. But then I remembered that the various parts were all reliant on each other anyway. Altering an earlier pattern often required also altering a later pattern. This actually makes the fact that they are tied together more clear. And it is easy enough for an extension to subclass the Processor class and override the relevant pieces. :+1:

facelessuser: > I wonder if a named tuple would make sense here.

I was actually considering doing a named tuple, so I fully agree.

facelessuser: > I hesitated for a minute as we don't get the granularity to override specific parts with extensions with this implementation.

I definitely understand this.  This is why I originally brought this discussion up in the related issue.  But ultimately, I felt the best way to get the discussion going was to just get some code up 🙂.

I'm okay if we want to just insert the new pattern between other patterns via the old way.  We lose the pattern steps by 10s, but you retain similar layout. I'm open to either way if we change our mind before merging, but at least now it is clear what the alternative would look like.

waylan: If I had never seen the old way, I would find this less method surprising that what we have now. We will just need to document the change in the release notes as it could be backward incompatible for some extensions. Unfortunately, there is probably not a way to gracefully deprecate the old patterns.

Although, if `markdown.util.Registry.deregister` defaulted to `strict=False`, that might not be an issue. Of course, it defaults to `strict=true` right now. Not sure I want to change the default though.

facelessuser:
PyYaml 5.1 wrecking Python 2 tests 😡 .

waylan:
> PyYaml 5.1 wrecking Python 2 tests.

That *is* annoying. I understand why PyYAML made the changes they did, but we were relying on the fact that the PyYAML and json libs both used the same API and PyYAML could parse JSON. We also avoided requiring PyYAML as a hard dependency as we used the json lib as a fallback. Now we need a more complicated implementation to account for the differences between the libs' APIs.

facelessuser:
Do you know why it is only failing on Python 2?

waylan:
> Do you know why it is only failing on Python 2?

I do not. The lowdown on the changes are [here][1]. I suppose we could simply disable the warning as an easy fix. We only use the YAML loader in the CLI so if there is a vulnerability there, the user has much worse problems that arbitrary code injected in a YAML config file.

[1]: https://github.com/yaml/pyyaml/wiki/PyYAML-yaml.load(input)-Deprecation

waylan:
The PyYAML issue has been addressed in #806. Feel free to rebase your work on master now.

facelessuser:
Ugh, everything broke for some reason... Probably need to rebase maybe

facelessuser:
Tests appear to be broken due to the latest tox 3.8+.  By forcing 3.7 tox, tests begin passing again.

facelessuser:
Maybe related to https://github.com/tox-dev/tox/issues/1236.

waylan:
Let's see what tox does before we address this. If they fix it quickly in the next release, then we might not need to do anything.

facelessuser:
Yeah, I'm obviously in no hurry as I've been taking my time. The current work around was just to identify what I figured was the problem, it isn't meant to be the official solution unless we are desperate.

facelessuser:
Okay, I finally got around to rebasing this. Obviously `tox` is no longer broken, so I reverted the workaround. I'm pretty much done with the work here unless there are further requests. In the future, maybe the `NOT_STRONG_RE` logic could be incorporated into this, but that would really require rewriting some of the existing patterns which I'm not looking to do at this time.

waylan:
Look good. Just needs a note in the release notes. Feel free to added a new `release-3.2.md` file.

waylan:
As we are removing many existing inline processors and replacing them with two new ones, I think we probably need to make mention of that in the release notes.  For example, if an extension is removing all of the existing inline processors, the dev will need to do some work to make it compatible. The release notes needs to call this out. That's why I suggested `3.2` rather than `3.1.1` for the release.

facelessuser:
Makes sense. I'll adjust within the next day or two.

PR #806, opened 2019-03-14
Update CLI to support PyYAML 5.1

waylan:
This should avoid any warnings. We use `unsafe_load` because users may
need to pass in actual Python objects. As this is only available from
the CLI, the user has much worse problems if an attacker can use this
as an attach vector.

See yaml/pyyaml#265 for an explanation of the changes in 5.1.

### DIFF
--- a/markdown/__main__.py
+++ b/markdown/__main__.py
@@ -26,9 +26,17 @@ import codecs
 import warnings
 import markdown
 try:
-    import yaml
+    # We use `unsafe_load` because users may need to pass in actual Python
+    # objects. As this is only available from the CLI, the user has much
+    # worse problems if an attacker can use this as an attach vector.
+    from yaml import unsafe_load as yaml_load
 except ImportError:  # pragma: no cover
-    import json as yaml
+    try:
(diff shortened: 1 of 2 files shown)


markdown/__main__.py
```diff
@@ -26,9 +26,17 @@
...
 try:
-    import yaml
+    # We use `unsafe_load` because users may need to pass in actual Python
+    # objects. As this is only available from the CLI, the user has much
+    # worse problems if an attacker can use this as an attach vector.
+    from yaml import unsafe_load as yaml_load
 except ImportError:  # pragma: no cover
-    import json as yaml
+    try:
+        # Fall back to PyYAML <5.1
```
waylan, line 35: I should point out that we don't list PyYAML as a dependency, so we can't force a required version. The way it works is that we support both YAML and JSON. If the PyYAML lib is installed, the user may use a either a JSON or YAML config file, either of which is parsed by PyYAML. But if PyYAML is not installed, then the user may only use a JSON config file which is parsed by the `json` lib in the Python Standard Library.

PR #813, opened 2019-04-09
CodeHilite: allow custom lexers

AshleyYakeley:
This allows custom Pygments lexers to be specified in the configuration.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -22,7 +22,7 @@ from ..treeprocessors import Treeprocessor
 
 try:
     from pygments import highlight
-    from pygments.lexers import get_lexer_by_name, guess_lexer
+    from pygments.lexers import get_lexer_by_name, guess_lexer, load_lexer_from_file
     from pygments.formatters import get_formatter_by_name
     pygments = True
 except ImportError:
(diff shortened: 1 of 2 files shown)


facelessuser:
I'm confused. You can already use custom lexers by creating a pygments lexer according to their documentation. It should use the appropriate entry points, and if it does, it will be recognized. I'm not sure this is needed. Can you clarify why you think something like this is needed?

AshleyYakeley:
I am using MkDocs, which uses Python-Markdown. I can't figure out any other way to get it to load custom Pygments lexer files.

facelessuser:
I've done them before, you have to specify their entry points and then install them. Here  I have two lexers that I setup with entry points: https://github.com/facelessuser/pymdown-lexers/blob/master/setup.py#L7.

I install my custom lexer package, and then I specify fences with my custom lexer name.

AshleyYakeley:
Oh this makes a lot more sense than fiddling with the CodeHilite code.

Thanks for pointing this out...

facelessuser:
Not a problem. The idea is to work within the design constraints of Pygments instead of reimplementing Pygments plugin access here.

You can use that repo I pointed to as a reference. You should be able to use the basic framework of it to get your own working. You can use a similar approach to do custom Pygments styles as well.

PR #816, opened 2019-04-12
Refactor fenced code attributes

### LINKED ISSUE #775: Support multiple classes for fenced code blocks
Right now, fenced code blocks can be assigned a class (which is typically used to define the language). But it is impossible to assign more than one class.

You may want to do this, for example, to distinguish `input` code blocks from `output` code blocks in a vignette. Right now, if you try to define multiple classes in a markdown fence block, using {.class1 .class2}, the HTML is not parsed correctly as a code block.

For a use case, see: https://github.com/mkdocs/mkdocs/issues/1720

You may want to say: `{.python .inputblock}` or `{.bash .outputblock}` to add a second layer of distinguishing code blocks.

waylan:
Addresses #775. A work in progress.

The behavior as of the first commit is that a single string (with an optional dot as first character), defines the language of the block:

With dot:

    ``` .lang

Without dot:

    ``` lang

If more than one thing is to be defined, you must use an attribute list (wrapped in curly brackets (`{}`). The first class (startswith a dot) is the language. All other classes are simply set on the rendered `code` element. The hash tag is the id and all `key=value` pairs are codehilite config values. You cannot define HTML attributes with `key=value` pairs.

    ``` {.lang .otherclass #id hl_lines='1 3'}

You can even set other config values that normally are global (per document) on a per block basis:

    ```{.lang linenums=true}

Although, actually that might not work (untested at this point) as we don't (yet) convert the string to a Boolean.

However, we have a backward incompatibility in that the existing tests include the following 2 cases where a key=value pair is used, but not in an attribute list:

     ```hl_lines="1 3"

and

     ```.python hl_lines="1 3"

With these changes, those will no longer be valid. In fact, the two tests are failing. This better matches the behavior of the PHP implementation we claim to copy. But it could break existing documents. We can't even claim it worked by accident as we explicitly tested for it.


markdown/extensions/fenced_code.py
```diff
@@ -66,29 +69,41 @@ def run(self, lines):
...
+                elif m.group('lang'):
+                    lang = m.group('lang')
+                    classes.append(lang)
 
                 # If config is not empty, then the codehighlite extension
                 # is enabled, so we call it to highlight the code
                 if self.codehilite_conf:
+                    local_config = self.codehilite_conf.copy()
+                    local_config.update(config)
+                    # TODO: pass id and classes to codehilite
```
martsberger: Thanks for this work. My use case is to put a class on the outer div that codehilite generates. In that case you can just append `' '.join(classes)` to the css_class parameter passed to CodeHilite. It looks like it will take some changes to CodeHilite itself to allow passing an id in.

waylan:
I resolved conflicts and added back in support for `hl_lines` outside of an `attr_list` for backward-compatibility.

waylan:
Got as far as I can with this without also modifying the codehilite extension, but that will be addressed in #822. 

This still needs documented.

waylan:
I'm using this for both fenced_code changes and codehilite changes. In other words this replaces/overrides #822. In the latest commit, I've made the changes to codehilite and updated the existing tests. However, I need to add more tests for the new behavior and document everything.

waylan:
I need to confirm if `{class="some string of classes"}` works correctly. It does for attr_list so it should here.

facelessuser:
Seems the answer is in your own comment:

> The hash tag is the id and all `key=value` pairs are codehilite config values. You cannot define HTML attributes with `key=value` pairs.

waylan:
@facelessuser you make a good point. I was thinking that this should reflect the behavior we find in attr_lists, but if `key=value` pairs are not used for HTML attributes, then that doesn't make sense. So, never mind.

waylan:
From @wlupton in #977:

> I guess `hl_lines` is special. 

Yes, but only for backward compatibility. Originally, only the language was supported. At some point we added support for `hl_lines`. Now that we are adding support for all of pygments options, `hl_lines` does not make sense in its current form. So as not to break existing documents, we continue to support it outside of brackets. Of course, you can define the language and hl_lines within brackets like anything else. Therefore, we may deprecate support outside of brackets in the future. 

> Is there any way that the attribute handling could be more common across extensions, e.g., I think that pandoc attributes are the same wherever they're supported (and one of the places where they're supported is fenced code blocks). From https://pandoc.org/MANUAL.html:
> 
> ```
> Optionally, you may attach attributes to fenced or backtick code block using this syntax:
> 
> ~~~~ {#mycode .haskell .numberLines startFrom="100"}
> qsort []     = []
> qsort (x:xs) = qsort (filter (< x) xs) ++ [x] ++
>                qsort (filter (>= x) xs)
> ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
> 
> Here mycode is an identifier, haskell and numberLines are classes,
> and startFrom is an attribute with value 100. Some output formats can
> use this information to do syntax highlighting.
> 
> [...] the code block above will appear as follows:
> 
> <pre id="mycode" class="haskell numberLines" startFrom="100">
>   <code>
>   ...
>   </code>
> </pre>
> ```
> 
> In fact (and this is probably the last thing you want to hear!) is there any way that the fenced code block extension could depend on the attr_list extension for processing its attribute lists?

Actually the list parser in the attr_list extension is imported and used by fenced-code as of this PR. However, it is only used for anything defined within the brackets. But as only a single-word language and `hl_lines` are supported outside of brackets (as explained above), for all intents and purposes, it is an attr_list.

However, it does differ from an attr_list in at least one way. As the `key=value` pairs are used as Pygments options, it does not make sense to include them in the syntax highlighted output. The exception being classes, which can be defined with keys. We need a way to differentiate between values that should be passed to Pygments, and values which should be assigned as attributes. Additionally, Pygments only provides a way to assign classes. Therefore, as many `.classnames` as are defined will get added to the `class` attribute. IDs get ignored because Pygments doesn't provide that option. Everything else is a Pygments option which effects the output, but doesn't get included in the output.

Now, that behavior is dependent on syntax highlighting being enabled (codehilite extension is enabled). Without syntax highlighting, then the ID gets set on the `pre` tag and `class` gets set on the `code` tag. I suppose we could also assign `key=value` pairs. However, in its current state, that doesn't happen. `key=value` pairs are simply ignored. This was done to match PHP's behavior of only allowing class and id to be set. After all, we claim to imitate that implementation. It also happens to match GitHub's implementation.

However, an argument could be made that including random attributes in the output would allow JavaScript highlighting libs to make use of them. And we have Pandoc as another implementation which already supports that. We also have received bug reports from time to time from users who expected any attribute to work if/when the attr_list is enabled. I suppose we could change behavior based on that by only assigning id and class by default, but assigning all defined attributes if attr_list is also enabled (and code highlighting is off).

waylan:
> We need a way to differentiate between values that should be passed to Pygments, and values which should be assigned as attributes.

I should point out that I considered hard-coding the options that Pygments accepts and passing only those to Pygments, but setting everything else as attributes in the HTML. However, there are a few issues with that:

1. Pygments doesn't provide an option to include random attributes in its output anyway.
2. Pygments adds new options from time to time and I don't want to create a situation where I need to update my list of options every them they add a new one.

Given the above, it makes more sense to simply pass everything blindly to Pygments. That way we automatically support any new pygments feature and still don't loose any functionality.

of course, all of that is dependent on code highlighting being turned on. When it's off, we are open to considering other options.

waylan:
I just added support for `attr_list`, which follows the following rules:

1. All key/value pairs must be defined within curly brackets (except for `hl_lines` which may be deprecated in the future).
2. If code highlighting is enabled, the attr_list extension is ignored. All key/value pairs are passed to Pygments as config options.
3. If code highlighting is disabled and the attr_list extension is enabled, all key/value pairs (except `use_pygments`; see item 7, below) are assigned as HTML attributes on the `code` tag.
4. If both code highlighting and the attr_list extension are disabled, all key/value pairs are ignored. Only `#id` and `.class` are used (with or without curry brackets).
5. You can disable code highlighting in multiple ways:
    1. Do not have Pygments installed.
    2. Set `use_pygments=False` on the extension config (globally).
    3. Use `{ use_pygments=False }` on an individual code block.
6. Even if code highlighting is disabled globally (as described in 5.i or 5.ii above), setting `{ use_pygments=True }` on an individual code block will cause key/value pairs to not be included at HTML attributes. It is assumed that they are Pygments config options. 
7. The `use_pygments` key is always removed from the list after its value is evaluated and is therefore never included in the output.

facelessuser:
>  (except for hl_lines which may be deprecated in the future).

Can you elaborate on this a bit more?

waylan:
> > (except for hl_lines which may be deprecated in the future).
>
> Can you elaborate on this a bit more?

@facelessuser, as I explained above:

> Yes, but only for backward compatibility. Originally, only the language was supported. At some point we added support for hl_lines. Now that we are adding support for all of pygments options, hl_lines does not make sense in its current form. So as not to break existing documents, we continue to support it outside of brackets. Of course, you can define the language and hl_lines within brackets like anything else. Therefore, we may deprecate support outside of brackets in the future.

Perhaps I should clarify by pointing out that if brackets are being used, then everything must be defined within them. When I say "outside of brackets" I mean no brackets are being used at all. In other words, at this time the following are both valid:

````
``` .lang hl_lines=3
````

````
``` { .lang hl_lines=3 }
````

In the future the first of those may be deprecated for consistency. In fact, I expect to not document that `hl_lines` can be defined outside of brackets. Users who already have documents which include it won't notice as their documents will continue to render the same. And everyone else won't even know its an option unless they read the source code and/or this discussion.

Finally, when I say it may be deprecated in the future, I am using "may" because I haven't decided for sure (perhaps "might" would have been a better word choice). If it does happen, it would first be 'pending deprecation' to give users an opportunity to update their documents. I'm not sure how that would work though, as we don't generally raise warnings while parsing a document. Warnings are generally only raised when initializing the `Markdown` class. And if it later becomes fully deprecated, that would mean the entire code block would not be recognized as one. Maybe some future pending deprecation would mean that `hl_lines` outside of brackets is simply ignored. I guess I haven't decided, in part, because I haven't figured out how to go about it.

Regardless, at this time I only intend to exclude mentioning the option in the docs. Any additional steps towards deprecation would be in a future release if they happen at all.

waylan:
I have finished updating the docs for fenced code. However, I just realized that I never expanded the codehilite config options. I only expanded the internal API which fenced_code calls. We need to expand the global options as well.

wlupton:
Sorry for the delayed response, but re 2 above:

> 2. If code highlighting is enabled, the attr_list extension is ignored. All key/value pairs are passed to Pygments as config options.

I wondered whether in this case it would be possible for Pygments to consume ("eat") the options that it understands, leaving the remainder to be available on the node, e.g., to be included as HTML attributes. This would seem to be quite a Pythonic approach (ref. [Python's super() considered super!](https://rhettinger.wordpress.com/2011/05/26/super-considered-super), which is referenced from the Python `super()` docs).

waylan:
> I wondered whether in this case it would be possible for Pygments to consume ("eat") the options that it understands, leaving the remainder to be available on the node, e.g., to be included as HTML attributes. 

I have no idea how this would work. Our code has no knowledge of the arguments which Pygments accepts. We simply blindly pass all arguments on to Pygments. It just so happens that Pygments was designed so that it ignores arguments it doesn't understand (rather than raising an error).

Regardless, as I put it in the [docs][1]: 

> The `fenced_code` extension does not alter the output provided by Pygments. Therefore, only options which Pygments provides can be ultilized. As Pygments does not currently provide a way to define an ID, any ID defined in an attribute list will be ignored when syntax highlighting is enabled. Additionally, any key/value pairs which are not Pygments options will be ignored, regardless of whether the `attr_list` extension is enabled.

In other words, I am placing the blame squarely on Pygments. And I have no interest in modifying Pygments output. If someone wants to do that, they can create their own third-party extension.

If anyone is going to argue that the current behavior is inconsistent, then the only change I would consider is to only accept Pygments arguments and never accept anything else.

[1]: https://github.com/Python-Markdown/markdown/pull/816/files#diff-196e0e1da6180999be437acb3259b8eeR233-R235

wlupton:
Thanks. Makes sense. Re this:

> It just so happens that Pygments was designed so that it ignores arguments it doesn't understand (rather than raising an error).

I have not looked at Pygments but at least in theory it could return the arguments that it hasn't consumed.

waylan:
> I have not looked at Pygments but at least in theory it could return the arguments that it hasn't consumed.

It does not. Perhaps a subclass could address that, but I'm not interested in going down that road.

waylan:
I think this is done and ready to go.

waylan:
After doing a final review, there are a few concerns I noted, which I would appreciate some feedback on.

When the `codehilite` extension is enabled but `use_pygments` is `False`, the HTML generated from  indented code blocks is currently different that that generated from fenced code blocks. The reason for this is that the fenced code block extension generates its own HTML, whereas codehilite uses own separate fallback to generate a non-highlighted code block with the language defined as a class.

I had considered using the Codehilite extension's fallback for fenced code blocks, but codehilite is a treeprocessor working on the ElementTree object, whereas fenced code blocks are handled via a preprocessor and there is no ElementTree object to work with. Therefore, we will need to continue to maintain two different implementations until such time as fenced code blocks are no linger implemented as a preprocessor.

The current differences are as follows:

1. There may be some insignificant white space differences. I'm not particularly concerned about this but include it for completeness.
2. Indented code blocks have no mechanism to define any attributes other than the language. This is a limitation of indented code blocks rather than an oversight. I don't intend to make any changes here. However, I do anticipate feature requests from users to add feature parity to indented code blocks, which I'm not looking forward to.
3. With this change, fenced code blocks now prepend `language-` to the language class as per the HTML5 spec's suggestion. This is the major concern. This could be considered a backward compatibility issue if a user's CSS and/or JavaScript was not expecting the prefix. And it becomes difficult for new users if the two systems behave differently. Their CSS and JavaScript would need to account for both. It seems that they should behave the same, but how?
    - We could remove the prefix from fenced code blocks (reverting the behavior) and use the reasoning that users can add the prefix manually when they define the language. This is PHP's approach.
    - We could add the prefix to the codehilite extension as well and use it everywhere, which would be a change for all existing users.
    - We could add a setting which enable/disables adding the prefix. Presumably the default would be off?

Adding a conf setting gets weird though. Currently fenced code blocks contains no settings of its own. It relies on codehilite for global highlighting settings, which makes sense as we don't want different settings for fenced code blocks compared to indented code blocks. However, this setting would only be relevant when `use_pygments==False`. Therefore we would need to define it (and the user would need to set it) separately for each extension. I suppose we could have the one inherit from the other. But which way would that go? After all, its also possible that each extension could be used independent of the other and still need the setting. Maybe fenced code defaults to whatever codehilite is set to unless codehilite is not enabled? That could be a pain to document in an understandable manner.

facelessuser:
My main concern would be the prefix dilemma as well.

I kind of thought that syntax highlighted code blocks were consistent. I say this with admitting that I do not use, nor have ever used the indented form with a highlight header provided by CodeHilite, so obviously, I can't exactly speak on that.

Regardless, I think it would make sense that the format is consistent across extensions provided by Python Markdown. I do not see any issue with the prefix option. Most JS highlighters handle `language-name` just fine as it is stated in HTML5's spec. Personally, I would have both formats output the same `language-name`. That is my personal opinion though.

---

A secondary concern would be the fact that `attr_list` cannot target code blocks. I feel this is a fundamental issue with how Python Markdown handles things and is not an easy fix. I would honestly recommend most people use fenced code blocks if they wish to add additional classes, ids, and/or attributes. This is a little tricky though because `fenced_code` doesn't nest under other constructs well.

I personally never use indented code blocks anymore, one big reason is that they normalize consecutive line endings to one. I'd rather use one format, so I chose fences which always retains the expected formatting (newlines included). With that said, using fences as the default format is only possible due to the fact that I use my own SuperFences extension which allows for nesting.

PR #822, opened 2019-05-11
CodeHilite: Support Pygments options

p-ouellette:
Allows specifying any Pygments option as suggested in https://github.com/Python-Markdown/markdown/issues/334. This also allows using the Pygments inline option for line numbers (#652).

Example:
```python
In [2]: text = """ 
   ...:     :::php 
   ...:     print "Hello world"; 
   ...: """                                                                                                            

In [3]: markdown(text, 
   ...:          extensions=['codehilite'], 
   ...:          extension_configs={'codehilite': { 
   ...:              'lexer_options': {'startinline': True}, 
   ...:              'formatter_options': {'wrapcode': True, 'linenos': 'inline'}}})                                   
Out[3]: '<div class="codehilite"><pre><span></span><code><span class="lineno">1 </span><span class="k">print</span> <span class="s2">&quot;Hello world&quot;</span><span class="p">;</span>\n</code></pre></div>'

```

There's some messiness where the Pygments options overlap with the existing options. I marked the ones that apply only to Pygments as deprecated, but some, like linenums and css_class are used when Pygments is disabled.

### DIFF
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -56,14 +56,18 @@ class CodeHilite(object):
     * src: Source string or any object with a .readline attribute.
 
     * linenums: (Boolean) Set line numbering to 'on' (True),
-      'off' (False) or 'auto'(None). Set to 'auto' by default.
+      'off' (False) or 'auto' (None). Set to 'auto' by default.
 
     * guess_lang: (Boolean) Turn language auto-detection
       'on' or 'off' (on by default).
 
     * css_class: Set class name of wrapper div ('codehilite' by default).
 
-    * hl_lines: (List of integers) Lines to emphasize, 1-indexed.
+    * lexer_options: Dictionary containing options passed to the Pygments lexer
+      if use_pygments is set to True.
+
+    * formatter_options: Dictionary containing options passed to the Pygments
+      formatter if use_pygments is set to True.
 
     Low Level Usage:
         >>> code = CodeHilite()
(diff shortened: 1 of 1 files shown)


mitya57:
I like this. It does not actually close #334 though, as this code does not allow one to set the options in a document (only via API). But I guess that can be the next step.

Also, can you make it issue some `DeprecationWarning`s when the deprecated options are used?

waylan:
> this code does not allow one to set the options in a document

See #816 which has started the work on that. Interestingly, this makes that more complicated. As this breaks the arguments up into 2 dicts to differentiate between lexer and formatter options. Previously, we didn't care about that distinction which made specifying options in the document easier. I'll have to think about that.

p-ouellette:
I wasn't thinking about setting the options in the document. I'll change it to use a single dict if requested. Then any key=value pair specified in the document that isn't a valid codehilite option could be passed to Pygments.

markdown/extensions/codehilite.py
```diff
@@ -75,7 +79,8 @@ class CodeHilite(object):
 
     def __init__(self, src=None, linenums=None, guess_lang=True,
                  css_class="codehilite", lang=None, style='default',
-                 noclasses=False, tab_length=4, hl_lines=None, use_pygments=True):
+                 noclasses=False, tab_length=4, use_pygments=True,
+                 lexer_options={}, formatter_options={}):
```
waylan, line 83: I'm thinking we should just do `**options` here. That does two things:

1. It avoids the extra level of nesting when defining the configs.
2. It avoids separating lexer and formatter options.

As it stands now, Pygments uses `**options` in their lexer and formatter classes, and ignores any unknown option keywords. So as long as there are no naming conflicts between lexer and formatter options, we could pass all options to both the lexer and formatter. This has the added benefit of making it easier to define options from within documents (see #816).

waylan:
I'm closing this in favor of #816. However, @pauloue your work was not in vain as it helped me to see how to best approach the matter.

PR #824, opened 2019-05-13
Split version info into a separate file, load it using imp/importlib

### LINKED ISSUE #823: setup.py shouldn't import the package it installs
There is a :chicken: and :egg: problem regarding how setup.py works. the version number should be parsed from the text file rather than imported via python mechanism.

```
    File "setup.py", line 26, in <module>
      from markdown import __version__, __version_info__
  ModuleNotFoundError: No module named 'markdown'
```

mitya57:
To make sure `markdown` module is importable during installation.

It fixes Travis tests and also fixes `python3 -m pep517.check`, so I hope it closes #823.

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -3,6 +3,11 @@ title: Change Log
 Python-Markdown Change Log
 =========================
 
+_______, 2019: Released version 3.1.1 (a bug-fix release).
+
+* Fixed import failure in `setup.py` when the source directory is not
+  on `sys.path` (#823).
+
 Mar 25, 2019: Released version 3.1 ([Notes](release-3.1.md)).
 
 Sept 28, 2018: Released version 3.0.1 (a bug-fix release).
--- a/markdown/__meta__.py
+++ b/markdown/__meta__.py
@@ -1,0 +1,53 @@
+# -*- coding: utf-8 -*-
+"""
+Python Markdown
+
+A Python implementation of John Gruber's Markdown.
+
+Documentation: https://python-markdown.github.io/
+GitHub: https://github.com/Python-Markdown/markdown/
+PyPI: https://pypi.org/project/Markdown/
+
+Started by Manfred Stienstra (http://www.dwerg.net/).
+Maintained for a few years by Yuri Takhteyev (http://www.freewisdom.org).
+Currently maintained by Waylan Limberg (https://github.com/waylan),
+Dmitry Shachnev (https://github.com/mitya57) and Isaac Muse (https://github.com/facelessuser).
+
+Copyright 2007-2018 The Python Markdown Project (v. 1.7 and later)
+Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)
+Copyright 2004 Manfred Stienstra (the original version)
+
+License: BSD (see LICENSE.md for details).
+"""
+
+from pkg_resources.extern import packaging
+
+# __version_info__ format:
+# (major, minor, patch, dev/alpha/beta/rc/final, #)
+# (1, 1, 2, 'dev', 0) => "1.1.2.dev0"
+# (1, 1, 2, 'alpha', 1) => "1.1.2a1"
+# (1, 2, 0, 'beta', 2) => "1.2b2"
+# (1, 2, 0, 'rc', 4) => "1.2rc4"
+# (1, 2, 0, 'final', 0) => "1.2"
+__version_info__ = (3, 1, 0, 'final', 0)
+
+
+def _get_version():  # pragma: no cover
+    " Returns a PEP 440-compliant version number from version_info. "
+    assert len(__version_info__) == 5
+    assert __version_info__[3] in ('dev', 'alpha', 'beta', 'rc', 'final')
+
+    parts = 2 if __version_info__[2] == 0 else 3
+    v = '.'.join(map(str, __version_info__[:parts]))
+
+    if __version_info__[3] == 'dev':
+        v += '.dev' + str(__version_info__[4])
+    elif __version_info__[3] != 'final':
+        mapping = {'alpha': 'a', 'beta': 'b', 'rc': 'rc'}
+        v += mapping[__version_info__[3]] + str(__version_info__[4])
+
+    # Ensure version is valid and normalized
+    return str(packaging.version.Version(v))
+
+
+__version__ = _get_version()
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,6 +25,7 @@ matrix:
           packages:
           - aspell
           - aspell-en
+    - env: TOXENV=pep517check
 
 addons:
   apt:
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -25,46 +25,13 @@ from __future__ import unicode_literals
 from .core import Markdown, markdown, markdownFromFile
 from .util import PY37
 from .pep562 import Pep562
-from pkg_resources.extern import packaging
+from .__meta__ import __version__, __version_info__
 import warnings
 
 # For backward compatibility as some extensions expect it...
 from .extensions import Extension  # noqa
 
 __all__ = ['Markdown', 'markdown', 'markdownFromFile']
-
-# version must conform to PEP 440
-# https://www.python.org/dev/peps/pep-0440/
-
-# __version_info__ format:
-# (major, minor, patch, dev/alpha/beta/rc/final, #)
-# (1, 1, 2, 'dev', 0) => "1.1.2.dev0"
-# (1, 1, 2, 'alpha', 1) => "1.1.2a1"
-# (1, 2, 0, 'beta', 2) => "1.2b2"
-# (1, 2, 0, 'rc', 4) => "1.2rc4"
-# (1, 2, 0, 'final', 0) => "1.2"
-__version_info__ = (3, 1, 0, 'final', 0)
-
-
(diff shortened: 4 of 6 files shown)


waylan:
I wonder if we should revert back to including the old [`get_version`][1] function we used to have in `setup.py`.  The one concern is that the old solution explicitly avoided importing Markdown. Now that we include the version right in `markdown/__init__.py` its not possible to avoid importing Markdown. And I am not suggesting that we move the version back to a separate `__version__.py` file.

It's just that this feels like an dirty hack while that feels much cleaner as the module is never added to the PYTHONPATH and does not pollute any namespace. And as we used it for years, it is well tested and already developed.

[1]: https://github.com/Python-Markdown/markdown/commit/997063da6ae18692eab45e8d695e9dbc5aea3efe#diff-2eeaed663bd0d25b7e608891384b7298L8-L30

facelessuser:
I kind of agree with the old way. I currently do something very similar in my packages and haven't felt the need to change so far.

waylan:
And while we're at is perhaps we might want to add `python3 -m pep517.check path/to/source ` as a test env to tox. That should give us quick and easy to identify feedback if/when a problem is related to the build environment.

mitya57:
Thanks for the link to the old code. I don't mind reverting to it, but in such case it would be nice to replace deprecated [imp](https://docs.python.org/3/library/imp.html) (which was used in setup.py) with [importlib](https://docs.python.org/3/library/importlib.html).

Otherwise the approach is basically the same as here — passing `[path]` to `find_module` has a similar effect to changing `sys.path`.

And also +1 for running `pep517.check` as part of CI.

mitya57:
Replying to myself:
> Thanks for the link to the old code. I don't mind reverting to it, but in such case it would be nice to replace deprecated imp (which was used in setup.py) with importlib.

It [looks like](https://stackoverflow.com/a/67692) there is no way to pass a path to importlib that will work with both Python 2 and Python 3. So we have these options:

- Use deprecated imp on Python 3 too (which is not nice);
- Use different code paths for different Python versions;
- Add the path to sys.path (and maybe remove it back after importing) — which will be equivalent to what I am doing in this PR;
- Drop support for Python 2;
- Do something very different (like extracting the version from file using regular expressions).

waylan:
> It looks like there is no way to pass a path to importlib that will work with both Python 2 and Python 3. 

`importlib.import_module` is [available][1] in Python 2.7 as a simple wrapper around the old API for this very reason. Presumably we could adopt the Python3 solution now without dropping Python 2.7 yet.

[1]: https://docs.python.org/2/library/importlib.html

waylan:
And I just realized `importlib.import_module` won't work for us as it doesn't accept a file path. I suppose we could create a function with multiple paths through if statements or caught exceptions. But that seems ridiculous when dropping Python 2.7 is so immanent.

Let's just keep the existing change for now. We can reexamine when we drop Python 2.7. Add the `pep517.check` in this PR and I'll merge it.

mitya57:
> Add the `pep517.check` in this PR and I'll merge it.

Done.

facelessuser:
While I'm not concerned with how this is ultimately solved here (as long as it works 🙂), recreating something similar to the old way without imp isn't too bad. I've been able to do it here: https://github.com/facelessuser/pymdown-extensions/pull/569. In my case, I just left the old PY2 implementation under a conditional as I simply plan on ripping it out when PY2 is finally put down.

Anyways, just options.

waylan:
If someone wants to do the work, I'm open to @facelessuser's suggestion.

waylan:
> facelessuser/pymdown-extensions#569.

I'm not crazy about that using `exec`. I'm open to a solution which does not use `exec`.

mitya57:
The [StackOverflow answer I posted above](https://stackoverflow.com/a/67692) has a solution without `exec` (it uses importlib’s `exec_module` instead).

I can copy it here however I don’t see the benefits compared to my approach. Are there any?

waylan:
>  I don’t see the benefits compared to my approach. Are there any?

As I stated before...

> It's just that this feels like an dirty hack while that feels much cleaner as the module is never added to the PYTHONPATH and does not pollute any namespace.

facelessuser:
> The StackOverflow answer I posted above has a solution without exec (it uses importlib’s exec_module instead).

Either way is fine. Some people get uncomfortable using `exec`. I think in the proper situations it's fine, but I have no strong feelings either way.

> I can copy it here however I don’t see the benefits compared to my approach. Are there any?

And as far as benefits over this pull 🤷‍♂ . I'm not personally using `toml` files yet, so I'm still hanging around with  with how I was doing things previously. As the possibility of just importing a version file was discussed, I simply just wanted to provide how it **could** be done.

Ultimately, I'm not concerned with what gets done in the end. My *preference* is just importing the version stuff as a separate file, but if adding the `cwd` and importing `markdown` is preferred, I've no problems if it's working.

facelessuser:
I just tried out `spec_from_file_location` from file, and it works out very well too. I think it is a much cleaner option. Since I don't think we are supporting Python 3.4, I don't see a reason to not use it, and just use `imp` for Python 2.

mitya57:
> If someone wants to do the work, I'm open to @facelessuser's suggestion.

Done. Please review the latest version.

facelessuser:
LGTM

markdown/__meta__.py
```diff
@@ -0,0 +1,53 @@
...
+Dmitry Shachnev (https://github.com/mitya57) and Isaac Muse (https://github.com/facelessuser).
+
+Copyright 2007-2018 The Python Markdown Project (v. 1.7 and later)
+Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)
+Copyright 2004 Manfred Stienstra (the original version)
+
+License: BSD (see LICENSE.md for details).
+"""
+
+from pkg_resources.extern import packaging
```
waylan, line 23: While we're working on this, might as well address #825 as well.

waylan (commented):
I'm not crazy about moving the version stuff back out to a separate file (I suspect it would have worked fine with that left in `__init__.py`) but it's fine. And `__meta__.py` makes sense as a name.

mitya57:
> I'm not crazy about moving the version stuff back out to a separate file (I suspect it would have worked fine with that left in `__init__.py`) but it's fine. And `__meta__.py` makes sense as a name.

It worked on Python 3 but not on Python 2:
```python
  File "setup.py", line 46, in <module>
    __version__, __version_info__ = get_version()
  File "setup.py", line 36, in get_version
    markdown = imp.load_source('markdown', module_path)
  File "markdown/__init__.py", line 25, in <module>
    from .core import Markdown, markdown, markdownFromFile
ValueError: Attempted relative import in non-package
```

That’s why I had to split it into a separate file.

markdown/__meta__.py
mitya57: I need your opinion — are you OK with adding a new dependency on [packaging](https://pypi.org/project/packaging/), or I should leave a fallback to `pkg_resources.extern.packaging`?

waylan: My initial reaction is to leave the fallback. That's what I intended with my previous comment. 

But what about that user who doesn't have the fallback **and** doesn't have packaging installed?

Of course, this will most likely be a problem when users are installing from their OS packaging system (such as ArchLinux as reported). But then ArchLinux could include `packaging` as a dependency to their `markdown` package. After all, they created the problem for themselves.

But that doesn't help the user who installs `pkg_resources` from ArchLinux, but `markdown` from PyPI as the original reporter in #825 did.  

On second thought, this is more complicated that I was thinking. Let's leave it for a separate PR. Sorry.

waylan:
With a comment in the release notes, this should be good to go.

markdown/__meta__.py
mitya57: > But that doesn't help the user who installs `pkg_resources` from ArchLinux, but `markdown` from PyPI as the original reporter in #825 did.

Note that the `setuptools` package in ArchLinux depends on `packaging` — it was the point of their patching to depend on it rather than bundle it.

So unless `pkg_resources` stops bundling `packaging` or otherwise changes its private API, there will be no problem with fallback.

Anyway, I suggested another option in #825 that you may like better.

mitya57:
> With a comment in the release notes, this should be good to go.

Added.

markdown/__meta__.py
eli-schwartz: markdown depends on pkg_resources, and pkg_resources depends on packaging. But, the existence of its extern.packaging is an implementation detail (it could always be refactored to be pkg_resources._vendor.packaging, for example), and the only public API of pkg_resources is get_version as mentioned in the other issue.

As @mitya57 said, the only reason this undocumented functionality actually breaks on Arch Linux is because we de-vendor "packaging" so that pkg_resources depends on it rather than its internal code copy.

docs/change_log/index.md
```diff
@@ -3,6 +3,11 @@ title: Change Log
 Python-Markdown Change Log
 =========================
 
+_______, 2019: Released version 3.1.1 (a bug-fix release).
+
+* Fixed import failure in setup.py when the source directory is not
```
facelessuser: Maybe wrap `setup.py` and `sys.path` in backticks to avoid spellchecker from seeing them?

mitya57: Done.

greut:
@mitya57 good job! much appreciated.

PR #831, opened 2019-06-12
Remove `install_requires=['setuptools >= 36']` from setup.py

jdufresne:
A modern seutptools is required at build time, not install time. Adding
setuptools to install_requires does not play nice with pip-tools. It
leaves a very large warning at the bottom of requirements.txt files:

    # WARNING: The following packages were not pinned, but pip requires them to be
    # pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
    # setuptools==41.0.1        # via markdown

This becomes an issue when installing packages with --require-hashes as
now setuptools is not pinned. This results in the error:

    In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
        setuptools>=36 from ... (from markdown==3.1.1->-r .../requirements.txt (line 30))

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,8 +34,7 @@ addons:
     - libtidy-0.99-0
 
 install:
-  # NOTE: setuptools needs to be installed explicitly for py34 (trusty).
-  - pip install 'setuptools>=36' tox
+  - pip install tox
 
 script:
   - tox
--- a/markdown/__meta__.py
+++ b/markdown/__meta__.py
@@ -20,11 +20,6 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
-try:
-    import packaging.version
-except ImportError:
-    from pkg_resources.extern import packaging
-
 # __version_info__ format:
 # (major, minor, patch, dev/alpha/beta/rc/final, #)
 # (1, 1, 2, 'dev', 0) => "1.1.2.dev0"
@@ -49,8 +44,7 @@ def _get_version():  # pragma: no cover
         mapping = {'alpha': 'a', 'beta': 'b', 'rc': 'rc'}
         v += mapping[__version_info__[3]] + str(__version_info__[4])
 
-    # Ensure version is valid and normalized
-    return str(packaging.version.Version(v))
+    return v
 
 
 __version__ = _get_version()
--- a/setup.py
+++ b/setup.py
@@ -95,7 +95,6 @@ setup(
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=2.7,!=3.0.*,!=3.1.*,!=3.2.*,!=3.3.*,!=3.4.*',
-    install_requires=['setuptools >= 36'],
     extras_require={
         'testing': [
             'coverage',
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -1043,3 +1043,16 @@ class TestGeneralDeprecations(unittest.TestCase):
         self.assertTrue('__version__' in dir_attr)
         self.assertFalse('version_info' in dir_attr)
         self.assertTrue('__version_info__' in dir_attr)
+
+
+class TestVersion(unittest.TestCase):
+    def test_version(self):
+        self.assertIsInstance(markdown.__version_info__, tuple)
+        self.assertIsInstance(markdown.__version__, str)
+        # PEP 440 compliant.
+        # https://www.python.org/dev/peps/pep-0440/
+        regex = r'[0-9]+\.[0-9]+\.[0-9]+(?:(?:a|b|rc|\.dev)[0-9]+)?'
+        if PY3:
+            self.assertRegex(markdown.__version__, regex)
+        else:
+            self.assertRegexpMatches(markdown.__version__, regex)


waylan:
> A modern setuptools is required at build time, not install time.

Well, then we need a `build_requires...` which doesn't exist as far as I am aware. And without it this is the only way to get that. For example the `pyproject.toml` file specifies that the `build system` requires `setuptools >= 36` and that should be enough, except that older setuptools won't check for the existence of the `pyproject.toml` file let alone enforce its requirements. The only way to do that is to define the requirements in the `seup.py` file. If you can provide an alternative which enforces the requirement where necessary, then I'll consider it, but so far all this change does is remove a requirement that still exists.

jdufresne:
Would you like me to add `setup_requires`?

https://setuptools.readthedocs.io/en/latest/setuptools.html#new-and-changed-setup-keywords

> setup_requires
>
> A string or list of strings specifying what other distributions need to be present in order for the setup script to run. setuptools will attempt to obtain these (even going so far as to download them using EasyInstall) before processing the rest of the setup script or commands. This argument is needed if you are using distutils extensions as part of your build process; for example, extensions that process setup() arguments and turn them into EGG-INFO metadata files.
>
> (Note: projects listed in setup_requires will NOT be automatically installed on the system where the setup script is being run. They are simply downloaded to the ./.eggs directory if they’re not locally available already. If you want them to be installed, as well as being available when the setup script is run, you should add them to install_requires and setup_requires.)

mitya57:
> A modern seutptools is required at build time, not install time.

It was added as run-time dependency to make sure `pkg_resources.extern.packaging` module is available. However now we have [this line][1], so the dependency *should* look like `setuptools>=36 OR packaging`. However I am not aware of a way to specify alternative dependencies…

[1]: https://github.com/Python-Markdown/markdown/blob/master/markdown/__meta__.py#L23

jdufresne:
Added `setup_requires` requires to the latest revision.

mitya57:
`setup_requires` does not make much sense as we already specify a build-time dependency in `pyproject.toml`.

If the needed version of setuptools is not installed then (modern) pip won't even call `setup.py`.

waylan:
@mitya57 reminded me that we actually need this as run time at [markdown/__meta__.py#L23-L26][1]. Either a modern version of setuptools or `packaging` is an **install** requirement. However, as Python-Markdown has never had any dependencies (except the recent adoption of setuptools) I'd like to not have `packaging` as a requirement if I can help it. Therefore the requirement that setuptools be a modern version.

As an aside, some setuptools distributions (usually from OS packaging systems) actually list `packaging` as a dependency rather than a vendored lib of setuptools. Therefore, we can be sure one or the other is available for any recent version of setuptools. Thus the reason we attempt to import both.

[1]: https://github.com/Python-Markdown/markdown/blob/master/markdown/__meta__.py#L23-L26

jdufresne:
Ok, I've gone back to keeping `setup_requires` out of `setup.py`. Are there any other changes you're looking for this to pass review?

mitya57:
I would say that we do not want to merge this change for the reasons described above.

We depend on `pkg_resources` (which is provided by setuptools), but we do not want to pin the required version of it. If every package required some particular version of setuptools, a conflict would arise immediately. So requiring setuptools with `>=` relation is the right thing to do here. Even if we changed that to `packaging`, that would not change things in any way.

Also, Markdown is not the only package that depends on setuptools/pkg_resources. How do other such packages play with pip-tools?

jdufresne:
> but we do not want to pin the required version of it.

Just to be clear, I'm not suggesting this and agree that it is undesirable. The pinning happens by the end users, not the intermediate dependencies.

My suggestion is to drop the `install_requirements`, not pin it.

> Also, Markdown is not the only package that depends on setuptools/pkg_resources. How do other such packages play with pip-tools?

The dependencies of my projects don't depend on any. This is the first time I've hit this issue. But I can research this and get back to you. Do you have a specific package in mind? Perhaps it is possible to remove the runtime dependency.

mitya57:
> The dependencies of my projects don't depend on any. This is the first time I've hit this issue. But I can research this and get back to you. Do you have a specific package in mind?

E.g. Sphinx also has setuptools in install_requires: https://github.com/sphinx-doc/sphinx/blob/master/setup.py#L32

> Perhaps it is possible to remove the runtime dependency.

Can you explain to me what pip-tools expects? I have never used it. Does it expect packages to have no runtime dependencies without pinning versions?

We currently use pkg_resources in two places: for the version class and for the `iter_entry_points` function. In theory we can use our own class for versions, but for the `iter_entry_points` we do need some external dependency, be it pkg_resources or importlib_metadata or whatever.

jdufresne:
I have updated the PR to drop the runtime requirement on setuptools. This removes the need to include it in `setup.py`. It was being used to generate the `__version__` string, however, this is very simple to do without any helper libraries, so take the simpler and more compatible approach.

I added a test that the version is always well formed. Let me know what you think.

> E.g. Sphinx also has setuptools in install_requires: https://github.com/sphinx-doc/sphinx/blob/master/setup.py#L32

Thank you! I actually have Sphinx as a dependency, so this was quite useful to help debug. It turns out, simply having a setuptools dependency isn't enough to be an issue. It is only an issue when setuptools is a transient dependency **and** pip wants to upgrade it. At this point pip sees that the transient dependency is not pinned with a hash and so fails with the error pasted above. I was installing markdown in a very limited environment, that explains why the `'setuptools' >= 36` dependency kicked in.

waylan:
> I have updated the PR to drop the runtime requirement on setuptools. This removes the need to include it in setup.py. It was being used to generate the `__version__` string, however, this is very simple to do without any helper libraries, so take the simpler and more compatible approach.

This is an unacceptable approach. The purpose is to ensure we have a valid version number, which is much more complicated that simply jointing the parts of `__version_info__` into a string. We absolutely must keep this dependency.

jdufresne:
> The purpose is to ensure we have a valid version number, which is much more complicated that simply jointing the parts of __version_info__ into a string. 

Can you explain what cases are not covered by the approach in this PR? I'm happy to add a test for them.

waylan:
I have no interest in maintaining an implementation separate from the official implementation in `packaging`.

jdufresne:
I agree with this sentiment. I'm not suggesting markdown do so.

packaging is responsible for taking an arbitrary string, validating it, and normalizing it to a standard form.

Meanwhile markdown doesn't need to do this. markdown takes a well defined, fully controled data structure (tuple) and converts it to a string of a known format. The logic and testing surface area is much much smaller. packaging doesn't need to be involved for this step and is solving a very different problem.

waylan:
I'm not confident that i (or anyone else) will properly populate the tuple. In fact, my assumption is that I won't at some point. Using packaging ensures we always get a valid version number. That is not up for debate. I will not consider not using packaging.

PR #833, opened 2019-06-12
Remove unnecessary calls to dict.keys()

jdufresne:
In modern Python, iter(dict) is equivalent to iter(dict.keys()). There
is normally no reason to add the extra method call.

Inspired by Lennart Regebro's talk "Prehistoric Patterns in Python" at
PyCon 2017.

https://www.youtube.com/watch?v=V5-JH23Vk0I

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -210,7 +210,7 @@ class Markdown(object):
         try:
             self.serializer = self.output_formats[self.output_format]
         except KeyError as e:
-            valid_formats = list(self.output_formats.keys())
+            valid_formats = list(self.output_formats)
             valid_formats.sort()
             message = 'Invalid Output Format: "%s". Use one of %s.' \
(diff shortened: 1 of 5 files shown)


waylan:
What benefit does this provide?

Note that the [Contributing Guidelines][1] specifically state:

Unless there is some specific performance benefit to this change, it seems to fall under this category.

jdufresne:
To me, the benefits are:

1. Reduce unnecessary method calls. (This is a tiny tiny performance boost. But I re-iterate, very tiny and not enough on its own.)
2. Follows community conventions and practices.
3. Modernizes the code to make it friendlier to new contributors and tools.

In response to the policy statement, this change is fully backwards compatible.

waylan:
In the spirit of the policy, I am not accepting this PR.

PR #834, opened 2019-06-12
Run pyupgrade across the code base

jdufresne:
The pyupgrade tool automatically upgrades syntax and patterns to newer
versions of Python.

- Use dict comprehension.
- Use set literals.
- Remove unnecessary numeric placeholders from string formatting.
- Replace modulus string formatting with .format().

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -259,7 +259,7 @@ class CodeBlockProcessor(BlockProcessor):
             code = sibling[0]
             block, theRest = self.detab(block)
             code.text = util.AtomicString(
-                '%s\n%s\n' % (code.text, util.code_escape(block.rstrip()))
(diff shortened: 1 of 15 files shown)


waylan:
What benefit does this provide?

Note that the [Contributing Guidelines][1] specifically state:

Unless there is some specific performance benefit to this change, it seems to fall under this category.

jdufresne:
To me, the benefits are:

- Set literals are always faster than calls to `set()`
- Follows modern community conventions and practices.
- Modernizes the code to make it friendlier to new contributors and tools.

In response to the policy statement, this change is fully backwards compatible.

waylan:
In the spirit of the policy, I am not accepting this PR.

Sometimes new is better, but sometimes its just different and this seems to be more of that later than the former.

PR #835, opened 2019-06-12
Use skip_install=true for lint targets

jdufresne:
Avoids installing the package (and any potential dependencies) to the
virtualenv before running lint or static commands. The package is not
required to be installed to do simple static code analysis. Results in a
slightly faster run, as fetching and installing dependencies is skipped.

For additional information on the configuration option, see:

https://tox.readthedocs.io/en/latest/config.html#confval-skip_install=BOOL

> Do not install the current package. This can be used when you need the
> virtualenv management but do not want to install the current package
> into that environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,7 @@
 envlist = py27, py35, py36, py37, pypy, pypy3, flake8, checkspelling, pep517check
 requires = setuptools>=36
 isolated_build = True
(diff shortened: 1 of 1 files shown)


waylan:
I can't be bothered to track every new config setting tox adds over the years. But, if others want to, and submit PRs that actually improve things, I'll take it.

PR #836, opened 2019-06-12
Use more specific asserts throughout tests

jdufresne:
In the event that a test fails, a more specific assert will usually
provide more useful information about what went wrong.

Prefer assertIs(..., True) over assertTrue() to assert the type as well
as the truthy value. This is recommended by Python:

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue

> Note that this is equivalent to bool(expr) is True and not to expr is
> True (use assertIs(expr, True) for the latter). This method should
> also be avoided when more specific methods are available (e.g.
> assertEqual(a, b) instead of assertTrue(a == b)), because they provide
> a better error message in case of failure.

### DIFF
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -144,8 +144,8 @@ class TestBlockParser(unittest.TestCase):
         """ Test BlockParser.parseDocument. """
         lines = ['#foo', '', 'bar', '', '    baz']
         tree = self.parser.parseDocument(lines)
-        self.assertTrue(isinstance(tree, markdown.util.etree.ElementTree))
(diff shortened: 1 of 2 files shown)


waylan:
So the new test methods were added in 2.7 and 3.1 and it has been relatively recently that we dropped support for 2.6, which is why we were not using them. Therefore, this could also fail under the [contributing policy][1]: 

Although, that is a style guide and this is actually a functional change. As there is a clear improvement here, I'll accept this. but @jdufresne please try to be a little more respectful of the project's history in any future PRs. The seeming lack of respect (whether intentional or not) for the history in many of your recent PRs  was very off-putting.

PR #838, opened 2019-06-12
Use https:// links where available

jdufresne:


tests/extensions/toc.html
```diff
@@ -696,4 +696,4 @@ <h2 id="backslash-escapes">Backslash Escapes</h2>
 -   minus sign (hyphen)
 .   dot
 !   exclamation mark
-</code></pre>
\ No newline at end of file
+</code></pre>
```
waylan: Adding this newline has caused the test to fail. Please remove the newline.

waylan (changes requested):

jdufresne:
Thanks for catching. Fixed and rebased.

waylan (approved):

PR #846, opened 2019-07-14
Update cli.md

### LINKED ISSUE #844: Include example JSON configuration
Great library! Thanks for making it.

I am trying to use it to convert some local md files to html. It took me some trial and error to figure out the format for a JSON conf file. I was using the toc extension and tried many permutations on the following JSON

```
{
  "extension_configs":
  {
    "toc":
    {
      "toc_depth":"2-6"
    }
  }
}
```

Than I found out I had an extra layer of nesting and my file should be:
```
{
  "toc":
  {
    "toc_depth":"2-6"
  }
}
```

I think including some a few sample in the docs can help people get an idea of the format of the configuration files faster. I am using markdown 3.1.1

sleepyhollo:
closes #844
Add a JSON config file example for the CLI

Outstanding questions
- do I need to write and/or run tests for the contribution to documentation?
- how would I add non-json type values in a json config file? For example, [PyMdown Extensions][pymdownx] [emoji][pymdownx-emoji] extension has an option `emoji_index` which seems to take a reference to a function or a module as its input. This can be seen in the [Github-ish Configurations][pymdownx-faq] FAQ in PyMdown's docs. I tried passing `"emoji_index": "pymdownx.emoji.gemoji"` through a json config file, but that failed. So perhaps thats a limitation of the json configs file...

[pymdownx]: https://facelessuser.github.io/pymdown-extensions/
[pymdownx-emoji]: https://facelessuser.github.io/pymdown-extensions/extensions/emoji/
[pymdownx-faq]: https://facelessuser.github.io/pymdown-extensions/faq/

### DIFF
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -160,7 +160,7 @@ myext:
     option2: True
 ```
 
-Similarly, a [JSON] config file might look like this:
+Similarly, a [JSON] configuration file might look like this:
 
 ```json
 {
--- a/docs/cli.md
+++ b/docs/cli.md
@@ -174,7 +174,9 @@ Similarly, a [JSON] configuration file might look like this:
 
 Note that while the `--extension_configs` option does specify the
 `myext` extension, you still need to load the extension with the `-x` option,
-or the configuration for that extension will be ignored.
+or the configuration for that extension will be ignored. Further, if an
+extension requires a value that cannot be parsed in [JSON] (for example a
+reference to a function), one has to use a YML configuration file.
 
 The `--extension_configs` option will only support YAML configuration files if
 [PyYAML] is installed on your system. JSON should work with no additional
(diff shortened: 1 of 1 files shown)


docs/cli.md
```diff
@@ -160,6 +160,18 @@ myext:
     option2: True
 ```
 
+Similarly, a [JSON] config file might look like this:
```
waylan: `config` is [not a properly spelled word][1]. Please use the full word: `configuration`.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/558492124#L310

waylan:
> do I need to write and/or run tests for the contribution to documentation?

No, that's not necessary. Or at least we never have.

> how would I add non-json type values in a json config file?

You don't. Which is why we also support YAML. YAML has specific support for specifying Python types. In fact, JSON is a limited subset of YAML.

PR #855, opened 2019-08-01
Added trivial example to add support for fenced code blocks, fixes #854

### LINKED ISSUE #854: Code blocks with language not rendering correctly
When trying to markdown a code block with a specified language, the language is not added as a class like the docs seem to suggest, instead it just appears inside the code block.

Take the following example
```python
markdown.markdown("""```python
print('test')
```""")
```

This outputs
```
<p><code>python\nprint('test')</code></p>
```

What I'd expect to see is
```
<pre><code class="python">print('test')</code></pre>
```

I appreciate this part of the docs is for a different case but it's suggesting that the regular markdown syntax would produce the same output.
https://python-markdown.github.io/extensions/fenced_code_blocks/#language

Note also the use of `<p>` tags instead of `<pre>` which means when rendered in the browser any line breaks are not seen.

For reference, `markdown.__version__` reports I'm on 3.1.1 which appears to be the latest.

Jamie-:


waylan (changes requested):
Thanks for your contribution. Let's keep this well formed:

1. The code block should be wrapped in blank lines.
2. To avoid conflict with the fence within the example, the opening and closing fences should use a different number of backticks. I suggest 4.
3. This is Python code so let's identify it as such.

Also, while using string literals makes for compact Python code, it is not the easiest to read. Lets just pass in a variable which is assumed to be defined elsewhere (as we do in the examples elsewhere in the docs).

Finally, if we are going to do this for one extension, we should do this for all of them. The docs for every extension needs to be updated before I'll accept this.

Jamie-:
Is that better?

waylan (approved):

Issue #760, opened 2018-11-27
Python Version Support Timeline


waylan:
Just noting this so I can find it later. Python defines the status of its versions <del>[here][1]</del><ins>[here](https://devguide.python.org/versions/)</ins>.

The following end-of life cycles are currently scheduled:

* <del>Python 3.4 2019-03-16</del>
* <del>Python 2.7 2020-01-01</del>
* <del>Python 3.5 2020-09-13</del>
* <del>Python 3.6 2021-12-23</del>
* <del>Python 3.7 2023-06-27</del>
* <del>Python 3.8 2024-10<del>
* <del>Python 3.9 2025-10</del>
* Python 3.10 2026-10
* Python 3.11 2027-10
* Python 3.12 2028-10
* Python 3.13 2029-10
* Python 3.14 2030-10 
* Python 3.15 2031-10 (expected release date: 2026-10-01)

<del>Also of interest is https://python3statement.org/ where various projects have signed on to a pledge to discontinue support for Python 2,7 on or before 2020. I'm indifferent to joining the pledge, but the arguments given for dropping support could be useful when we take that move.</del>

I intend to keep this issue open indefinitely to track future updates. As a general rule, I expect we will drop support for a Python version with our next minor release after Python has dropped support. That said, if/when <del>Travis</del><ins>GitHub Actions</ins> drops support, our tests stop passing, so we have occasionally dropped support quickly in the past. Of course, as per our Contributing Guidelines, such a change should not happen in a bugfix release.

[1]: https://devguide.python.org/#status-of-python-branches

facelessuser:
I personally don't support Python 2.7 on any new projects if I don't have a direct need. This has been a personal choice as I think it is past time for Python 2.7 to retire (and I am tired of supporting it). Some Python GUI projects I've already dropped Python 2.7 just due to being tired of dealing with Unicode hacks for narrow vs wide characters and the very different Unicode handling that Python 2.7 generally does.

It doesn't matter to me if Python Markdown pledges or doesn't, but I personally lean towards pledging.

I personally plan on dropping Python 2.7 for all of my projects around EOL. As far as Python 3.X termination, I usually drop them when Travis complains as I am more indifferent to most of the 3.x EOL; the impact of these is far less significant.

EDIT: Thinking about this more, I don't care about the pledge site itself as much as I care about the idea behind the pledge.

waylan:
> I think it is past time for Python 2.7 to retire (and I am tired of supporting it)

[...]

PR #865, opened 2019-10-23
Drop support for legacy Python 2.7

hugovk:
For #760

### DIFF
--- a/markdown/__init__.py
+++ b/markdown/__init__.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 """
 Python Markdown
 
@@ -20,8 +19,6 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
-from __future__ import absolute_import
-from __future__ import unicode_literals
 from .core import Markdown, markdown, markdownFromFile
 from .util import PY37
 from .pep562 import Pep562
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -119,7 +119,7 @@ class Markdown:
 
         """
         for ext in extensions:
-            if isinstance(ext, util.string_type):
+            if isinstance(ext, str):
                 ext = self.build_extension(ext, configs.get(ext, {}))
             if isinstance(ext, Extension):
                 ext._extendMarkdown(self)
@@ -218,7 +218,7 @@ class Markdown:
 
     def is_block_level(self, tag):
         """Check if the tag is a block level HTML tag."""
-        if isinstance(tag, util.string_type):
+        if isinstance(tag, str):
             return tag.lower().rstrip('/') in self.block_level_elements
         # Some ElementTree tags are not strings, so return False.
         return False
@@ -250,7 +250,7 @@ class Markdown:
             return ''  # a blank unicode string
 
         try:
-            source = util.text_type(source)
+            source = str(source)
         except UnicodeDecodeError as e:  # pragma: no cover
             # Customise error message while maintaining original trackback
             e.reason += '. -- Note: Markdown only accepts unicode input!'
@@ -318,7 +318,7 @@ class Markdown:
 
         # Read the source
         if input:
-            if isinstance(input, util.string_type):
+            if isinstance(input, str):
                 input_file = codecs.open(input, mode="r", encoding=encoding)
             else:
                 input_file = codecs.getreader(encoding)(input)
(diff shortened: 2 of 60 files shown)


markdown/util.py
```diff
@@ -20,32 +19,15 @@
...
-=============================================================================
-"""
-PY3 = sys.version_info[0] == 3
 PY37 = (3, 7) <= sys.version_info
 
-if PY3:  # pragma: no cover
-    string_type = str
-    text_type = str
-    int2str = chr
-    iterrange = range
```
waylan, line 42: It is entirely possible that third party extensions are importing these variables. Perhaps we should leave them defined, even though we don't use them. We might even consider wrapping them with `deprecated` and issue a DeprecationWarning on use. Of course, in a future release they would then be removed.

markdown/__init__.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
waylan, line 1: It appears that this was removed from all files. Why? This is not related to Python version at all and is still a valid comment to have for people who use certain editors.

markdown/core.py
```diff
@@ -321,15 +318,15 @@ def convertFile(self, input=None, output=None, encoding=None):
 
         # Read the source
         if input:
-            if isinstance(input, util.string_type):
+            if isinstance(input, str):
                 input_file = codecs.open(input, mode="r", encoding=encoding)
```
waylan, line 322: This code makes heavy use of `codecs.open` because it was developed prior to the `io` lib existing. It should probably be reworked to use the default `open` builtin. That said, it might make sense to do that as a separate PR.

waylan (changes requested):
Wow. That was a lot of work. Thanks. I took a rather quick look at the changes and have noted a few minor issues below. Get those addressed and I'll take a closer look when I have more time.

markdown/__init__.py
waylan: So I took another look at this and noticed that this change was part of commit c62d978, which was generated by `pyupgrade`.  The [documentation][1] for `pyupgrade` points to [PEP3120][2] as justification for removing the comments. I had mis-remembered why we needed those comments (I was incorrectly thinking the comment was for editors, not the Python compiler). Given the above, I am fine with leaving the change as-is. 

[1]: https://github.com/asottile/pyupgrade#-coding--comment
[2]: https://www.python.org/dev/peps/pep-3120/

tests/test_apis.py
```diff
@@ -442,12 +434,6 @@ def tearDown(self):
         # Reset warning behavior back to default
         warnings.simplefilter('default')
 
-    def testNonUnicodeSource(self):
-        """ Test falure on non-unicode source text. """
-        if not PY3:
-            source = "foo".encode('utf-16')
-            self.assertRaises(UnicodeDecodeError, markdown.markdown, source)
-
```
waylan: Let's leave this test in. It is entirely reasonable to expect that a user could pass in something other than `utf-8` source text and we should continue to return an appropriate error. I haven't checked this, but if the test was removed because it was failing, then we need to make some changes to the code to make it pass again.

waylan:
I did a search for `TODO` comments in the code and found one comment relating to dropping PY2 at [markdown/test_tools.py#L150-L156][1]:

```python
# Define LegacyTestCase class with metaclass in Py2 & Py3 compatible way.
# See https://stackoverflow.com/a/38668373/866026
# TODO: If/when py2 support is dropped change to:
# class LegacyTestCase(unittest.Testcase, metaclass=LegacyTestMeta)


class LegacyTestCase(LegacyTestMeta('LegacyTestCase', (unittest.TestCase,), {'__slots__': ()})):
```

We should make the change anticipated in that comment (and remove the comment) as part of this PR.

It does not appear that any other `TODO` comments are related to dropping PY2. However, if anyone finds any, feel free to point them out.

[1]: https://github.com/Python-Markdown/markdown/blob/dbdb41469e4101c0fdf4b30f69be35d9cd801430/markdown/test_tools.py#L150-L156

waylan:
Also, a comment needs to be added to the release notes.

markdown/util.py
hugovk: I've reverted them in e51e15b3f518ac62d1e99223eb58a5c6b65ebae5 and added a deprecation comment for now.

markdown/core.py
hugovk: Yes, I think a separate PR would be sensible.

hugovk:
Thanks, updated!

waylan (approved):

waylan:
Not sure why the latest [Travis build][1] isn't showing up, but it is passing.

@hugovk note the changes I made in cfa1aa6 which marks the variables as deprecated. Conveniently, we already have a vendored backport of `pep562` included, so it was easy to make use of it.

[1]: https://travis-ci.org/Python-Markdown/markdown/builds/602322487?utm_medium=notification&utm_source=github_status

hugovk:
Great, thanks for the finishing touches!

PR #866, opened 2019-10-23
Proposal: Add support for definition terms <dt> in TOC

pdaoust:
This is a need that showed up for me recently. In a glossary document, it's nice to have definition terms show up in the table of contents and have those permalinks and anchors added. This permits deep-linking into glossaries from other resources.

By default, this functionality is turned off -- the default depth is still 6, but `<dt>` elements get an imaginary hierarchy depth of 7. Turning this functionality on is as simple as specifying `toc_depth: 7`. The assumption here is that, if people don't want `h6` elements in their hierarchy, they probably won't want `<dt>`s in there either.

If this feature seems useful enough to warrant inclusion, I request a pair of eyes to look over two concerns (will create a review to comment on the specific lines).

Apologies for the big diff; if I didn't had to de-indent a bunch of lines, the diff would be much smaller.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -134,7 +134,8 @@ class TocTreeprocessor(Treeprocessor):
         self.use_permalinks = parseBoolValue(config["permalink"], False)
         if self.use_permalinks is None:
             self.use_permalinks = config["permalink"]
-        self.header_rgx = re.compile("[Hh][123456]")
+        self.header_rgx = re.compile("h[123456]", re.IGNORECASE)
+        self.dt_rgx = re.compile("dt", re.IGNORECASE)
         if isinstance(config["toc_depth"], string_type) and '-' in config["toc_depth"]:
             self.toc_top, self.toc_bottom = [int(x) for x in config["toc_depth"].split('-')]
         else:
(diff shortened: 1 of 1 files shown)


markdown/extensions/toc.py
```diff
@@ -237,31 +238,47 @@ def run(self, doc):
...
-                if self.use_anchors:
-                    self.add_anchor(el, el.attrib["id"])
-                if self.use_permalinks:
-                    self.add_permalink(el, el.attrib["id"])
+                # Then get its level.
+                level = int(el.tag[-1])
+            else if self.dt_rgx.match(el.tag):
+                # Definition terms are outside of the header hierarchy,
+                # so let's give them a 7.
+                level = 7 + self.base_level
```
pdaoust, line 253: I don't know if the `+ self.base_level` is even necessary here; I don't quite understand the base level logic.

pdaoust:
Also, I noticed that near the end ([line 323/343](https://github.com/Python-Markdown/markdown/pull/866/files#diff-a67e43616df2478aae8e2e2c57f70fdfL323)) it's putting itself after the headerid extension. It would probably also be good to put it after the 'extras' extension which has support for MarkdownExtra definition lists. On my machine it always runs in the right order though.

waylan:
I'm inclined to not add this feature. We have had TOC support for many years and this is the first request we have received for the feature. Additionally, I am not aware of any other Markdown implementation which offers such a feature. It would seem this is not a highly sought after feature, which is very much non-standard, even among non-standard features.

That said, this would be great for a third party extension. Whether you fork the TOC extension (within the license terms) or create a separate new extension is up to you. And please add any new extension to the [wiki][1] if/when it is released.

[1]: https://github.com/waylan/Python-Markdown/wiki/Third-Party-Extensions

pdaoust:
Righto; thanks for the consideration and further guidance. Didn't know if it was useful enough to others to warrant inclusion; just thought, "hey, I've done the work; if it's useful to others then yay" :)

PR #877, opened 2019-11-11
Added permalink_title option

### LINKED ISSUE #781: TOC: Configurable permalink title attribute
Would be nice to have an ability to set a custom title text for permalink title attribute. For example:

```yml
- toc:
    title: 'Ссылка на этот раздел' #Custom title here
    permalink: '#'
    slugify: !!python/name:pymdownx.slugs.uslugify
```

For now, "Permament link" text is simply hardcoded within `toc.py`:
https://github.com/Python-Markdown/markdown/blob/9ec83c4c414f27bc8aa490dbac1fe30b49058ae6/markdown/extensions/toc.py#L194

Related: #657

wilhelmer:
Fixes #781.

### DIFF
--- a/docs/change_log/release-3.2.md
+++ b/docs/change_log/release-3.2.md
@@ -58,6 +58,9 @@ The following new features have been included in the release:
 
 * Add support for Python 3.8.
 
+* A new `permalink_title` parameter has been added to the
+  [Table of Contents Extension](../extensions/toc.md).
+
 ## Bug fixes
 
 The following bug fixes are included in the 3.2 release:
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -163,7 +163,7 @@ The following options are provided to configure the output:
     text.
 
 * **`permalink_title`**:
-    Title attribute of the permalink. Defaults to `Permanent link`.
+    Title attribute of the permanent link. Defaults to `Permanent link`.
 
 * **`baselevel`**:
     Base level for headers. Defaults to `1`.
--- a/docs/extensions/toc.md
+++ b/docs/extensions/toc.md
@@ -162,6 +162,9 @@ The following options are provided to configure the output:
     the link text. When set to a string, the provided string is used as the link
     text.
 
+* **`permalink_title`**:
+    Title attribute of the permalink. Defaults to `Permanent link`.
+
 * **`baselevel`**:
     Base level for headers. Defaults to `1`.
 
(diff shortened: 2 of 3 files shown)


waylan:
Look good. As noted in the [Contributing Guide][1], this needs tests and an update to the Release Notes. Add those and I'll merge it.

wilhelmer:
Concerning Release Notes - forgot those, will add that in a minute.

Concerning tests - from the Contributing Guide: 

> While it is generally fine to leave those tests for the Travis server to run when a pull request is submitted, for more advanced changes, you may want to run those tests locally.

I don't consider this as an advanced change, and the tests on the Travis server have all passed, so which other tests are needed?

waylan:
You need to add tests which confirm that the new option works as it is supposed to. Does it actually change the title when set? Does the default actually get used when not set?

I haven't checked, but the second question above might already get answered in the existing tests. If so, you would only need to add at least one new test. If not, then you would need to add at least two.

waylan:
Note that as per the tests policy, any new tests should go in [/tests/test_syntax/extensions][1]. As none of the TOC tests have been migrated to the new system, you'll need to add a new file `test_toc.py` there and create your tests. At a minimum, you should create the two tests which confirm the default and a custom title get used.

Note that the existing tests for the TOC extension are in [tests/extensions][2] (in various files with names starting with `toc`) and the `TestTOC` class of [tests/test_extensions.py][3]. If any of those tests cover the same behavior, they should be converted to the new style tests and removed from here. I'm thinking perhaps the `testPermalink` tests in the `TestTOC` class. You don't have to covert any other tests, but are certainly welcome to.

[1]: https://github.com/Python-Markdown/markdown/tree/master/tests/test_syntax/extensions
[2]: https://github.com/Python-Markdown/markdown/tree/master/tests/extensions
[3]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_extensions.py#L696

wilhelmer:
Uh, I have zero experience in testing and have no clue what those files do. Maybe you could add those for me? Sorry :(

waylan:
That's fine. However, I have very little time to do it. So unless someone else wants to do the work it will likely be some time before I get to it. I am not going to merge this (or anything else) without the tests.

waylan:
As this was on a `master` branch, it was easier to just create a new PR than try to push changes to this. Therefore, this is being closed in favor of #886 which uses this as a base, resolves the conflicts and adds a test.

PR #901, opened 2020-01-31
Don't emit stashed HTML tag placeholders in `.toc_tokens`

### LINKED ISSUE #899: Heading names in toc_tokens contain stashed HTML placeholders
If a Markdown heading contains HTML, the corresponding entry in the `.toc_tokens` property ends up with HTML placeholders when returned to the user. The following example should illustrate the problem:

```
>>> import markdown
>>> md = markdown.Markdown(extensions=['toc'])
>>> md.convert('# <code>Heading</code>\n')
'<h1 id="heading"><code>Heading</code></h1>'
>>> md.toc_tokens
[{'level': 1, 'id': 'heading', 'name': '\x02wzxhzdk:0\x03Heading\x02wzxhzdk:1\x03', 'children': []}]
```

While this isn't *too* hard to fix (we could just un-stash the HTML immediately before returning it to the user), it does raise a bigger question: what should the data format of the `name` field in `toc_tokens` be? Is it...

* Markdown (so the value would be exactly as in the source file: `<code>Heading</code>`)
* Plain text (so the value would strip HTML: `Heading`)
* HTML (similar to the Markdown format, but with HTML entities replaced, so `<code>a>b</code>` becomes `<code>a&gt;b</code>`)

In particular, this is relevant for https://github.com/mkdocs/mkdocs/pull/1970. Prior to that PR, MkDocs would build an internal representation of the TOC by parsing the HTML from `.toc`. With the change, it (tries to) use `.toc_tokens`, but fails due to this issue.

FWIW, I think MkDocs wants this to be plain text in the end, but HTML makes the most sense to me in general: after all, the purpose of this lib is to convert Markdown to HTML. (MkDocs would then just need to parse the HTML fragment and strip out the tags.)

jimporter:
This should resolve #899. The fix is a bit subtle, since we want to emit HTML entities for `.toc_tokens`, but not HTML tags. We also want to be careful to clean up `data-toc-label` as needed. Note: this change means that raw HTML in a heading is no longer passed through to the HTML for `.toc`.

I'm not *entirely* happy with this. In practice, I think the ideal would be for the toc extension to include the (HTML-ized) Markdown from a heading in its TOC entry, and to have `.toc_tokens` include both `html` and `plaintext` fields. This patch gives us a sort of middle ground, where we just use the HTML with tags stripped. The "ideal" is a larger behavioral change though, and I'm not sure it makes sense to do that for a minor release...

If you think it makes sense to implement the "ideal" solution now, I can try and do that. However, this patch is sufficient for what MkDocs needs.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -43,7 +43,7 @@ def unique(id, ids):
     return id
 
 
-def stashedHTML2text(text, md):
+def stashedHTML2text(text, md, strip_entities=True):
     """ Extract raw HTML from stash, reduce to plain text and swap with placeholder. """
     def _html_sub(m):
         """ Substitute raw html with plain text. """
@@ -51,8 +51,11 @@ def stashedHTML2text(text, md):
             raw = md.htmlStash.rawHtmlBlocks[int(m.group(1))]
         except (IndexError, TypeError):  # pragma: no cover
             return m.group(0)
-        # Strip out tags and entities - leaveing text
-        return re.sub(r'(<[^>]+>)|(&[\#a-zA-Z0-9]+;)', '', raw)
+        # Strip out tags and/or entities - leaving text
+        res = re.sub(r'(<[^>]+>)', '', raw)
+        if strip_entities:
+            res = re.sub(r'(&[\#a-zA-Z0-9]+;)', '', res)
+        return res
 
     return HTML_PLACEHOLDER_RE.sub(_html_sub, text)
 
@@ -259,7 +262,9 @@ class TocTreeprocessor(Treeprocessor):
                 toc_tokens.append({
                     'level': int(el.tag[-1]),
                     'id': el.attrib["id"],
-                    'name': el.attrib.get('data-toc-label', text)
+                    'name': unescape(stashedHTML2text(
+                        el.attrib.get('data-toc-label', text), self.md, strip_entities=False
+                    ))
                 })
 
                 # Remove the data-toc-label attribute as it is no longer needed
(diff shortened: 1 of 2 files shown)


jimporter:
Hmm, one problem with the "ideal" solution mentioned above is that we'd probably want to strip `<a>` tags from the heading: those are obviously going to cause problems with the links generated for navigating from the TOC itself. Given that, maybe what I have here is the best solution for now...

jimporter:
One other option here would be that we strip HTML tags from the heading, but if a user really wants HTML tags in their heading, they could explicitly use `{ data-toc-label="Foo <code>Bar</code>" }`. This is more flexible, though attr lists do seem to have some unusual behavior when trying to embed Markdown in them, e.g. `{ data-toc-label="Foo *Bar*" }`. I haven't spent enough time digging through what's happening to understand how to fix that, but I assume it's because attr lists are just looking at the raw text, not the HTML elements generated by the initial Markdown parse.

waylan:
For `data-toc-label` to support raw HTML, then the content would need to be stashed in the html stash. Otherwise, it would get escaped when the elementtree is serialized.

For `data-toc-label` to support Markdown syntax, it gets complicated. For example, we do not want `a` elements. And what about images or block level elements? I suppose we could allow a whitelist of known specific inline elements only. I would need to explore this more as well. But that is separate from the issue we are trying to address here.

While the fix here is more restrictive than previously, it fixes the obviously wrong output and actually does what we intended to do previously. If we want to add support for formatting TOC entries, then that would be a separate new feature.

PR #902, opened 2020-02-01
Simplify xml.etree.ElementTree loading

mitya57:
cElementTree [is a deprecated alias for ElementTree][1] since Python 3.3. As we are now supporting only Python ≥ 3.5, we do not need to use it.

The version check is also no longer needed. The module has [version 1.3.0][2] since 2010, so even Python 2.7 and 3.2 have it (see python/cpython@f15351d938e76c4b).

[1]: https://docs.python.org/3/library/xml.etree.elementtree.html#module-xml.etree.ElementTree
[2]: https://bugs.python.org/issue6472

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -24,6 +24,7 @@ import sys
 from collections import namedtuple
 from functools import wraps
 import warnings
+import xml.etree.ElementTree
 from .pep562 import Pep562
 
 
@@ -32,6 +33,7 @@ PY37 = (3, 7) <= sys.version_info
 
 # TODO: Remove deprecated variables in a future release.
 __deprecated__ = {
+    'etree': ('xml.etree.ElementTree', xml.etree.ElementTree),
     'string_type': ('str', str),
     'text_type': ('str', str),
     'int2str': ('chr', chr),
@@ -81,23 +83,6 @@ RTL_BIDI_RANGES = (
     # Thaana (0780-07BF), Nko (07C0-07FF).
     ('\u2D30', '\u2D7F')  # Tifinagh
 )
-
-# Extensions should use "markdown.util.etree" instead of "etree" (or do `from
-# markdown.util import etree`).  Do not import it by yourself.
-
-try:  # pragma: no cover
-    # Is the C implementation of ElementTree available?
-    import xml.etree.cElementTree as etree
-    from xml.etree.ElementTree import Comment
-    # Serializers (including ours) test with non-c Comment
-    etree.test_comment = Comment
-    if etree.VERSION < "1.0.5":
-        raise RuntimeError("cElementTree version 1.0.5 or higher is required.")
-except (ImportError, RuntimeError):  # pragma: no cover
-    # Use the Python implementation of ElementTree?
-    import xml.etree.ElementTree as etree
-    if etree.VERSION < "1.1":
-        raise RuntimeError("ElementTree version 1.1 or higher is required")
 
 
 """
--- a/docs/change_log/release-3.2.md
+++ b/docs/change_log/release-3.2.md
@@ -49,15 +49,19 @@ Internally, this change relies on the Pygments 2.4, so you must be using at leas
 that version to see this effect. Users with earlier Pygments versions will
 continue to see the old behavior.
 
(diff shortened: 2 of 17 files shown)


markdown/util.py
```diff
@@ -85,19 +85,7 @@
 # Extensions should use "markdown.util.etree" instead of "etree" (or do `from
 # markdown.util import etree`).  Do not import it by yourself.
 
-try:  # pragma: no cover
-    # Is the C implementation of ElementTree available?
-    import xml.etree.cElementTree as etree
-    from xml.etree.ElementTree import Comment
-    # Serializers (including ours) test with non-c Comment
-    etree.test_comment = Comment
```
waylan, line 93: The [failing test][1] is caused by removing this line. The fix would be to update both the [tests][2] and the [serializer][3] to both use `from xml.etree.ElementTree.Comment` without the extra custom `test_comment` object.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/644817928#L247
[2]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_apis.py#L472-L515
[3]: https://github.com/Python-Markdown/markdown/blob/master/markdown/serializers.py#L45-L48

markdown/util.py
```diff
@@ -85,19 +85,7 @@
 # Extensions should use "markdown.util.etree" instead of "etree" (or do `from
 # markdown.util import etree`).  Do not import it by yourself.
```
waylan: These comments should be rewritten as they are no longer correct. Instead, we would now be importing `etree` here for backward compatibility with old third party extensions which used to rely on our custom import.

However, as that custom import is no longer necessary, the `markdown` library itself should not use this import at `markdown.util` but simply import `tree` directly. Perhaps we could even raise a deprecation warning if third party code uses this import.

markdown/util.py
```diff
@@ -85,19 +85,7 @@
...
-    # Serializers (including ours) test with non-c Comment
-    etree.test_comment = Comment
-    if etree.VERSION < "1.0.5":
-        raise RuntimeError("cElementTree version 1.0.5 or higher is required.")
-except (ImportError, RuntimeError):  # pragma: no cover
-    # Use the Python implementation of ElementTree?
-    import xml.etree.ElementTree as etree
-    if etree.VERSION < "1.1":
-        raise RuntimeError("ElementTree version 1.1 or higher is required")
+import xml.etree.ElementTree as etree
```
waylan: This is getting [two flake8 errors][1]. Given the situation, those errors are acceptable. Therefore a comment needs to be added to this line to tell Flake8 to ignore the line: `# noqa`.

[1]: https://travis-ci.org/Python-Markdown/markdown/jobs/644817931#L242

waylan (changes requested):
Simplifying the code by removing old cruft like this is great. However, the cruft has only been half removed. See my comments below for what need sot happen before this can be accepted.

mitya57:
Now that we deprecated this, I had to update a lot of internal code. Commit 0f48d17e deprecates the old `markdown.util.etree` wrapper, commits 80729a3a and 8b1f49a1 make the replacements, and commit 3b7a55be updates the documentation.

waylan (commented):
This looks good. However, one thing is bothering me. Sometimes we import `as etree` and other times we don't do the `as etree` bit. Lets be consistent throughout the lib.

mitya57:
Ok. I decided to go with the `as etree` form as it is shorter and requires less changes throughout the code.

waylan (approved):

waylan:
We should probably add a note to the release notes informing extension devs that `markdown.util.etree` is deprecated and that they should import `xml.etree.ElementTree` directly. With that small change, I think this is ready to go. Nice work.

mitya57:
Done in 01ca7399. Please check if the wording is fine.

waylan:
> Please check if the wording is fine.

It was pretty good. I made a few adjustments to the first sentence for consistency ("`some.module.path` module" requires the article "the" but if you leave out the word "module" then you also should leave out the article). The other changes were mostly additions which I probably wouldn't have bothered with if I wasn't already making changes.

PR #907, opened 2020-02-07
Fix escaping of HTML special chars (<, >, &) in `.toc_tokens`

jimporter:
This should fix the problem I found in https://github.com/mkdocs/mkdocs/pull/1970. I think the tests cover all the necessary cases. (I'm a little surprised that escaping where I did works though; I guess the `etree.Element.text` setter used for making the HTML TOC accepts escaped *or* unescaped HTML special chars.)

Sorry I didn't catch this the first time around!

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -3,9 +3,10 @@ title: Change Log
 Python-Markdown Change Log
 =========================
 
-Under Development: Released version 3.2.1 (a bug-fix release). The `name`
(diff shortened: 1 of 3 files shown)


waylan:
Looks good. As per the [contributing guide][1], this just needs a comment in the [change log][2] under 3.2.1. 

[2]: https://github.com/Python-Markdown/markdown/blob/1de3083d82a5245015f0f0b9471a965916134d6d/docs/change_log/index.md

PR #928, opened 2020-03-28
Use simplified regex for html placeholders

raabrp:
`util` already provides the regex we need for all placeholders, and we're doing extra work parsing for the explicit union of all of them.
In documents with ~30000 placeholders, I get a ~5x speedup with this change.


waylan:
Can you add a note to the release notes?

raabrp:
Updated release notes for 3.2 under "bug fixes"

PR #933, opened 2020-04-05
Revert: Use simplified regex for html placeholders

### LINKED ISSUE #932: Last commit breaks a bunch of tests
This commit broke a bunch of our tests: https://github.com/Python-Markdown/markdown/commit/c5f1395a944e76e95250740452706f6b16830cae.

We should revert it. When reverting locally, tests once again pass.

facelessuser:
Not sure how tests didn't get run on this, but it was completely
breaking ~~all~~ a lot of tests. This reverts the change.

Fixes #932

### DIFF
--- a/docs/change_log/release-3.2.md
+++ b/docs/change_log/release-3.2.md
@@ -93,5 +93,4 @@ The following bug fixes are included in the 3.2 release:
 * HTML tag placeholders are no longer included in  `.toc_tokens` (#899).
 * Unescape backslash-escaped characters in TOC ids (#864).
 * Refactor bold and italic logic in order to solve complex nesting issues (#792).
-* Always wrap CodeHilite code in <code> tags (#862) </code>
-* Regex for HTML placeholders simplified for speed improvement (#928).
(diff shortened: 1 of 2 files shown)


facelessuser:
Spellchecker is finding `lt` in code blocks, that is why the spellchecker is failing. Not sure how you want us to handle this @waylan. Do I just add `lt` to the dictionary? Or can we exclude `code` tags?

waylan:
Weird. I distantly recall checking that the tests passed on that because I expected them to fail. When the weren't failing, I accepted the change. But now, looking back, I don't see any indication that any tests even ran. Not sure what happened there. In any event we're getting 60 failing tests so this is the right move.

Also I just noticed that the note was added to the release notes for a previous release rather than a future release. Guess I wasn't paying close attention that day.

> Spellchecker is finding `lt` in code blocks,

Code blocks are supposed to be skipped already. Not sure why that isn't happening.

facelessuser:
@waylan #934 fixes it, it was finding the `lt` in the TOC.

PR #934, opened 2020-04-06
TOC fix for AtomicString handling

### LINKED ISSUE #931: Less/greater than signs in headings appear as &lt/&gt in TOC
screen

That's an mkdocs site with the readthedocs theme. It's using git master HEAD of mkdocs and markdown. The issue was caused by  https://github.com/Python-Markdown/markdown/commit/74c0166c6903b07b3b2b9cbba36705d1414cbaf7. I haven't tested this using just markdown so it might be related to mkdocs.

facelessuser:
Fixes #931

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -3,7 +3,11 @@ title: Change Log
 Python-Markdown Change Log
 =========================
 
-Feb 12, 2020: Released version 3.2.1 (a bug-fix release). 
+Released version 3.2.2 (a bug-fix release).
+
+* Fixed issue where double escaped entities could end up in TOC.
+
+Feb 12, 2020: Released version 3.2.1 (a bug-fix release).
 
 * The `name` property in `toc_tokens` from the TOC extension now
   escapes HTML special characters (`<`, `>`, and `&`).
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -3,7 +3,7 @@ title: Change Log
 Python-Markdown Change Log
 =========================
 
-Released version 3.2.2 (a bug-fix release).
+Under development: version 3.2.2 (a bug-fix release).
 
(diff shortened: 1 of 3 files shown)


docs/change_log/release-3.2.md
```diff
@@ -93,5 +93,4 @@ The following bug fixes are included in the 3.2 release:
 * HTML tag placeholders are no longer included in  `.toc_tokens` (#899).
 * Unescape backslash-escaped characters in TOC ids (#864).
 * Refactor bold and italic logic in order to solve complex nesting issues (#792).
-* Always wrap CodeHilite code in <code> tags (#862) </code>
-* Regex for HTML placeholders simplified for speed improvement (#928).
+* Always wrap CodeHilite code in `code` tags (#862)
```
waylan: We already released version 3.2. Therefore, release notes for a future release would not go here. Assuming we do a bug fix release, they would go in the changelog right before the 3.2.1 release (and in the same format):

https://github.com/Python-Markdown/markdown/blob/c5f1395a944e76e95250740452706f6b16830cae/docs/change_log/index.md#L6-L9

Of course the date would be left blank until we actually do a release.

waylan (changes requested):

docs/change_log/release-3.2.md
facelessuser: This was just reverting the last commits addition here, but I can add a comment in the changelog if that is what you are requesting. This is based off the reverting pull request I just made.

waylan: Yeah, we should add a comment as we are fixing a bug introduced in a previous release.

facelessuser: Rebased and added changelog entry.

waylan (approved):

PR #939, opened 2020-04-14
extend_markdown() TypeErrors swallowed because of deprecation test.

merriam:
Currently, an extension's `extendMarkdown()` raising a TypeError will be swallowed, or not reported.  This occurs because the calling `_extendMarkdown()` assumes the TypeError occurs because an `md_globals` argument was also passed.  This pull request modifies `_extendMarkdown()` to check this assumption.

Also, trying to delete an invalid key is a KeyError as opposed to a TypeError.

Have a wonderful day!

### DIFF
--- a/docs/change_log/release-3.2.md
+++ b/docs/change_log/release-3.2.md
@@ -94,3 +94,5 @@ The following bug fixes are included in the 3.2 release:
 * Unescape backslash-escaped characters in TOC ids (#864).
 * Refactor bold and italic logic in order to solve complex nesting issues (#792).
 * Always wrap CodeHilite code in `code` tags (#862)
+* `_extend_markdown()` now correctly reports if an extension's `extend_markdown()` raises a TypeError (#939).
+* Attempting to delete a nonexistent key from the patterns registry now raises KeyError instead of TypeError (#939). 
--- a/docs/change_log/release-3.2.md
+++ b/docs/change_log/release-3.2.md
@@ -94,5 +94,3 @@ The following bug fixes are included in the 3.2 release:
 * Unescape backslash-escaped characters in TOC ids (#864).
 * Refactor bold and italic logic in order to solve complex nesting issues (#792).
 * Always wrap CodeHilite code in `code` tags (#862)
-* `_extend_markdown()` now correctly reports if an extension's `extend_markdown()` raises a TypeError (#939).
-* Attempting to delete a nonexistent key from the patterns registry now raises KeyError instead of TypeError (#939). 
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -6,7 +6,10 @@ Python-Markdown Change Log
 Under development: version 3.2.2 (a bug-fix release).
 
 * Load entry_points (for extensions) only once using `importlib.metadata`.
-* Fixed issue where double escaped entities could end up in TOC.
+* Do not double escape entities in TOC.
+* Correctly report if an extension raises a `TypeError` (#939).
+* Raise a `KeyError` when attempting to delete a nonexistent key from the
+  extension registry (#939). 
 
 Feb 12, 2020: Released version 3.2.1 (a bug-fix release).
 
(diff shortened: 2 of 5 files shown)


mitya57:
Tests in `tests/test_apis.py` should be changed to expect KeyError instead of TypeError.

I hope none of third-party extensions relied on this wrong exception type…

merriam:
Updated unit test

merriam:
While nothing in the repository uses this error outside of unit tests, it cannot be proven that some project, somewhere, relies upon it.  It would be a stretch to imagine why.

That said, I found a bug.  Fixed it.  Went back and fixed the buggy unit tests.   I'm OK with it never being adopted but not OK writing more makework tests.  It's a bad practice, and, if we ever meet for beers, I can give twenty years of research on it.

Other fish to fry...

waylan:
This should get a comment in the [release notes][1] under version 3.2.2.

[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/index.md

mitya57 (approved):

docs/change_log/release-3.2.md
```diff
@@ -94,3 +94,5 @@ The following bug fixes are included in the 3.2 release:
 * Unescape backslash-escaped characters in TOC ids (#864).
 * Refactor bold and italic logic in order to solve complex nesting issues (#792).
 * Always wrap CodeHilite code in `code` tags (#862)
+* `_extend_markdown()` now correctly reports if an extension's `extend_markdown()` raises a TypeError (#939).
+* Attempting to delete a nonexistent key from the patterns registry now raises KeyError instead of TypeError (#939).
```
waylan: Version 3.2 has already been released. Future changes should not be included in the release notes for that version. Instead, these notes should be added to the changelog here:

https://github.com/Python-Markdown/markdown/blob/ada40c6619fd0be740af646b043937e560716c95/docs/change_log/index.md#L6-L8

Also, `TypeError` and `KeyError` are causing the [spellchecking][1] test to fail. As they are Python objects, they should be in code spans.

Finally, the [Contributing Guide][2] states:

> Use the present tense (“Add feature” not “Added feature”).
> 
> Use the imperative mood (“Move item to…” not “Moves item to…”).

Generally, the same guidelines should apply to release notes.

[1]: https://travis-ci.org/github/Python-Markdown/markdown/jobs/676115628#L312
[2]: https://python-markdown.github.io/contributing/#commit-message-style-guide

merriam:
Er, Travis fails because it doesn't know KeyError and TypeError spelling?

mitya57:
@merriam As Waylan noticed, you need to put the class names into backticks: `KeyError` → `` `KeyError` ``.

waylan:
> Er, Travis fails because it doesn't know KeyError and TypeError spelling?

You don't expect KeyError and TypeError to be in the English dictionary do you? Sure, they are known Python words, but then they are code and the spellchecker ignores code. So, put then in code spans where they belong.

Tests accomplish two purposes: (1) ensure we don't break our code and (2) ensure we follow our own style guide. This accomplishes the later.

merriam:
Sorry, I'm giving up.   

The code changes took about 20 minutes.  Getting it accepted has taken several hours, and I have no idea how many more steps will appear.   It may a joy for some, though my way is more on the development side.

The current state is that any TypeError thrown by an `extendMarkdown()` will cause a secondary error when it is called again with too many parameters.

waylan:
Our process and requirements for PRs are all documented in our [Contributing Guide][1]. If we were to ignore those requirements our code and documentation would be a mess. Not sure how that would be better for anyone.

[1]: https://python-markdown.github.io/contributing/

waylan:
Oh and I would have happily made a few edits to fix the "spelling errors." However, you put the notes in the wrong file. To fix them I need to clone your fork locally, make the changes and push the changes. If you had put them in the correct location, I could have simply used GitHub's web interface to add a few backticks. I even linked to the correct location when I asked for the release notes.

Issue #897, opened 2020-01-24
HR preceeded by ***word*** doesn't render as HR


kutetapolu:
Input:

```
***test***
***
```

Output (second line will be rendered as a literal `***`):

```html
<p><strong><em>test</em></strong>
***</p>
```

Expected output according to Markdown.pl:

```html
<p><strong><em>test</em></strong>
</p>
<hr />
```

The same problem occurs with `---` and `___` as HR.

I think it can be fixed by changing `RE` in `HRProcessor`:

```diff
497c497
<     RE = r'^[ ]{0,3}((-+[ ]{0,2}){3,}|(_+[ ]{0,2}){3,}|(\*+[ ]{0,2}){3,})[ ]*'
---
>     RE = r'^[ ]{0,3}((-+[ ]{0,2}){3,}|(_+[ ]{0,2}){3,}|(\*+[ ]{0,2}){3,})[ ]*$'
```

waylan:
Here's what [Babelmark][1] shows. Looks like a bug.

[1]: https://johnmacfarlane.net/babelmark2/?normalize=1&text=***test***%0A***

PR #946, opened 2020-04-20
Extension Api doc

merriam:
Updating the API docs as I write an extension.   

* Fixed links
* Rewrote opening paragraph

Work in progress


docs/extensions/api.md
```diff
@@ -3,24 +3,31 @@ title:      Extensions API
...
+* [*Postprocessors*](#postprocessors) munge of the output of the parser just before it is returned. 
+
+The parser loads text, applies the preprocessors, creates an
+[ElementTree][ElementTree] added to by the block processors and 
+inline patterns, renders the ElementTree as Unicode text, and then 
+then applies the postprocessors.
+
+There are classes and helpers provided to ease writing your extension. 
+Each part of the API is discussed in its respective section below.
+Additionally, you can walk through the [Tutorial on Writing Extensions][tutorial];
```
waylan: The tutorial is out-of-date. It was never updated for version 3.0 and won't work as-is. While I agree that linking to the tutorial here would be ideal, I think the tutorial needs to be updated first.

waylan:
Looks great so far. I added the [work in progress] label to this. When you are finished, feel free to remove the label.

docs/extensions/api.md
merriam: The tutorial is up-to-date.   I updated it.

waylan: awesome. Thanks

merriam:
Hmm.. The spell checker is causing more problems than it is fixing.   It flagged 'encodings' and cannot see that table entries should not be checked.

docs/extensions/api.md
```diff
@@ -26,34 +28,51 @@ other issues on the [bug tracker].
 
 ## Preprocessors {: #preprocessors }
 
-Preprocessors munge the source text before it is passed into the Markdown
-parser. This is an excellent place to clean up bad syntax, extract things the
-parser may otherwise choke on and perhaps even store it for later retrieval.
+Preprocessors munge the source text before it is passed to the Markdown
+parser. This is an excellent place to clean up bad encodings or to extract
```
waylan: We definitely want "syntax" not encodings" here. Markdown only accepts Unicode text, therefore any encodings will have been lost by time time preprocessors get the text. Yes, I realize you might be speaking of something else, but I don't want anyone to misunderstand.

docs/extensions/api.md
```diff
@@ -26,34 +28,51 @@ other issues on the [bug tracker].
...
+class NoRender(Preprocessor):
+    """ Skip any line with words 'NO RENDER' in it. """
     def run(self, lines):
         new_lines = []
         for line in lines:
-            m = MYREGEX.match(line)
+            m = re.search("NO RENDER", line)
             if m:
-                # do stuff
+                pass   # skipping this line
```
waylan: To those of us who are well versed in Python, this change is a non-issue. But to newbies, `pass` is confusing. This doesn't improve things at all. I realize that means the code won't run as-is, but that is why we called it "pseudo code" above. Let's keep it that way.

docs/extensions/api.md
```diff
@@ -26,34 +28,51 @@ other issues on the [bug tracker].
...
+Some preprocessors in the Markdown source tree include:
+
+
+| Name  | kind | Priority |  Description |
+| ----------------------------|-----------|----|-----------------------------------------------
+| [NormalizeWhiteSpace][c1]   | built-in  | 30 | Normalizes whitespace by expanding tabs, fixing `\r` line endings, etc. |
+| [HtmlBlockPreprocessor][c2] | built-in  | 20 | Removes html blocks from the text and stores them for later processing |
+| [ReferencePreprocessor][c3] | built-in  | 10 | Removes reference definitions from text and stores for later processing |
+| [MetaPreprocessor][c4]      | extension | 27 | Strips and records meta data at top of documents |
+| [FootnotesPreprocessor][c5] | extension | 15 | Removes footnote blocks from the text and stores them for later processing |
```
waylan: This is interesting. I don't think I had considered doing this. The problem is that the links to the code are not for a specific blob. Therefore, as changes happen in the future, the line numbers will become wrong. At the same time, updating the blob id is not something we want to do either.

Perhaps instead we should be using [tomchristie/mkautodoc][1] to pull in the documentation from the source code for the Python objects. To be clear, we still want the prose we have now, but we can supplement it with the autodoc generated documentation for the Python objects.

[1]: https://github.com/tomchristie/mkautodoc

docs/extensions/api.md
```diff
@@ -807,8 +837,8 @@ module and call the `makeExtension` function to initiate your extension.
 [makeExtension]: #makeextension
 [ElementTree]: https://effbot.org/zone/element-index.htm
 [Available Extensions]: index.md
-[Footnotes]: https://github.com/Python-Markdown/mdx_footnotes
-[Definition Lists]: https://github.com/Python-Markdown/mdx_definition_lists
+[Footnotes]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/footnotes.py
+[Definition Lists]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/definition_lists
```
waylan, line 864: Why were these links changed? They were linking to the documentation for the extensions. That's what we want.

I can see the benefit of helping people find the source code, but that should be a secondary link. Perhaps something like: `... [Footnotes Extension][footnotes] (see [source code][footnotes_src])...`

mitya57: The old links are 404…

waylan:
If you find a legitimate word which is not in the spelling dictionary, you can add it to [.spell-dict][1] (one word per line). However, I find that is rarely necessary. Usually the problem is that an abbreviated form is being used (when the full form should be used) or the word should be in a  code span.

[1]: https://github.com/Python-Markdown/markdown/blob/master/.spell-dict

docs/extensions/api.md
waylan: > The old links are 404…

Fair point. I was going to say that they used to point to documentation, but looking at them again, I don't know what they pointed to. I don't think those URL's were ever valid. I'd have to check the history to see what they pointed to.

docs/extensions/api.md
merriam: I respectfully disagree.

First, this is now code, instead of pseudo-code and does run if it is registered.  The example now does something; removing NO RENDER lines.  Second, `pass` is necessary here.   The `if` syntax requires  a statement  and a comment is not a statement.  Finally, it is reasonable to assume people on their first programming project are not writing Markdown extentions.

docs/extensions/api.md
merriam: Removed encodings; you are correct.  It now says 'bad characters' which is accurate.

merriam:
Well; it passes spelling and another check gets run. :)

I hope to get out of the house today and hike up a mountain; it may stay broken for some time.

docs/extensions/api.md
merriam: I expect if the links move around by a few lines people will still be able to find the class in question.  GitHub should do the fix, allowing something like `http.....thing.py?F-Class-NormalizeWhiteSpace` to search the blob for the first matching string.  

I'll go ahead and file a decent bug report on this.  The previous many times a bug was opened, there was no clear suggestion for GitHub to proceed.  GitHub is now getting the Microsoft mindset and yet-another-feature is likely to be looked on favorably.

facelessuser:
Coverage failure can be ignored as nothing you've touched is code.

docs/extensions/api.md
waylan: But it still needs to say "syntax." That was intentional.

docs/extensions/api.md
waylan: I am not persuaded. After years of dealing with this sort of thing, the best is what we had. Please change it back.

mitya57: You can also consider using string literals:
```python
if m:
    "do stuff"
```

This will be valid Python code, and maybe less confusing.

facelessuser: I'm not sure it has to be valid. It's really just a placeholder for real code. I think we are bikeshedding here. I'm personally indifferent. I'm not sure users are actually getting hung up on this.

facelessuser: I could understand fixing it if we were actually getting real world complaints.

docs/extensions/api.md
merriam: Hmm.... I'll still ask GitHub to put in the feature.  The people I know there left :(

When I'm done with this section, I'll also implement a post processor extension to allow links like `source:extensions/footnotes.py#F-Class-NormalizeWhiteSpace`, though it would use I/O access.

merriam:
OK.  This completes the Phases of Processing section.  I would like to integrate this now.  

There will always be more to write.

merriam:
Approve?

docs/extensions/api.md
```diff
@@ -2,425 +2,563 @@ title:      Extensions API
...
+## Phases of processing {: #stages }
+### Preprocessors {: #preprocessors }
+
+Preprocessors munge the source text before it is passed to the Markdown
+parser. This is an excellent place to clean up bad characters or to extract 
+portions for later processing that the parser may otherwise choke on.
+
+Preprocessors inherit from `markdown.preprocessors.Preprocessor` and
+implement a `run` method, which takes a single parameter `lines`. This parameter is
+the entire source text stored as a list of Unicode string, one per line.  `run` should
```
waylan: Missing an `s` here: "... a list of Unicode string<ins>s</ins>, one per line."

docs/extensions/api.md
```diff
@@ -2,425 +2,563 @@ title:      Extensions API
...
+return its processed list of of Unicode strings, one per line.
```
waylan: An extra `of` here.

docs/extensions/api.md
waylan: We still need to remove the `pass`. I'm not approving this otherwise.

docs/extensions/api.md
```diff
@@ -2,425 +2,563 @@ title:      Extensions API
...
+
+A block processor parses blocks of text and adds new elements to the `ElementTree`. 
+Blocks of text, separated from other text by blank lines, may have a different syntax 
+and produce a differently structured tree than
+other Markdown.  Block processors excel at code formatting, 
+equation layouts, and tables.  
+
+Block processors inherit from `markdown.blockprocessors.BlockProcessor`, 
+are passed `md.parser` on initialization,
+and implement both the `test` and `run` methods:
```
waylan: We have some weird line wrapping going on here.

waylan: Actually, there appears to be less-than-ideal wrapping throughout the document. It would be nice to have that all cleaned up.

docs/extensions/api.md
```diff
@@ -2,425 +2,563 @@ title:      Extensions API
...
+| [`HtmlBlockPreprocessor`][c2] | built-in  | Removes html blocks from the text and stores them for later processing |
+| [`ReferencePreprocessor`][c3] | built-in  |Removes reference definitions from text and stores for later processing |
+| [`MetaPreprocessor`][c4]      | extension |Strips and records meta data at top of documents |
+| [`FootnotesPreprocessor`][c5] | extension | Removes footnote blocks from the text and stores them for later processing |
+
+[c1]: https://github.com/Python-Markdown/markdown/blob/master/markdown/preprocessors.py#L62
+[c2]: https://github.com/Python-Markdown/markdown/blob/master/markdown/preprocessors.py#L74
+[c3]: https://github.com/Python-Markdown/markdown/blob/master/markdown/preprocessors.py#L339
+[c4]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/meta.py#L32
+[c5]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/footnotes.py#L205
```
waylan: I previously expressed concern about these tables pointing to specific lines in the source code. I remain uncomfortable with this. Yes, even with minor code changes, the line will be "close enough" but that is not good enough. If GitHub provided a means to point at the specific class or function by name and it adjusted the line number, that would be fine. But we don't want to move forward on on some feature which may or may not be implemented in the future. This is a no-go.

waylan (changes requested):
I started reviewing this, but with the order of section changed around, its very difficult to see what has changed and what has simply been moved. Also, the formatting of some sections seems to have been completely reworked (definition lists removed). I'll need to look at the before and after in their rendered forms to properly review this. And with my limited availability, that will be some time from now.

However, I have previously raised some concerns which seem to either have been ignored or brushed off as not important. I even raised them early on, but the rest of the document has been edited as if they were never mentioned. Sorry, but this is a no-go without those concerns addressed. 

Finally, In the early sections of the document, as I began my review, I pointed out every little issue (an extra word hear, a missing word there). As I got further in, I paid less attention to those things. At first my thinking was that If that all there was, when I was finished I would go through and make those changes myself. But, now I much bigger concerns that need addressed first.

docs/extensions/api.md
```diff
@@ -2,425 +2,563 @@ title:      Extensions API
...
+ElementTree. This is an excellent place for creating summaries, adding collected references, or last
+minute adjustments.
 
-# an oversimplified regex
-MYPATTERN = r'\*([^*]+)\*'
+A tree processor inherits from `markdown.treeprocessors.Treeprocessor`. Note it is `Treeprocessor` not `TreeProcessor`.
+A tree processor must implement a `run` method which takes 
+a single argument `root`, the current `ElementTree`.  It may either return `None`, probably modifying `root`, or an
+entirely new `ElementTree` to replace the current tree.  For efficiency, it is preferred to return `None` and 
+modify the `root` in place.
```
waylan: It appears that this needs to be reworked as per #949. Interestingly, all references to `ElementTree` have been put into code spans here. Previously they were not because they are not referring to the `ElementTree` class, but rather any object of the ElementTree library except for the `ElementTree` object (`Element`, `Comment`, etc.).

docs/extensions/api.md
waylan: So I took another look at this and it occurs to me that this could be a working code block without the `pass`. 

Instead of this:

```python
if m:
    pass   # skipping this line
else:
    new_lines.append(line)
```

Do this:

```python
if not m:
    new_lines.append(line)
```

In fact, that is better Python code. So we have two options:

1. Revert to the previous pseudo code
2. Update the example to as per above

docs/extensions/api.md
waylan: Perhaps as a half-solution, we could link to the file where the object is defined, but leave out the line number.

Another option might be to add a feature to the [mdx_gh_links][1] extension which converted a link to a function or class in a source file to something else. I'm not exactly sure how that would work, but that would be how we could implement our own solution.

[1]: https://github.com/Python-Markdown/github-links

merriam:
Before I put in more work, is the answer simply "this will never be approved?"  

There are many options on a project, and simply keeping it static, without changes is an option.  As this document has only had three small commits in the last few years, it may be a project in stasis.  

Are you expecting a path towards acceptance or are you choosing stasis.  While the choice of 'throw out the long document because new Python programmers might not understand `pass`' does seem like an indicator for stasis, an explicit response would clarify your position.

waylan:
From your last comment I get the impression that you expect us to just wholesale accept your large change. I realize that many projects tend to not scrutinize larger changes closely, but I am not that sort of person. A larger change should receive more scrutiny. In fact, your work is full of minor typos. It requires a lot of work before it will be accepted.

Given the large amount of time that will take and the fact that the few concerns which were previously raised were ignored, I am not inclined to spend the time to review if you are just going to ignore my review. However, if you are willing to take my review seriously and make the adjustments, then I am happy to work through a review with you.

In other words, if you address all of the concerns I have already raised, then I will take the time to continue my review and get this into an acceptable condition.

As an aside, you seem to be really stuck on the `pass` matter. I could get stuck on a lot of your changes as I wrote most of the original document, which you have completely ripped apart. But, if it is better for the users, then I can overlook that. However, in the `pass` case, past experience has proven that it is better to leave out `pass` (years ago I was making the same arguments you are today; turns out I was wrong). However, if you look at my followup comments you can see that in all other respects I have come around to using your example with on minor change (removing the unnecessary `pass`).

waylan:
I should add there there are some good changes in here. Regardless of what you do going forward @merriam, we could very well use much of this work. Its just that my time to spend on this is very limited right now. So it may be awhile.

docs/extensions/api.md
merriam: Good point.  I will change it shortly.

docs/extensions/api.md
merriam: How about I leave out the line references for now?  The examples will link to the correct file and an extension writer can scroll through the code.  A later PR could add fancy links and a post-processor for the doc building.

docs/extensions/api.md
merriam: I'll look at this; didn't know about #949.  I'll check consistency; usually I try to refer to just 'the tree' for the current, complete `ElementTree` object of the entire Markdown document.   Let me whack at it and get back to you.

docs/extensions/api.md
merriam: I'll try to clean some up; it's a manual process PyCharm.  The ragged edge is only visible to those editing the document and will not show when the documentation is rendered.  Be forgiving if some stays ragged. :)

merriam:
Thank you for your quick response.  I understand it is a lot to review and I will try to address each of your issues.  I will submit updates to this PR for points you bring up and keep things clean by not submitting any other PRs until this one is done (though #897 is on my short list).

docs/extensions/api.md
merriam: Thank you.  Done.

docs/extensions/api.md
merriam: Thank you.  Done.

docs/extensions/api.md
merriam: OK.  Leaving out the number for now.  Done.

docs/extensions/api.md
waylan: Our [Contributing Guide][2] is pretty clear that we want neat code and documentation. And yes, that means the source documentation. As the creator of Markdown [explains][1]:

> The overriding design goal for Markdown’s formatting syntax is to make it as readable as possible. The idea is that a Markdown-formatted document should be publishable as-is, as plain text...

Ragged text is not "as readable as possible." Always assume the document is being read as raw Markdown text.

That said, sometimes it is not possible due to a long URL or some other formatting. And yes, there have been a few times we let something messy slip in. However, in most cases, a good text editor should be able to clean up hard line breaks with a single command. 

[1]: https://daringfireball.net/projects/markdown/
[2]: https://python-markdown.github.io/contributing/

docs/extensions/api.md
waylan: That seems like a reasonable compromise.

waylan (approved):

docs/extensions/api.md
```diff
@@ -172,37 +172,34 @@ def test_block_processor():
...
-<p>Let's not.</p>
-
-<hr>
+!!! note ""
+    <p>A regular paragraph of text.</p>
+    <div style="display: inline-block; border: 1px solid red;">
+    <p>First paragraph of wrapped text.</p>
+    <p>Second Paragraph of **wrapped** text.</p>
+    </div>
+    <p>Another regular paragraph of text.</p>
```
waylan: Ideally these would be in blockquotes, but the theme we are using does not define any styles for blockquotes. Therefore, I wrapped them in `note` admonitions which have a blank label. The styling works out quite nicely:

Screen Shot 2020-05-07 at 1 48 28 PM

waylan (approved):
I just made a bunch of updates to this. I have mostly broken my changes up into commits by category. Because one of the changes included re-wrapping almost every line in the document, it may be best to review each commit separately to see what I did.

docs/extensions/api.md
waylan: And I just realized I commented on a different example than the one I screenshot. Anyway, the point remains.

waylan:
@merriam thanks for spearheading this. That was a lot of much-needed work.

PR #955, opened 2020-05-08
Add checklinks tox env and shell script.

waylan:
This is using the Node.js tool from [tcort/markdown-link-check][1] which runs on Markdown input.

Note that there are currently quite a few broken links in the documentation. We need to be checking these regularly. I'll be fixing them as I continue to work on this PR.

I should also note that this tool will properly verify MkDocs' internal links if they are proper relative links, but will error on internal absolute links. This fits MkDocs' requirements and ensures absolute links don't slip in. This provides an added benefit we might miss if we were using an HTML based tool on the built site.

If only the tool reported the line numbers where the links were defined (see tcort/markdown-link-check#93).

[1]: https://github.com/tcort/markdown-link-check


waylan:
Ran into tcort/markdown-link-check#86 in the Contributing Guide, so I used a normal link, rather than an automatic link.

PR #956, opened 2020-05-08
Documentation cleanup.

waylan:
This fixes a few errors in the documentation, clarifies a few things, and updates a few examples for Python 3 only. Also, some TOC items were given more readable labels with the `data-toc-label` attribute.

Note: @facelessuser, we are now linking to your SuperFences extension in the warning that the `fenced_code` extension cannot be nested. I no longer have any delusions of updating the `fenced_code` extension to support nesting. We might as well point users at a working solution.

### DIFF
--- a/.spell-dict
+++ b/.spell-dict
@@ -113,6 +113,7 @@ stdout
 Stelios
 Stienstra
 subclasses
+SuperFences
 svn
 Swartz
 Szakmeister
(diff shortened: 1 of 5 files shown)


facelessuser:
Thanks for the heads up.

The only way to fix `fenced_code` is to rework how blocks are handled in Python Markdown. SuperFences, I'll admit, it is more hacky in it's approach as it has to work around the current flaws, but it does work.

If we one day resolve the block processing, I think we could provide a fairly solid `fenced_code` extension. But it's tricky as I feel  there are some core processes that frustrate it right now. How we split blocks being one of them.

Anyways, it's a big issue to fix, and probably requires some pretty deep changes, which is most likely why none of us have fixed block processors yet 🙂.

PR #958, opened 2020-05-08
Switch to GitHub Action from Travis

waylan:
Just seeing if we can replace Travis. 

This is currently incomplete.


facelessuser:
From my personal experience, you can replace almost everything with actions, deployment, testing, document deployment, etc. There may be some things it can't handle. I don't run pypy tests on any of mine, so I can't answer for that yet, but I assume there is a way.

facelessuser:
Just read that you can specify Python versions `3.5`, `3.6`, etc. and `pypy2`, `pypy3`. So I imagine you can do everything through Actions if desired.

waylan:
How do I get the action to run? I assumed creating the workflow file with a `pull_request` trigger would be enough to trigger the action on the PR. Or does the workflow file need to be in the master branch first?

One thing I haven't found is how to install the `tidylib` C library. On Travis we install it via `apt`. Same for `aspell` which we use for spellchecking.

facelessuser:
> One thing I haven't found is how to install the `tidylib` C library. On Travis we install it via apt. Same for `aspell` which we use for spellchecking.

You just use apt: https://github.com/facelessuser/pymdown-extensions/blob/master/.github/workflows/pullrequest.yml#L100

I'm not sure yet why it isn't running your tests though. I assume you have actions enabled for the group? I've never done them under a group, but I imagine they work the same.

facelessuser:
From what I understand, you don't have to have it on master yet. I developed mine in a pull request as well. Granted I used `on: [pull_request]`, but I would think what you are doing should work too.

waylan:
Now they are working. Not sure which change caused that.

waylan:
Hmm, I broke something

facelessuser:
Your's is being weirdly finicky. Your recent change looks fine. The config isn't what broke it because if it does, you'll just get an action run with a yaml config error. I'm running actions now on 90% of my projects on GitHub without issues. For some reason I think actions just isn't reliably triggering for you, which is odd. It must be a weird org thing.

waylan:
I got it working now. Each TOXENV is a separate job in the matrix and we only install the deps we need for each TOXENV.

Turns out the time it failed to run the config had an indentation error. At least I'm assuming that was the issue.

facelessuser:
Nice, I use a matrix for the toxenv as well. Pretty straight forward to manage.

waylan:
Yeah, the example of yours which you linked to was extremely helpful. Thanks.

PR #962, opened 2020-05-11
Fix importlib-metadata requirement name for Python < 3.8

### LINKED ISSUE #961: embedding markdown package in app
Inside an macOS app a package is not installed or pipped but used as plain package inside a `python<version>` folder inside `<appName>.app/Contents/Resources`.

the latest version (3.2.2) raises an error on import of `importlib_metadata` as it cannot find his own version... (traceback below), I assume, even if this could be fixed, `importlib_metadata` will not find any registered extensions as its will never will be installed and registers as an entry_point... 

Any advice on how to use markdown without installing it? this seems now not possible

thanks for all the good work!! 

```
File "/Contents/Resources/lib/python3.7/markdown/__init__.py", line 29, in <module>
  from .core import Markdown, markdown, markdownFromFile # noqa: E402
 File "/Contents/Resources/lib/python3.7/markdown/core.py", line 26, in <module>
  from . import util
 File "/Contents/Resources/lib/python3.7/markdown/util.py", line 34, in <module>
  import importlib_metadata as metadata
 File "/Contents/Resources/lib/python3.7/importlib_metadata/__init__.py", line 602, in <module>
  __version__ = version(__name__)
 File "/Contents/Resources/lib/python3.7/importlib_metadata/__init__.py", line 564, in version
  return distribution(distribution_name).version
 File "/Contents/Resources/lib/python3.7/importlib_metadata/__init__.py", line 537, in distribution
  return Distribution.from_name(distribution_name)
 File "/Contents/Resources/lib/python3.7/importlib_metadata/__init__.py", line 205, in from_name
  raise PackageNotFoundError(name)
importlib_metadata.PackageNotFoundError: importlib_metadata
```

mitya57:
The [`importlib-metadata`](https://pypi.org/project/importlib-metadata/) package on PyPI has a hyphen in its name, not underscore.

This might fix #961.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -88,7 +88,7 @@ setup(
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=3.5',
-    install_requires=["importlib_metadata;python_version<'3.8'"],
+    install_requires=["importlib-metadata;python_version<'3.8'"],
     extras_require={
         'testing': [
             'coverage',


mitya57:
The linkcheck test failed because https://opensource.org/ is unreachable for some reason.

waylan:
> The linkcheck test failed because https://opensource.org/ is unreachable for some reason.

Yeah, I've seen random URLs fail a few times. I'm not sure if we'll keep this test. Or perhaps we should run it as a cron job (GitHub Actions support that) rather than on PRs. The real purpose is to ensure no links in our docs go stale. But we don't want to block a PR when a completely unrelated URL goes offline. Maybe a simple approach it to simply set the config to allow failure.

PR #991, opened 2020-06-26
Limit depth of blockquotes using Python's recursion limit.

### LINKED ISSUE #799: Parsing >>>>>>… takes quadratic time and eventually crashes with recursion error
```pycon
>>> import markdown, timeit
>>> num, dur = timeit.Timer(lambda: markdown.markdown(">" * 100)).autorange(); dur / num
0.004344235559692607
>>> num, dur = timeit.Timer(lambda: markdown.markdown(">" * 200)).autorange(); dur / num
0.016907797248859425
>>> num, dur = timeit.Timer(lambda: markdown.markdown(">" * 300)).autorange(); dur / num
0.05682634799741208
>>> markdown.markdown(">" * 331)
…
RecursionError: maximum recursion depth exceeded while calling a Python object
```

Same results in 3.0.1 and master.

waylan:
If the Python stack comes within 100 frames of the recursion limit,
then the nesting limit of blockquotes is met. Any remaining text,
including angle brackets, are simply wrapped in a paragraph. To
increasing the nesting depth, increase Python's recursion limit.
However, be aware that each level of recursion will likely result in
multiple frames being added to the Python stack. Therefore, the
recursion depth and nesting depth are not one-to-one.

Performance is an concern here. However, the current solution seems like
a reasonable compromise. It doesn't slow things down too much, but also
avoids Markdown input resulting in an error. This is mostly only a concern
with contrived input anyway. For the average Markdown document, this will
likely never be an issue.

Fixes #799.

### DIFF
--- a/docs/change_log/release-3.3.md
+++ b/docs/change_log/release-3.3.md
@@ -53,6 +53,7 @@ The following new features have been included in the 3.3 release:
 
 The following bug fixes are included in the 3.3 release:
 
+* Avoid a `RecursionError` from deeply nested blockquotes (#799).
 * Fix issues with complex emphasis (#979).
 * Limitations of `attr_list` extension are Documented (#965).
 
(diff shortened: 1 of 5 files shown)


waylan:
This still needs tests and a note in the release notes.

PR #1007, opened 2020-07-26
Fix complex scenarios with definition, ordered, and unordered lists

### LINKED ISSUE #918: definition lists containing multi-paragraph ordered lists rendered as code
Given Markdown with a definition list that contains an ordered list and the ordered list has multi-paragraph elements, each ordered list element is rendered as a separate ordered list and the paragraphs are rendered as code blocks.

This appears specific to definition lists (using `def_list`). It does not appear that other types of lists make a similar formatting decision for multi-paragraph sub-lists.

The markdown input:

```markdown
term

:   this is a definition for term. it has 
    multiple lines in the first paragraph.

    1.  first thing
    
        first thing details in a second paragraph.
        
    1.  second thing

        second thing details in a second paragraph.

    1.  third thing

        third thing details in a second paragraph.
```

Expected:

```
<dl>
<dt>term</dt>
<dd>
<p>this is a definition for term. it has
multiple lines in the first paragraph.</p>

<ol>
<li>first thing
<p>first thing details in a second paragraph.</p>
</li>

<li>second thing
<p>second thing details in a second paragraph.</p>
</li>

<li>third thing
<p>third thing details in a second paragraph.</p>
</li>
</ol>

</dd>
</dl>
```

Actual:

```
<dl>
<dt>term</dt>
<dd>
<p>this is a definition for term. it has
multiple lines in the first paragraph.</p>
<ol>
<li>first thing</li>
</ol>
<pre><code>first thing details in a second paragraph.
</code></pre>
<ol>
<li>second thing</li>
</ol>
<pre><code>second thing details in a second paragraph.
</code></pre>
<ol>
<li>third thing</li>
</ol>
<pre><code>third thing details in a second paragraph.
</code></pre>
</dd>
</dl>
```

Tested with:

```bash
/usr/local/bin/markdown_py -x def_list definitions.md
/usr/local/bin/markdown_py --version
# => markdown_py 3.2.1
```

facelessuser:
Adds logic to the definition list indentation parser to be aware of unordered and ordered lists.

Fixes #918

### DIFF
--- a/docs/change_log/release-3.3.md
+++ b/docs/change_log/release-3.3.md
@@ -76,7 +76,7 @@ The following bug fixes are included in the 3.3 release:
 * Fix issues with complex emphasis (#979).
 * Fix unescaping of HTML characters `<>` in CodeHilite (#990).
 * Fix complex scenarios involving lists and admonitions (#1004).
-* Fix complex scenarios with nested ordered and unordered lists in a defintion list (#918).
+* Fix complex scenarios with nested ordered and unordered lists in a definition list (#918).
 
 [spec]: https://www.w3.org/TR/html5/text-level-semantics.html#the-code-element
 [fenced_code]: ../extensions/fenced_code_blocks.md
--- a/docs/change_log/release-3.3.md
+++ b/docs/change_log/release-3.3.md
@@ -75,7 +75,8 @@ The following bug fixes are included in the 3.3 release:
 * Avoid a `RecursionError` from deeply nested blockquotes (#799).
 * Fix issues with complex emphasis (#979).
 * Fix unescaping of HTML characters `<>` in CodeHilite (#990).
-* Fix complex scenarios involving lists and admonitions (#1004)
+* Fix complex scenarios involving lists and admonitions (#1004).
+* Fix complex scenarios with nested ordered and unordered lists in a defintion list (#918).
 
 [spec]: https://www.w3.org/TR/html5/text-level-semantics.html#the-code-element
 [fenced_code]: ../extensions/fenced_code_blocks.md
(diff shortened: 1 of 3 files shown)


facelessuser:
I want to look into why coverage isn't like I expect. It seems the very minimum to get the tests to pass is to add `ul` and `ol` to the def list indention processor and `li` as well. With that said, none of the tests in the entire code base seem to exercise the `create_item` function. I want to try and force that area if possible to see what it actually does. Right now, I think it may not do what I think it does. And to be honest, it may never get executed at all, but I need to at least explore and understand what is supposed to trigger it.

facelessuser:
@waylan I removed changes to `def_list`'s `create_item` function. Nothing in the test suite exercises that code, not in the original, which `def_list` derives from, nor in the `def_list` one. I'm not even sure it can be triggered. Maybe it did at one time 🤷 .

Regardless, it makes no sense for me to alter code that can not be validated, isn't validated, and probably never actually run. I figure if a scenario comes up in the future that can actually trigger the code, and there is a problem, we can fix it then.

At this point, I think I'm done with this pull.

waylan:
I did a quick search, and it appears the only place `create_item` is called is here:

https://github.com/Python-Markdown/markdown/blob/be7ba7bfa17e7cb0c795e101662d1e490334ec84/markdown/blockprocessors.py#L203-L205

As it turns out, line 204 is listed among the lines with no test coverage, even for `ol` and `ul` lists. It appears that it is simply a safety net which was added at the end of a bunch of `elif` statements. My guess is it would never get called. Which begs the question: why was it ever added to the `def_list` extension?

In any event, thanks for the work on this.

facelessuser:
It's possible that at some point it did get exercised. Whatever the case, whether just a safety net, or made that way over time,  it doesn't seem necessary today 🙂.

Issue #950, opened 2020-04-28
3.2.1 breaks when defusedxml is loaded


rcarmo:
I was upgrading one of my projects from `markdown==3.1` to `markdown==3.2.1`, and had a _very_ weird behaviour that was quite hard to track down. 

In my test case, I had this in my imports:

```python
from markdown import Markdown
from nbconvert import HTMLExporter

def render_markdown(raw):
    return markdown_renderer.reset().convert(raw)

def render_ipynb(raw):
    exporter = HTMLExporter()
    exporter.template_file="basic"
    return smartypants(exporter.from_notebook_node(reads(renumber_notebook(raw), 4))[0]
```

When I called _only_ `render_markdown`, I got this pretty obscure exception:

```
Traceback (most recent call last):
  File "markdown_test.py", line 81, in <module>
    print(render_markdown(test_block))
  File "markdown_test.py", line 79, in render_markdown
    return markdown_renderer.reset().convert(raw)
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/core.py", line 269, in convert
    newRoot = treeprocessor.run(root)
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/extensions/footnotes.py", line 388, in run
    footnotesDiv = self.footnotes.makeFootnotesDiv(root)
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/extensions/footnotes.py", line 181, in makeFootnotesDiv
    self.parser.parseChunk(surrogate_parent, self.footnotes[id])
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/blockparser.py", line 105, in parseChunk
    self.parseBlocks(parent, text.split('\n\n'))
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/blockparser.py", line 123, in parseBlocks
    if processor.run(parent, blocks) is not False:
  File "/Users/rcarmo/.pyenv/versions/3.6.9/lib/python3.6/site-packages/markdown/blockprocessors.py", line 591, in run
    p = etree.SubElement(parent, 'p')
TypeError: SubElement() argument 1 must be xml.etree.ElementTree.Element, not Element
make: *** [smoke-test] Error 1
```

[...]

PR #1022, opened 2020-09-16
Made package do specific imports from xml.etree.ElementTree to prevent collision with defusedxml

OrBin:
This comes in order to fix an error I experience when using both `markdown` and `defusedxml`:
```
TypeError: insert() argument 2 must be xml.etree.ElementTree.Element, not Element
```
My solution is based on [this comment](https://github.com/tiran/defusedxml/issues/43#issuecomment-562995041) on the issue in `defusedxml`.

### DIFF
--- a/markdown/blockparser.py
+++ b/markdown/blockparser.py
@@ -19,7 +19,7 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
-import xml.etree.ElementTree as etree
+from xml.etree.ElementTree import Element, ElementTree
 from . import util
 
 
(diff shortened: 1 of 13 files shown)


OrBin:
Actually it looks like it still doesn't fix my issue, I guess this is due to the fact that utils.py imports the whole `xml.etree.ElementTree` anyway. Do I have another way to keep this functioning without importing the whole module?

OrBin:
Turns out that the fix is probably needed in `pymdown-extensions`, not in `markdown`, so I'm closing this PR.

waylan:
Please see #950. This is an issue that needs to be fixed by defusedxml. Even if your "fix" worked, would we always maintain the same behavior with future changes? We can't guarantee that. And its not reasonable for us to test against some random third party lib which we don't use. After all, we are using common Python APIs that always work everywhere else. Defusedxml is the only thing that is broken here, so that is where the fix needs to happen.

facelessuser:
I am the author of pymdown-extensions, and I will not support such a hack either. Effort should be put into getting defusedxml to fix this on their end instead.

PR #1025, opened 2020-10-01
Update makefile

MeerAzzum:


mitya57 (changes requested):
The new command is invalid. What did you want to achieve?

waylan:
As @mitya57 observes, stringing two commands together with `or` is not valid. Perhaps you want `&&`, but then shouldn't `build` come before `install`, not after?

However, if you wanted to run both the `build` and `install` commands together, that already happens with the `install` command. `python setup.py install` runs `build` first automatically. There is no reason to run it separately.

However, if you want to only build, but not install, then that would need to be a separate command, not part of the `install` command. And that sub-command already exists as `make build`.

Given the above, I'm inclined to close this. If you were trying to accomplish something else, leave a comment and we will consider reopening.

PR #1026, opened 2020-10-01
Add PY39 support and drop PY35

waylan:
Python 3.5 reached end-of-life on 2020-09-12. Python 3.9 is scheduled for release on 2020-10-05. Might as well address both together.

Note, I haven't tested on PY39 yet. The GH Action/Check will be the first and I haven't checked if anything special is required to run a pre-release version. If it fails, I'll come back after the final release and rerun it.

### DIFF
--- a/.github/workflows/tox.yml
+++ b/.github/workflows/tox.yml
@@ -20,16 +20,16 @@ jobs:
       fail-fast: false
       max-parallel: 4
       matrix:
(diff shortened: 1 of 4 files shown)


waylan:
I had to update from `actions/setup-python@v1` to `actions/setup-python@v2` which added support for installing Python versions which are not yet available on the local system. With that, PY39 passes. 👍

PR #1029, opened 2020-10-07
Drop unneeded dependency python_version

philip-khor:
Line 78 implies that the dependency [python_version](https://pypi.org/project/python_version/) is required for a [minimal installation](https://packaging.python.org/discussions/install-requires-vs-requirements/) of markdown. 

However, searching through the code and looking at https://github.com/Python-Markdown/markdown/commit/102e01c2462f20fdd29073bc0b1fe2252b517add#diff-2eeaed663bd0d25b7e608891384b7298 I don't see python_version imported in this package. 

Apologies in advance if I'm mistaken.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -75,7 +75,7 @@ setup(
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=3.6',
-    install_requires=["importlib-metadata;python_version<'3.8'"],
+    install_requires=["importlib-metadata"],
     extras_require={
         'testing': [
             'coverage',


waylan:
You misunderstand the purpose of that line. Note the semicolon between `importlib-metadata` and `python_version<'3.8'` as well as the fact that it is part of the same string. It is specifying that the `importlib-metadata` dependency is only required for a version of Python less than 3.8. As it happens, [importlib_metadata][1] is a backport of Python 3.8’s standard library [importlib.metadata][2] module. Therefore, we do not need the dependency in Python version 3.8 and later. Removing that part of the string would cause the dependency to be installed unnecessarily in recent versions of Python.

[1]: https://importlib-metadata.readthedocs.io/en/latest/
[2]: https://docs.python.org/3/library/importlib.metadata.html

Issue #1030, opened 2020-10-07
Tests fail with pygments-2.7.0+


mgorny:
Pygments 2.5.2 (as used in tox.ini) are pretty old. After upgrading to 2.7.0 or 2.7.1, I get the following test failures.

I can make a pull request updating the tests if you wish. Please just tell me if you prefer updating them purely for new pygments version, or if I should try to fit some regexps in to support both.

```
======================================================================
FAIL: test_codehilite_linenos_hl_lines (test_syntax.extensions.test_code_hilite.TestCodeHiliteClass)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/markdown/tests/test_syntax/extensions/test_code_hilite.py", line 289, in test_codehilite_linenos_hl_lines
    self.assertOutputEquals('line 1\nline 2\nline 3', expected, lang='text', linenos='inline', hl_lines=[1, 3])
  File "/tmp/markdown/tests/test_syntax/extensions/test_code_hilite.py", line 43, in assertOutputEquals
    self.assertMultiLineEqual(output.strip(), expected)
AssertionError: '<div[58 chars]inenos">1</span><span class="hll">line 1\n</sp[119 chars]div>' != '<div[58 chars]ineno">1 </span><span class="hll">line 1\n</sp[119 chars]div>'
- <div class="codehilite"><pre><span></span><code><span class="linenos">1</span><span class="hll">line 1
?                                                                    -
+ <div class="codehilite"><pre><span></span><code><span class="lineno">1 </span><span class="hll">line 1
?                                                                       +
- </span><span class="linenos">2</span>line 2
?                           -
+ </span><span class="lineno">2 </span>line 2
?                              +
- <span class="linenos">3</span><span class="hll">line 3
?                    -
+ <span class="lineno">3 </span><span class="hll">line 3
?                       +
  </span></code></pre></div>

======================================================================
FAIL: test_codehilite_linenos_inline (test_syntax.extensions.test_code_hilite.TestCodeHiliteClass)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/tmp/markdown/tests/test_syntax/extensions/test_code_hilite.py", line 216, in test_codehilite_linenos_inline
    self.assertOutputEquals('plain text', expected, lang='text', linenos='inline')
  File "/tmp/markdown/tests/test_syntax/extensions/test_code_hilite.py", line 43, in assertOutputEquals
    self.assertMultiLineEqual(output.strip(), expected)
[...]

PR #1034, opened 2020-10-08
Some test tweaks.

waylan:
* Pygments specific tests now only run when the pygments version installed
  matches the expected version. That version is defined in an environment
  variable (PYGMENTS_VERSION) in the 'pygments' tox env (see #1030).
* When the Python lib tidylib is installed but the underlying c lib is not,
  the relevant tests are now skipped rather than fail. This matches the
  behavior when the Python lib is not installed. The tox envs are now useful
  on systems which don't have the c lib installed.

### DIFF
--- a/markdown/test_tools.py
+++ b/markdown/test_tools.py
@@ -140,8 +140,11 @@ class LegacyTestMeta(type):
                     expected = f.read().replace("\r\n", "\n")
                 output = markdown(input, **kwargs)
                 if tidylib and normalize:
-                    expected = _normalize_whitespace(expected)
-                    output = _normalize_whitespace(output)
+                    try:
+                        expected = _normalize_whitespace(expected)
+                        output = _normalize_whitespace(output)
+                    except OSError:
+                        self.skipTest("Tidylib's c library not available.")
                 elif normalize:
                     self.skipTest('Tidylib not available.')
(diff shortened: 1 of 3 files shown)


mitya57:
Here you changed `test_fenced_code.py`, but in `test_code_hilite.py` there are also tests that may depend on Pygments version.

For them it is not enough to set `has_pygments = False`, because that would trigger the fallback path that tests output when Pygments is not installed, but that path will fail because Pygments actually *is* installed. I think we should just skip the tests in that case.

So I propose adding to `TestCodeHiliteClass` and `TestCodeHiliteExtension` something like this:
```python
    def setUp(self):
        required_pygments_version = os.environ.get('PYGMENTS_VERSION', '')
        if has_pygments and pygments.__version__ != required_pygments_version:
            self.skipTest('Pygments=={} required'.format(required_pygments_version))
```

What do you think?

waylan:
> but that path will fail because Pygments actually is installed. 

You are correct. In fact, the same problem exists with with the changes I made here in `test_fenced-code.py`. Apparently I didn't run a test of that scenario before committing.

In any event, your solution is a good one. Adopted in #1035. I broke out the fenced_code tests into a separate test class so I could do the same with them.

mitya57:
> Adopted in #1035.

Thank you!

PR #1051, opened 2020-10-22
Add require_changelog workflow action

waylan:
This action checks that an update was made to any file in docs/change_log/
but only if changes were made to files in markdown/. Presumably,
any changes outside of markdown/ do not affect the behavior and do not
require a notation in the change_log.

When a change_log entry is missing, a comment is added to the PR informing
the PR author. If a later commit to the PR includes a change_log entry,
then the comment is hidden. A comment should only be added once per PR.

### DIFF
--- a/.github/workflows/process.yml
+++ b/.github/workflows/process.yml
@@ -22,3 +22,4 @@ jobs:
           change log describing the changes you have made. Doing so will help to ensure your contribution is accepted.
 
           Please see the [Contributing Guide](https://python-markdown.github.io/contributing/#pull-requests) for details.
+        token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/process.yml
+++ b/.github/workflows/process.yml
@@ -1,7 +1,8 @@
 name: bot
 
 on:
-  pull_request:
+  pull_request_target:
+    types: [opened, synchronize, reopened]
     branches:
     - '**'
 
--- a/.github/workflows/process.yml
+++ b/.github/workflows/process.yml
@@ -1,7 +1,7 @@
 name: bot
 
 on:
-  pull_request_target:
+  pull_request:
     branches:
     - '**'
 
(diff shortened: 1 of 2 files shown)


waylan:
I looked at the following actions before settling on the first one.

1. [Missing Changelog Reminder](https://github.com/marketplace/actions/missing-changelog-reminder)

   I chose this one because it provides a regex for the changelog path. As we sometimes have updates to the change_log and other times have updates to a release notes file, this was the best way to include both. However, there are some PRs (like this one) which wouldn't necessarily need a changelog entry. Therefore we can use the `include` option to narrow the check to changes within the `markdown/` dir.  It will also add a comment to the PR when an update to the PR is missing as a reminder to the author.

2. [Changelog Reminder](https://github.com/marketplace/actions/changelog-reminder)

    This appears to be a simpler variation on the previous one. It does not support the `include` option and would run the check for all changes. There is no way to conditionally disable the check. It also adds a comment to the PR.

3. [Changelog Enforcer](https://github.com/marketplace/actions/changelog-enforcer)

    This takes a filename rather than a regex and assumes the changelog is a single file, which won't work for us. A nice feature is the ability to disable the check by assigning a specific label to the PR. There is no way to add a comment to the PR with this.

4. [Changelog checker](https://github.com/marketplace/actions/changelog-checker)

    This appears to be a clone of the previous item in that the feature set is exactly the same.

waylan:
Hmm, well it seems to be working except for posting the comment. We're getting `Error: Resource not accessible by integration`. After doing some searching I finally landed upon [GitHub Actions improvements for fork and pull request workflows](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) which explains how to work around the fact that a PR from a fork does not normally have write permissions on the base repo.

waylan:
Well, that's annoying. Switching from the `pull_request` event to the `pull_request_target` event causes the entire action to no longer even be listed as an action.

waylan:
Setting the `message` to an empty string does not disable the comment feature of the action. Guess I'll just leave it alone and we can ignore the error message. The fact that the test fails under the correct conditions should be enough to remind everyone.

PR #1054, opened 2020-10-24
Fix issues related to hr tags

facelessuser:
Ensure that start/end tag handler does not include tags in the previous
paragraph.

Provide special handling for tags like hr that never have content.

### DIFF
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -36,8 +36,9 @@ class HTMLExtractorExtra(HTMLExtractor):
         # Block-level tags which never get their content parsed.
         self.raw_tags = ['canvas', 'math', 'option', 'pre', 'script', 'style', 'textarea']
         # Block-level tags in which the content gets parsed as blocks
-        self.block_tags = [tag for tag in self.block_level_tags if tag not in self.span_tags + self.raw_tags]
         super().__init__(md, *args, **kwargs)
+
+        self.block_tags = [tag for tag in self.block_level_tags if tag not in self.span_tags + self.raw_tags + self.empty_tags]
 
     def reset(self):
         """Reset this instance.  Loses all unprocessed data."""
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -30,15 +30,18 @@ class HTMLExtractorExtra(HTMLExtractor):
 
     def __init__(self, md, *args, **kwargs):
         # All block-level tags.
-        self.block_level_tags = md.block_level_elements.copy()
+        self.block_level_tags = set(md.block_level_elements.copy())
         # Block-level tags in which the content only gets span level parsing
-        self.span_tags = ['address', 'dd', 'dt', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'legend', 'li', 'p', 'td', 'th']
+        self.span_tags = set(
+            ['address', 'dd', 'dt', 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'legend', 'li', 'p', 'td', 'th']
+        )
         # Block-level tags which never get their content parsed.
-        self.raw_tags = ['canvas', 'math', 'option', 'pre', 'script', 'style', 'textarea']
+        self.raw_tags = set(['canvas', 'math', 'option', 'pre', 'script', 'style', 'textarea'])
         # Block-level tags in which the content gets parsed as blocks
         super().__init__(md, *args, **kwargs)
 
-        self.block_tags = [tag for tag in self.block_level_tags if tag not in self.span_tags + self.raw_tags + self.empty_tags]
+        self.block_tags = set(self.block_level_tags) - (self.span_tags | self.raw_tags | self.empty_tags)
+        self.span_and_blocks_tags = self.block_tags | self.span_tags
 
     def reset(self):
         """Reset this instance.  Loses all unprocessed data."""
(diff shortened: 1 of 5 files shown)


facelessuser:
Needs tests and is a work in progress. Mainly something that can be discussed.

markdown/htmlparser.py
```diff
@@ -56,6 +56,10 @@ class HTMLExtractor(htmlparser.HTMLParser):
     def __init__(self, md, *args, **kwargs):
         if 'convert_charrefs' not in kwargs:
             kwargs['convert_charrefs'] = False
+
+        # Block tags that should contain no content (self closing)
+        self.empty_tags = ['hr']
```
waylan: You seem to think we need to treat `<hr>` specially. However, I would expect all 'block-level tags' to behave the same way. Therefore `<script />` would be no different. We only need to check if a tag is a block-level tag in `handle_empty_tags`. No need to specifically list empty tags separately.

facelessuser: `<hr>` is a different case than `<script />`. The latter is giving a hint that it is self closed.

`<hr>` gives no hint, and usually, you would keep looking for an ending tag. But `<hr>` *never* has content. Maybe I'm missing something, but this is a very special case.

markdown/htmlparser.py
```diff
@@ -264,6 +272,14 @@ def parse_starttag(self, i):  # pragma: no cover
         if end.endswith('/>'):
             # XHTML-style empty tag: <span attr="value" />
             self.handle_startendtag(tag, attrs)
+        elif tag in self.empty_tags:
```
waylan: There should be no changes to the `parse_starttag` method. We only override this method to fix the upstream issue. In all other respects this is an exact copy of the parent class. See the comment on this method would suggest the same. We need to address the issue in the public functions first if we can.

markdown/htmlparser.py
facelessuser: In fact, given:

```
<hr>
some text
</hr>
```

You should treat them both as separate `<hr>` like browsers do.

waylan: I'm operating on the assumption that the HtmlParser properly recognizes that `<hr>` is a self-closing tag. If that is not the case, then that might be different.

markdown/extensions/md_in_html.py
```diff
@@ -36,9 +36,10 @@ def __init__(self, md, *args, **kwargs):
         # Block-level tags which never get their content parsed.
         self.raw_tags = ['canvas', 'math', 'option', 'pre', 'script', 'style', 'textarea']
         # Block-level tags in which the content gets parsed as blocks
-        self.block_tags = [tag for tag in self.block_level_tags if tag not in self.span_tags + self.raw_tags]
         super().__init__(md, *args, **kwargs)
 
+        self.block_tags = [tag for tag in self.block_level_tags if tag not in self.span_tags + self.raw_tags + self.empty_tags]
```
waylan: Btw, I see you first tried to use sets here. The appropriate way to do that would be:

```python
self.block_tags = set(self.block_level_tags) - set(self.span_tags + self.raw_tags + self.empty_tags)
```

I debated which way to go with it. In the end I chose to keep using lists rather than changing to sets. I did not check performance. If sets are faster, then by all means, feel free to make the change.

markdown/htmlparser.py
waylan: Sure enough, the HtmlParser does not recognize `<hr>` as a self-closing tag.

```python
>>> from test.test_htmlparser import EventCollector
>>> ec = EventCollector()
>>> ec.feed('<hr>')
>>> ec.close()
>>> ec.get_events()
[('starttag', 'hr', [])]
>>> ec = EventCollector()
>>> ec.feed('<hr/>')
>>> ec.close()
>>> ec.get_events()
[('startendtag', 'hr', [])]
```

waylan: So I see that `<hr />` fails in a different way:

```html
<p>foo
    <hr />
</p>
<p>bar</p>
```

Sigh. Apparently it never occurred to me we would need separate tests specifically for `hr` tags.

facelessuser:
I've added tests. For cases with ending `</hr>` tags, I'm not overly concerned with the unexpected behavior. One should never use ending `</hr>` tags. I was examining browser behavior, and they seem to ignore them. `<hr>` and `<hr/>` should get treated in a sane manner though.

facelessuser:
This took me longer than I planned. A second set of eyes to double check I'm not missing something may be good. Let me know if there are any test cases I'm missing.

Hopefully, the approach is fine. It took me a bit to massage it into a state where I was able to handle our failing cases, but not break other cases.

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -1402,3 +1402,67 @@ def test_inline_script_tags(self):
                 """
             )
         )
+
+    def test_hr_only_start(self):
+        self.assertMarkdownRenders(
+            "*emphasis1*\n<hr>\n*emphasis2*",
```
waylan: I find these condensed strings a little more difficult to parse (as a human) than I like. Can we please use `dedent` and string literals? It is much easier to read that way.

tests/test_syntax/blocks/test_html_blocks.py
```diff
@@ -1402,3 +1402,67 @@ def test_inline_script_tags(self):
...
+
+    def test_hr_start_and_end(self):
+        self.assertMarkdownRenders(
+            "*emphasis1*\n<hr></hr>\n*emphasis2*",
+            self.dedent(
+                """
+                <p><em>emphasis1</em></p>
+                <hr>
+                <p></hr>
+                <em>emphasis2</em></p>
```
waylan, line 1457: This one is surprising to me. I would expect:

```html
<p><em>emphasis1</em></p>
<hr>
</hr>
<p><em>emphasis2</em></p>
```

I realize the `/` is at the start of the tag, not the end. I suppose the browser probably is weird about this. I might need to think some more about this.

waylan (changes requested):
This is looking good. I have a couple notes below.

tests/test_syntax/blocks/test_html_blocks.py
waylan: > One should never use ending `</hr>` tags. I was examining browser behavior, and they seem to ignore them. 

You're right. They are completely ignored by browsers. That being the case, then I guess it doesn't matter how we handle them. Perhaps we should add a comment on this test about browser behavior.

waylan (approved):

Issue #1068, opened 2020-11-16
md_in_html: broken `code span`


dbader:
`markdown.extensions.md_in_html` fails to escape HTML tags in `monospace` text placed inside a `markdown="1"` div wrapper:

```pycon
>>> import markdown
>>> markdown.markdown('<div markdown="1">\n`<h1>escaped</h1>`\n</div>', extensions=["markdown.extensions.md_in_html"])
'<p><h1><div markdown="block">\n`escaped&lt;/h1&gt;`\n</div></p>'
```

```html
<p><h1><div markdown="block">
`escaped&lt;/h1&gt;`
</div></p>
```

The inner `<code>` element should look like this: `<code>&lt;h1&gt;escaped&lt;/h1&gt;</code>`, but instead the h1 inside the `monospace` text appears as an actual `<h1>` tag in the output.

## Versions

* Markdown==3.3.3
* CPython 3.8.6

dbader:
Interestingly this seem to work correctly on `Markdown==3.2.2`:

```pycon
>>> import markdown
>>> markdown.markdown('<div markdown="1">\n`<h1>escaped</h1>`\n</div>', extensions=["markdown.extensions.md_in_html"])
'<div>\n<p><code>&lt;h1&gt;escaped&lt;/h1&gt;</code></p>\n</div>
```

dbader:
Btw I super appreciate the work you guys are doing here ❤️  I'm using Python-Markdown for realpython.com (also your amazing [PyMdown extensions](https://facelessuser.github.io/pymdown-extensions/) @facelessuser) and it's a pleasure using this library. Let me know if I can provide additional info here to help track this down 🙂

facelessuser:
> Interestingly this seems to work correctly on `Markdown==3.2.2`

This is because a new HTML parser was introduced in 3.3

Though the first wave of bugs was the kind I expected, I'm starting to get a little concerned about the new parser. The handling of block elements in inline code is a little troubling, coupled with some of the recent bugs.

As far as the code block part goes, that is one thing the old parser took into consideration. It understood that it didn't have all the context it needed as Python Markdown actually takes multiple passes while some other parsers tokenized everything in one pass.

Unfortunately, since we do not tokenize everything in one pass, I really do think block HTML logic should only come into play when the block tag is at the start of a line. We should only process inline tags once we've processed a Markdown block with the code step.

[...]

PR #1069, opened 2020-11-17
Properly parse code spans in md_in_html

waylan:
This reverts part of 2766698 and causes other tests to break. Although,
those should be addressed in the same manner as in the core (by using
the intail attribute).

A partial fix for #1068.


waylan:
Down to three failing tests.

PR #1083, opened 2020-12-03
Use simplified regex for html placeholders, correctly

raabrp:
A reimplementation of c5f1395a944e76e95250740452706f6b16830cae

after issue #932 noted that it broke a bunch of tests

and the change was reverted by 7c595e28491d3cca28bc9901bb099bca8bcf4626

Mea culpa.

This one passes the test suite of 927 tests with 119 skipped (this is desired result, correct?), and restores the performance improvement.

### DIFF
--- a/markdown/postprocessors.py
+++ b/markdown/postprocessors.py
@@ -85,7 +85,7 @@ class RawHtmlPostprocessor(Postprocessor):
 
         if replacements:
             base_placeholder = util.HTML_PLACEHOLDER % r'([0-9]+)'
-            pattern = re.compile( "<p>{}</p>".format(base_placeholder) + '|' + base_placeholder )
+            pattern = re.compile("<p>{}</p>".format(base_placeholder) + '|' + base_placeholder)
             processed_text = pattern.sub(substitute_match, text)
         else:
             return text
--- a/markdown/postprocessors.py
+++ b/markdown/postprocessors.py
@@ -79,13 +79,13 @@ class RawHtmlPostprocessor(Postprocessor):
             key = m.group(0)
 
             if key not in replacements:
-                return '<p>' + replacements[key[3:-4]] + '</p>'
+                return f'<p>{ replacements[key[3:-4]] }</p>'
 
             return replacements[key]
 
         if replacements:
             base_placeholder = util.HTML_PLACEHOLDER % r'([0-9]+)'
-            pattern = re.compile("<p>{}</p>".format(base_placeholder) + '|' + base_placeholder)
+            pattern = re.compile(f'<p>{ base_placeholder }</p>|{ base_placeholder }')
             processed_text = pattern.sub(substitute_match, text)
         else:
             return text
--- a/markdown/postprocessors.py
+++ b/markdown/postprocessors.py
@@ -75,9 +75,18 @@ class RawHtmlPostprocessor(Postprocessor):
                     self.md.htmlStash.get_placeholder(i))] = html
             replacements[self.md.htmlStash.get_placeholder(i)] = html
 
+        def substitute_match(m):
+            key = m.group(0)
+
+            if key not in replacements:
+                return '<p>' + replacements[key[3:-4]] + '</p>'
+
+            return replacements[key]
+
         if replacements:
-            pattern = re.compile("|".join(re.escape(k) for k in replacements))
-            processed_text = pattern.sub(lambda m: replacements[m.group(0)], text)
+            base_placeholder = util.HTML_PLACEHOLDER % r'([0-9]+)'
+            pattern = re.compile( "<p>{}</p>".format(base_placeholder) + '|' + base_placeholder )
+            processed_text = pattern.sub(substitute_match, text)
         else:
             return text
 
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -8,6 +8,7 @@ Under development: version 3.3.4 (a bug-fix release).
 * Properly parse unclosed tags in code spans (#1066).
 * Properly parse processing instructions in md_in_html (#1070).
 * Properly parse code spans in md_in_html (#1069).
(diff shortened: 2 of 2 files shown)


raabrp:
It appears I need a changelog and should adhere to the style linter. Will do.

raabrp:
It appears that the checklinks fail for CI is the result of missing documentation, but I'm not sure how to resolve it. (?)

markdown/postprocessors.py
```diff
@@ -75,9 +75,18 @@ def run(self, text):
                     self.md.htmlStash.get_placeholder(i))] = html
             replacements[self.md.htmlStash.get_placeholder(i)] = html
 
+        def substitute_match(m):
+            key = m.group(0)
+
+            if key not in replacements:
+                return '<p>' + replacements[key[3:-4]] + '</p>'
```
waylan: Here it would be:

```python
return f'<p>{ replacements[key[3:-4]] }</p>' 
```

markdown/postprocessors.py
```diff
@@ -75,9 +75,18 @@ def run(self, text):
...
+
+            return replacements[key]
+
         if replacements:
-            pattern = re.compile("|".join(re.escape(k) for k in replacements))
-            processed_text = pattern.sub(lambda m: replacements[m.group(0)], text)
+            base_placeholder = util.HTML_PLACEHOLDER % r'([0-9]+)'
+            pattern = re.compile("<p>{}</p>".format(base_placeholder) + '|' + base_placeholder)
```
waylan: And this would be:

```python
pattern = re.compile(f'<p>{ base_placeholder }</p>|{ base_placeholder }')
```

waylan (changes requested):
As we only support Python 3.6+, we can use [f-strings][1] which are much more performant that any other method of building strings. Let's try to use them anywhere we are updating code. 

Oh, and you can ignore the checklinks failure. I think we need to update the way that works.

[1]: https://www.python.org/dev/peps/pep-0498/

markdown/postprocessors.py
raabrp: Excellent. Thank you for the recommendation.

waylan (approved):

markdown/postprocessors.py
```diff
@@ -75,9 +75,18 @@ def run(self, text):
                     self.md.htmlStash.get_placeholder(i))] = html
             replacements[self.md.htmlStash.get_placeholder(i)] = html
 
+        def substitute_match(m):
+            key = m.group(0)
+
+            if key not in replacements:
+                return f'<p>{ replacements[key[3:-4]] }</p>'
```
waylan, line 82: I figured out what I don't like about this. If `key[3:-4]` is also not in `replacements`, then this will raise an error. Also, this assumes that `key` includes `<p>` and `</p>`. However it might not. Seems to me that the correct check would be:

```python
if key not in replacements and key.startswith('<p>') and key.endswith('</p>') and key[3:-4] in replacements:
```

I wonder if perhaps there could be some other scenarios which would require slightly different behavior. For example, we might need an `elif` statement. However, until we find an edge case that fits, we'll have to wait to see if they ever get reported. It's also possible that such cases will never occur.

waylan: I suppose `key.startswith('<p>') and key.endswith('</p>')` doesn't need to be in the if statement. `key[3:-4] in replacements` will only evaluate to `True` when that is the case anyway.

waylan: I'm thinking this should be:

```python
if key not in replacements:
    if key[3:-4] in replacements:
        return f'<p>{ replacements[key[3:-4]] }</p>'
    else:
        return key
```

We just need a test case for the final `else`.

waylan: Here is the test:

```python
def test_placeholder_in_source(self):
    # This should never occur, but third party extensions could create weird edge cases.
    md = markdown.Markdown()
    # Ensure there is an htmlstash so relevant code (nested in `if replacements`) is run.
    md.htmlStash.store('foo')
    # Run with a placeholder which is not in the stash
    placeholder = md.htmlStash.get_placeholder(md.htmlStash.html_counter + 1)
    result = md.postprocessors['raw_html'].run(placeholder)
    self.assertEqual(placeholder, result)
```

Without my proposed change, this test raises a `KeyError`. With my proposed change and this test, we get 100% test coverage of the diff.

waylan:
This is closed in favor of #1086.

markdown/postprocessors.py
raabrp: If I understand correctly, the case of concern is if both `key` and `key[3:-4]` are not in `replacements`, which implies that the regexp search found some sequence that looks like `'\x02wzxhzdk:1234\x03'`(for any digits substituting 1234) that is not a place-holder for a pending substitution, correct?

The `replacements` ordered dict is constructed immediately prior in lines [70-76](https://github.com/Python-Markdown/markdown/blob/48720f1f79480307d384bbfd8e46765ca1430439/markdown/postprocessors.py#L70), so the code is tightly coupled without assumptions from other files or pieces of the codebase. All substitutions in the htmlStash get a key in the `replacement` dict without the `<p>` tags, but some also get a key featuring the `<p>` tags (and with higher priority, since this is an ordered dict).

Based on how `replacements` is constructed as of now, `key not in replacements` is `True` only when `key` has the `<p>` tags. A subsequent `key[3:-4] in replacements` will always evaluate to `True`.

Let me know if this answers questions, or if there are still concerns you have about robustness.

waylan: First of all, if you pass text into Markdown which would match, by the time we get to the postporcessor that is no longer a concern. However, a third-party extension could insert text into the document which would match, but have no corresponding key. Therefore, we need to account for that. My fix does that.

raabrp: Ah. I had not delved into the interaction of extensions with this code. Excellent. Thank you for your help and for accepting the overall changes.

PR #1086, opened 2020-12-08
Use simplified regex for html placeholders

waylan:
This is a copy of #1083 with additions to avoid a `KeyError`. As #1083 was on the `master` branch, it was easier to create a new PR than to push my changes to the existing PR (see the [Contributing Guide][1]).

### DIFF
--- a/markdown/postprocessors.py
+++ b/markdown/postprocessors.py
@@ -85,7 +85,7 @@ class RawHtmlPostprocessor(Postprocessor):
 
         if replacements:
             base_placeholder = util.HTML_PLACEHOLDER % r'([0-9]+)'
-            pattern = re.compile( "<p>{}</p>".format(base_placeholder) + '|' + base_placeholder )
+            pattern = re.compile("<p>{}</p>".format(base_placeholder) + '|' + base_placeholder)
             processed_text = pattern.sub(substitute_match, text)
         else:
             return text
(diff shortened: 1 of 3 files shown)


markdown/postprocessors.py
```diff
@@ -75,9 +75,21 @@ def run(self, text):
                     self.md.htmlStash.get_placeholder(i))] = html
             replacements[self.md.htmlStash.get_placeholder(i)] = html
 
+        def substitute_match(m):
+            key = m.group(0)
+
+            if key not in replacements:
+                if key[3:-4] in replacements:
+                    return f'<p>{ replacements[key[3:-4]] }</p>'
+                else:
```
raabrp, line 84: This line will never be reached as per [this comment](https://github.com/Python-Markdown/markdown/pull/1083#discussion_r538580902)

waylan: Except that it is in the test I added with this fix (see the [coverage report][1]). The assumption is that a third-party extension could create this scenario. We never want to allow a situation where the parser could raise an error based on user input. Therefore, we are leaving this in.

[1]: https://github.com/Python-Markdown/markdown/runs/1518569326

raabrp: I see. I had not delved into the interaction of extensions with this code. Thank you for your help and for accepting the overall changes.

PR #1090, opened 2020-12-21
Optimize away a `len` call in InlineProcessor

oprypin:
The gains are tiny but when the total number of calls to these is in the hundreds of thousands, it makes a sizeable difference.

* ~Eliminate most calls to the _sort function.
  When sorting isn't needed, don't check the condition inside the function only to almost always drop out of it, instead replace the whole function with a no-op.~
* Eliminate most calls to util.py:\_\_len__, just get the length once at the beginning.

~For my tests this was \~25% faster for the function itself and saved a second of build time out of total 20s when running under the profiler.~

Before:

before

After:

after

### DIFF
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -126,7 +126,8 @@ class InlineProcessor(Treeprocessor):
         """
         if not isinstance(data, util.AtomicString):
             startIndex = 0
-            while patternIndex < len(self.inlinePatterns):
+            count = len(self.inlinePatterns)
+            while patternIndex < count:
                 data, matched, startIndex = self.__applyPattern(
                     self.inlinePatterns[patternIndex], data, patternIndex, startIndex
                 )


markdown/util.py
```diff
@@ -378,9 +379,21 @@ def _sort(self):
...
+        return self._sort is _noop
+
+    @_is_sorted.setter
+    def _is_sorted(self, is_sorted):
+        if is_sorted:
+            # Subsequent calls to _sort() will do nothing.
+            self._sort = _noop
+        else:
+            # Restore the original sort function.
+            self.__dict__.pop('_sort', None)
```
waylan: While I understand what is going on here, this feels a little too much like "magic." The next time I look at this code, I may not remember why we are jumping through these hoops. And that means that next time this section of code has any work done, it's likely this would just get ripped out. As a reminder, performance has never been a top priority for us.

oprypin: Yes, in this shape with the 2nd commit it is quite magic. Could you please check out how it looks after just the [**first commit**](https://github.com/Python-Markdown/markdown/pull/1090/commits/a9cbc4fe5ffb5dc184e791b8b42a4cc3e928ffa3) in this PR, and whether we could get *that* into an acceptable shape? I think with enough comments it can be fine.

I can sympathize with the general sentiment. But note that I already filtered out some changes and left only the juiciest ones. And if you consider how many times this code runs on a global scale, it becomes easier to justify.

waylan: So, the options are to either:

1. Call a random builtin C function (you happened to pick `bool`), or
2. Call a Python function which executes one line (`if not self._is_sorted` which evaluates `False`).

The only reason (1) is even an option is because we always ignore the return value. Otherwise, we would need to stick with the Python function.

This seems to really be slitting hairs. I just don't care enough to bother with it and the potential maintenance headache in the future fare outweighs any potential gains. The only reason I haven't closed the PR completely is because the other change (in treeprocessors.py) is a reasonable change.

oprypin: Makes sense.
I'm also now seeing that running under the profiler seems to give Python functions a disadvantage, so the effect of the `_sort` change is even smaller. The `__len__` one is still the bigger one and is noticeable, so let's keep it.

PR #1092, opened 2020-12-30
Don't wrongly discard text immediately before an admonition

oprypin:
```markdown
foo
**foo**
!!! note "Admonition"
```
    
In the previous state, the "foo" text would be completely obliterated, but now it will be kept, and the admonition will be rendered as well.


oprypin:
This is a bug, but it's not obvious what behavior is the correct one.
This PR is the **more pragmatic one**, and https://github.com/Python-Markdown/markdown/pull/1093 is an alternative.
One of the two should be chosen.

PR #1093, opened 2020-12-30
Disallow admonitions in the middle of a block

oprypin:
```markdown
foo
**foo**
!!! note "Admonition"
```
    
In the previous state, the "foo" text would be completely obliterated, but now the admonition will be ignored and all of this will be seen as normal text (with inline processing, of course).

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -8,6 +8,8 @@ Under development: version 3.3.4 (a bug-fix release).
 * Properly parse unclosed tags in code spans (#1066).
 * Properly parse processing instructions in md_in_html (#1070).
 * Properly parse code spans in md_in_html (#1069).
+* Disallow admonitions in the middle of a block (#1093).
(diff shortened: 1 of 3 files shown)


oprypin:
This is a bug, but it's not obvious what behavior is the correct one.
This PR is the **stricter one** (perhaps more in the spirit of original Markdown?), and https://github.com/Python-Markdown/markdown/pull/1092 is an alternative.
One of the two should be chosen.

waylan:
As the original regex begins with `(?:^|\n)` is seems pretty clear to me that the original intent was to not require a blank line before an admonition. This fits quite well with the behavior of headers and other block level elements which provide an obvious start token without needing a blank line.

There is also the possibility that existing documents rely on the existing behavior of not requiring a blank line. If we were to suddenly require one, those documents would break. 

Given the above, this is not the correct way to address this issue.

PR #1097, opened 2021-01-18
Update th/td to use style attribute

gaige:
This allows better interoperation with CSS style sheets, as the align
object on the TH is skipped if the css uses 'text-align: inherit' and
the previous 'text-align' is used instead (or the default: left).

With the existing `align` attribute, standard CSS libraries, such as Bootstrap, can ignore the locally-attached `align` attribute by incorporating `text-align: inherit`. By setting the `text-align` CSS attribute in a `style` attribute attached to the th/td element, the local attribute overrides the CSS.  Without it, the lower-priority `align` attribute is ignored in these circumstances.

This is my first contribution to this project, so please let me know if there's something else you'd like. I think the commit messages, test updates, and change log updates are per spec.


waylan:
Interesting. However, this extension is a strict clone of the PHP Markdown Extra implementation of tables. We expect our output to match that libraries’ output. This would change that. Are there any other implementations which use style attributes rather than align attributes for tables?

gaige:
`align` and `valign` (along with attributes that directly manipulated display attributes) were deprecated in HTML4.  The current PHP implementation has 2 methods of doing alignment, the default is to still use the `align` attribute, but conditionally (with the definition of `table_align_class_tmpl`), it will insert a `class='$table_align_class_tmpl-{left|right|center}'`, which would allow for a somewhat more flexible implementation.  Unfortunately, that requires the introduction of an additional class, but if you'd prefer something like that, I could re-implement with that option. 

In that case, I'd be adding a config item to tables for `ALIGN_CLASS_TMPL` which would default to `None` and, if set, would add `class="<align_class>"` after replacing the `%%` in the string with the `left`, `right`, or `center` string.

That will stick most closely to the implementation in https://github.com/michelf/php-markdown.  I'm not entirely sure where to implement the test, although it appears test_syntax/extensions/test_tables might be a home for it.

Thoughts?

markdown/extensions/tables.py
```diff
@@ -126,7 +126,7 @@ def _build_row(self, row, parent, align):
             except IndexError:  # pragma: no cover
                 c.text = ""
             if a:
-                c.set('align', a)
+                c.set('style', 'text-align:'+a)
```
mitya57: The only thing I would suggest is using a space in `style` attribute, and maybe adding a trailing semicolon.
```suggestion
                c.set('style', f'text-align: {a};')
```

mitya57 (commented):
I agree with @gaige — the attribute way of specifying alignment was deprecated in HTML 4 and the HTML 5 specification does not even mention it:
https://www.w3.org/TR/html53/tabular-data.html#attributes-common-to-td-and-th-elements

Mozilla’s documentation lists it with “should no longer be used” mark:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/td#attr-align

I think there is no need to add a configuration option as the CSS `text-align` property should be supported by all major browsers:
https://caniuse.com/mdn-css_properties_text-align

gaige:
@waylan Any thoughts about my comments and/or @mitya57 comments?  Anything I can do to help move this forward?

waylan:
> current PHP implementation has 2 methods of doing alignment

That helps tremendously. Quite frankly I was surprised they were still using the `align` attribute. That said, technically, the original Markdown rules require the output to be XHTML. Of course, XHTML like syntax is a valid form of HTML5 so we mostly ignore the XHTML output requirement except that the tags take on that form (`<hr />` as opposed to `<hr>`). No doubt this is where the current default of using the `align` attribute comes from.

We may still have some users who are expecting the older formats and are using the related doctypes in their documents. Would changing this cause their pages to no longer be valid? True XHTML (not XHTML served as HTML) is especially problematic in this regard and can cause a browser to crash hard if it is invalid (i don't recall if `style` attributes are valid or not). If there is any chance that some users are doing so (and I suspect there is), then we would need to have a configuration option to allow users to select there preferred option if we make any changes at all.

Another consideration are (X)HTML sanitizers. For example, GitHub's sanitizer strips all `style` attributes out of all user supplied content on their site(s). Therefore, they are still [using][1] `align` attributes in their output. True, our output is not being run through GitHub's sanitizer, but it certainly could be run through similarly configured sanitizers by some of our users.

Given the above, I think a configuration option for the extension makes the most sense. I think we could probably change the default and place a big warning in the release notes informing users that if they want/need `align` attributes rather than `style` attributes, then they need to set a custom option on the extension.

> I'm not entirely sure where to implement the test, although it appears test_syntax/extensions/test_tables might be a home for it.

Yes, that is correct, In fact, the existing tests should be moved there as part of this change. Our policy has been to move any related tests to the new system when the code related to those tests is updated.

[1]: https://github.github.com/gfm/#example-199

markdown/extensions/tables.py
waylan: Yes, I agree with all three changes here. Add the space and the semicolon and use an f-string, which is the most performant way to combine strings.

gaige:
@waylan Changes have been made per discussion. Please let me know if there are any other changes.  I made a notation about the change, but perhaps there's either a better place to warn about it, or somewhere else you'd prefer, please let me know.

Also, checklinks is failing, but not not in files that I've been working in. Not sure what you want to do about that, but those are the only failing checks at this point.

docs/change_log/index.md
```diff
@@ -5,6 +5,9 @@ Python-Markdown Change Log
 
 Under development: version 3.3.4 (a bug-fix release).
 
+* Use `style` attribute in tables for alignment instead of `align` for better CSS
+  inter-operation. The old behavior is available by setting `use_align_attribute=True` when
+  adding the extension.
```
waylan: As this is a backward incompatible change, this can't be included in a point release (see the [Contributing Guide][0]). It will need to wait until the next MINOR release (version 3.4.0) and the note will go in the Backward Incompatible Changes section of a separate document (see the release notes for [version 3.3][1] for an example. The note can then be more lengthy and perhaps include an example of how to set the config option to restore the old behavior.

[0]: https://python-markdown.github.io/contributing/#versions
[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/release-3.3.md

tests/extensions/extra/tables.html
```diff
@@ -38,22 +38,22 @@ <h2>Table Tests</h2>
 <table>
 <thead>
 <tr>
-<th align="left">Item</th>
-<th align="right">Value</th>
+<th style="text-align: left;">Item</th>
+<th style="text-align: right;">Value</th>
```
waylan: Any tests which have been replicated in `tests/test_syntax/extensions/test_tables.py` should be removed from this file. If that leaves this file empty, then it can be deleted long with the entry for it in [tests/test_legacy.py](https://github.com/Python-Markdown/markdown/blob/master/tests/test_legacy.py#L185)

tests/test_syntax/extensions/test_tables.py
```diff
@@ -60,3 +61,855 @@ def test_empty_cells(self):
...
+   First Header | Second Header
+   ------------ | -------------
+   Content Cell | Content Cell
+   Content Cell | Content Cell
+
+   | First Header | Second Header |
+   | ------------ | ------------- |
+   | Content Cell | Content Cell  |
+   | Content Cell | Content Cell  |
+""",
```
waylan: You can still wrap this in `self.dedent`, which is easier to read. The first line sets the amount of indentation to remove. As it is outdented from the table, it will work correctly.

Same goes for a number of other tests below.

waylan (changes requested):
This is a great start. I have pointed out a few things below.

You can ignore the failing checklinks test. We are aware of the problem and need to resolve it separately.

docs/change_log/index.md
gaige: Moved to release 3.4

tests/test_syntax/extensions/test_tables.py
gaige: I've indented/dedented where possible. One specific line was too long in a later test and because it already contains a large number of `\`s, I opted to leave it flush for clarity instead of wrapping it with a `\` at the end of the line prior.

waylan (approved):

tests/test_syntax/extensions/test_tables.py
```diff
@@ -60,3 +61,861 @@ def test_empty_cells(self):
...
+        )
+
+    def test_issue_449(self):
+        self.assertMarkdownRenders(
+            """
+Add tests for issue #449
+
+Odd backticks | Even backticks
+------------ | -------------
+``[!\\"\\#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^_`{|}~]`` | ``[!\\"\\#$%&'()*+,\\-./:;<=>?@\\[\\\\\\]^`_`{|}~]``
```
waylan: This is only 107 characters long. However, we allow up to 128 character line length. There is plenty of room to indent this.

waylan:
This looks good. I will merge this once we are ready to start development on version 3.4.

tests/test_syntax/extensions/test_tables.py
gaige: I had indented this line before I sent it in. However, running `tox` against it blew a warning on the line if I didn't split it, which is why I pulled it back flush. I'm not sure if that's a tox problem or something else, but I figured that if `tox` was complaining I should fix it. As mentioned previously, if there weren't already so many escapes in the line, I would have quoted the newline and left it indented, but that didn't seem a good idea here.

mitya57: > if there weren't already so many escapes in the line

Maybe you can use a raw string here (`r"""`) to reduce the number of escapes?

gaige: Good call. That resolved it. Code updated.

docs/change_log/release-3.4.md
```diff
@@ -0,0 +1,42 @@
...
+
+The former behavior is available by setting the setting `use_align_attribute` configuration
+option to `True` when adding the extension.
+
+For example, to configure the old `align` behavior:
+
+```python
+from markdown.extensions.tables import TableExtension
+
+markdown.markdown(src, extensions=[TableExtensioon(use_align_attribute=True)])
```
crisluengo: Typo: "TableExtensioon" -> "TableExtension"

gaige: Thanks. Not sure how that happened, I thought I'd copied and pasted that from my other window.

Issue #679, opened 2018-07-22
Supporting namespaces in extension-output


GammaSQ:
I want to implement a TeX to MathML parser as extension. It works fine if I use plain `<math></math>`-tags as extension-output. However, if I understood the specs correctly, MathML2 requires an `xmlns="http://www.w3.org/1998/Math/MathML"` attribute:
https://www.w3.org/TR/MathML2/chapter7.html
(It is never stated specifically weather the attribute is *required* or just *allowed*. MathML3 does not require this attribute in HTML-only context.)

The moment I add this attribute I get the following error:

    Traceback (most recent call last):
      File "test.py", line 3, in <module>
        print(markdown("$x^2$", extensions=[TeXExtension()]))
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/__init__.py", line 494, in markdown
        return md.convert(text)
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/__init__.py", line 380, in convert
        output = self.serializer(root)
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/serializers.py", line 282, in to_xhtml_string
        return _write_html(ElementTree(element).getroot(), format="xhtml")
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/serializers.py", line 203, in _write_html
        _serialize_html(write, root, qnames, namespaces, format)
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/serializers.py", line 173, in _serialize_html
        items.sort(key=lambda x: x[1])  # sort on prefix
    AttributeError: 'dict_items' object has no attribute 'sort'

This is a [Python3-issue](https://stackoverflow.com/questions/28039155/dict-items-object-has-no-attribute-sortreplace-min), so I changed `items.sort( ` to `items=sorted(items, `, but now I get this error:

    Traceback (most recent call last):
      File "/home/user/Documents/projects/rocketcollider/venv/lib64/python3.6/site-packages/markdown/__init__.py", line 384, in convert
        '<%s>' % self.doc_tag) + len(self.doc_tag) + 2
    ValueError: substring not found
    
    During handling of the above exception, another exception occurred:
    
    Traceback (most recent call last):
      File "test.py", line 3, in <module>
[...]

PR #1099, opened 2021-01-28
Add a small HtmlStash example to docs

crisluengo:
Adding a small HtmlStash example in the extension API documentation.

As I mentioned yesterday (https://github.com/Python-Markdown/markdown/issues/679#issuecomment-768659064), having this bit of information in the documentation would have saved me hours of Googling and experimentation.

### DIFF
--- a/docs/extensions/api.md
+++ b/docs/extensions/api.md
@@ -552,7 +552,7 @@ el = etree.Element("div")
 el.text = self.md.htmlStash.store(html)
 ```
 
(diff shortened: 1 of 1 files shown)


waylan:
@crisluengo thanks for doing the leg work on this. I know I rewrote much of it, but starting from your work was a lot easier than starting from scratch. It would have likely never been done if not for your work.

crisluengo:
@waylan Thanks for fixing my language. No need to apologize for that. Glad this bit is now documented.

PR #1102, opened 2021-02-05
Fix a case where content under admonitions weren't dettabbed properly

facelessuser:
Admonitions would not render the following case properly:

```
!!! note "Admonition"
    - Parent 1

        - Child 1
        - Child 2
```

But would render this properly:

```
!!! note "Admontion"
    - Parent 1
        - Child 1
        - Child 2
```

The problem stemmed from bad detabbing of the content, where in the first example, the child list was parsed like this:

```
  - Child 1
      - Child 2
```

This pull addresses the issue of ensuring that each block is processed properly so that content that is highly sensitive to indentation works correctly.

### DIFF
--- a/docs/change_log/release-3.3.md
+++ b/docs/change_log/release-3.3.md
@@ -101,6 +101,7 @@ The following bug fixes are included in the 3.3 release:
 * Fix unescaping of HTML characters `<>` in CodeHilite (#990).
 * Fix complex scenarios involving lists and admonitions (#1004).
 * Fix complex scenarios with nested ordered and unordered lists in a definition list (#918).
+* Fix corner cases with lists under admonitions.
 
 [spec]: https://www.w3.org/TR/html5/text-level-semantics.html#the-code-element
 [fenced_code]: ../extensions/fenced_code_blocks.md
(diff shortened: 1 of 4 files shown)


facelessuser:
I'm ignoring the "checklinks" failure. I feel that should be fixed in a separate issue.

waylan:
I think you have an error in your first example. As is, it renders to 

```html
<div class="admonition note">
<p class="admonition-title">Admonition</p>
</div>
<ul>
<li>
<p>Parent 1</p>
<ul>
<li>Child 1</li>
<li>Child 2</li>
</ul>
</li>
</ul>
```

.. which is correct, The `- Parent 1` line is not indented by four spaces and is therefore not a child of the admonition. Therefore it gets parsed as a list outside of the admonition as it should.

However, if you use proper indentation, like this:

```markdown
!!! note "Admonition"
    - Parent 1

        - Child 1
        - Child 2
```

Then we get the rather strange output:

```html
<div class="admonition note">
<p class="admonition-title">Admonition</p>
<ul>
<li>
<p>Parent 1</p>
<ul>
<li>Child 1<ul>
<li>Child 2</li>
</ul>
</li>
</ul>
</li>
</ul>
</div>
```

And yes that is a bug which needs fixing. So, thank you.

facelessuser:
Sorry, that was a formatting issue for the comment, not the actual test or failure.

facelessuser:
I've corrected the comment in the pull, but the issue is valid.

PR #1115, opened 2021-03-09
Allow markdown to be null

Ochmar:
Sometimes (e.g. in Django) we want markdown field to be null, in this scenario, when we try to delete object with empty markdown (and rendered markdown field) we get

> File "/usr/local/lib/python3.7/site-packages/django/core/handlers/exception.py", line 47, in inner response = get_response(request) File "/usr/local/lib/python3.7/site-packages/django/core/handlers/base.py", line 181, in _get_response response = wrapped_callback(request, *callback_args, **callback_kwargs) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/options.py", line 614, in wrapper return self.admin_site.admin_view(view)(*args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/utils/decorators.py", line 130, in _wrapped_view response = view_func(request, *args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/views/decorators/cache.py", line 44, in _wrapped_view_func response = view_func(request, *args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/sites.py", line 233, in inner return view(request, *args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/utils/decorators.py", line 43, in _wrapper return bound_method(*args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/utils/decorators.py", line 130, in _wrapped_view response = view_func(request, *args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/options.py", line 1735, in changelist_view response = self.response_action(request, queryset=cl.get_queryset(request)) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/options.py", line 1402, in response_action response = func(self, request, queryset) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/actions.py", line 40, in delete_selected modeladmin.delete_queryset(request, queryset) File "/usr/local/lib/python3.7/site-packages/django/contrib/admin/options.py", line 1103, in delete_queryset queryset.delete() File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 747, in delete deleted, _rows_count = collector.delete() File "/usr/local/lib/python3.7/site-packages/django/db/models/deletion.py", line 436, in delete sender=model, instance=obj, using=self.using File "/usr/local/lib/python3.7/site-packages/django/dispatch/dispatcher.py", line 179, in send for receiver in self._live_receivers(sender) File "/usr/local/lib/python3.7/site-packages/django/dispatch/dispatcher.py", line 179, in <listcomp> for receiver in self._live_receivers(sender) File "/usr/local/lib/python3.7/site-packages/simple_history/models.py", line 491, in post_delete self.create_historical_record(instance, "-", using=using) File "/usr/local/lib/python3.7/site-packages/simple_history/models.py", line 526, in create_historical_record history_instance.save(using=using) File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 754, in save force_update=force_update, update_fields=update_fields) File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 792, in save_base force_update, using, update_fields, File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 895, in _save_table results = self._do_insert(cls._base_manager, using, fields, returning_fields, raw) File "/usr/local/lib/python3.7/site-packages/django/db/models/base.py", line 935, in _do_insert using=using, raw=raw, File "/usr/local/lib/python3.7/site-packages/django/db/models/manager.py", line 85, in manager_method return getattr(self.get_queryset(), name)(*args, **kwargs) File "/usr/local/lib/python3.7/site-packages/django/db/models/query.py", line 1254, in _insert return query.get_compiler(using=using).execute_sql(returning_fields) File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1396, in execute_sql for sql, params in self.as_sql(): File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1341, in as_sql for obj in self.query.objs File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1341, in <listcomp> for obj in self.query.objs File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1340, in <listcomp> [self.prepare_value(field, self.pre_save_val(field, obj)) for field in fields] File "/usr/local/lib/python3.7/site-packages/django/db/models/sql/compiler.py", line 1291, in pre_save_val return field.pre_save(obj, add=True) File "/usr/local/lib/python3.7/site-packages/markdownfield/models.py", line 85, in pre_save extension_configs=EXTENSION_CONFIGS File "/usr/local/lib/python3.7/site-packages/markdown/core.py", line 387, in markdown return md.convert(text) File **"/usr/local/lib/python3.7/site-packages/markdown/core.py", line 248, in convert if not source.strip(): AttributeError: 'NoneType' object has no attribute 'strip'**

Proposed change fixes it without any downsite I can think of

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -245,7 +245,7 @@ class Markdown:
         """
 
         # Fixup the source text
-        if not source.strip():
+        if not source or (source and not source.strip()):
             return ''  # a blank unicode string
 
         try:


waylan:
The problem is that we have historically only accepted Unicode strings (`str` in Python 3). Back when we supported Python 2, we would raise an error if a byte string (then `str`) was received. And that was Python's native string type at the time. If we won't even accept a byte string, it doesn't make sense to accept something which is not a string at all.

Based on the traceback, it would appear you are using [dmptrluke/django-markdownfield][1]. I'm assuming you also have your field set with `null=True`. Any reason you can't do `blank=True` instead, which would use an empty string and avoid this issue? 

Actually, if the markdownfield allows `null=True`, then it should check for a null value and transform to an empty string before passing to Markdown. I would consider this a bug in markdownfield if it doesn't do that.

[1]: https://github.com/dmptrluke/django-markdownfield

Ochmar:
Actually I did not think about Python 2 and byte strings at all, so I guess doing blank=True is a way to go. I made some tweaks in my code and actually can work without this commit.

Thanks!

PR #1116, opened 2021-03-14
Add toc level class

summersz:
Add a toc level class attribute to each list item to allow for individual level styling, e.g. class="toc1", class="toc2", etc.


waylan:
My initial reaction was to push back on this. But in the end I don't really have a good reason to not allow this, especially if someone is willing to do the work. However, before this can be merged, all of the failing tests need to be updated to include the new class in the expected output. Also, a comment needs to be added to the change log.

summersz:
@waylan Please let me know if this is acceptable or if any further work is required. Thanks

waylan:
Note that some TOC related tests have been moved to the new framework and now reside in [tests/test_syntax/extensions/test_toc.py][1]. However, some older TOC tests have not yet been updated and still reside in [tests/test_extensions.py][2]. If any of the old tests need to be modified, then they should be moved to the new location and updated to make use of the [`assertMarkdownRenders`][3] (and `dedent` if needed) methods.

The same applies to the `txt` and `html` files in `tests/extensions/`. Any tests which need to be modified should be moved to `[tests/test_syntax/extensions/test_toc.py` and refactored to use the newer framework.

To summarize, we should not be modifying any tests which are not under `tests/test_syntax`. Instead, those tests should be deleted and new tests should be created under `tests/test_syntax` to replace them.

[1]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_syntax/extensions/test_toc.py
[2]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_extensions.py
[3]: https://python-markdown.github.io/test_tools/#markdowntest_toolstestcase

docs/change_log/release-3.3.md
```diff
@@ -87,6 +87,9 @@ The following new features have been included in the 3.3 release:
   `markdown.extensions.headerid.slugify_unicode` to the `slugify` configuration option
   to use the new behavior.
 
+* A table of contents level class has been added to the [Table of Contents](../extensions/toc.md)
+  extension, e.g. class="toc1". This allows for individual styling for each table of contents level.
+
```
waylan, line 92: Version 3.3 has already been released so any new changes should not be documented here. As this is a new feature, it would need to be in a new release (presumably 3.4). The appropriate file will need to be added along with an entry in the [change log][1] which links to it.

[1]: https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/index.md

waylan: Note that the 3.4 release notes now exist. If the requested changes are made, I'll review and merge this.

waylan:
In addition to the requested changes above, this should be off by default and have a config option to turn it on.

That said, I don't see this as necessary and am not interested in spending any time working on it myself. If the requested changes are not made soon, I am inclined to simply close this.

waylan:
In addition to the requested changes above, this should be off by default and have a config option to turn it on.

That said, I don't see this as necessary and am not interested in spending any time working on it myself. If the requested changes are not made soon, I am inclined to simply close this.

waylan:
I am closing this due to inaction. It can be reopened if/when the requested changes are made.

PR #1133, opened 2021-05-22
Add custom Pygments lexer support for Codehilite

CorvusPrudens:
Custom Pygments lexers can now be passed as a keyword argument -- `custom_lexers=` -- to the `CodeHilite` and `CodeHiliteExtension` classes. The lexer must be a class that inherits from `RegexLexer` or other valid Pygments lexers [as described here](https://pygments.org/docs/lexerdevelopment/). Simply pass the class inside a list -- `CodeHiliteExtension(custom_lexers=[CustomLexer])` -- and it will be instantiated with whatever extra keyword arguments you passed in like a standard lexer. Naturally, multiple lexers can be put in the list.

This is obviously a very niche use-case, so I wouldn't be gutted if it isn't worth the trouble. I'm also not sure I updated the right file for the change log.

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -6,6 +6,7 @@ Python-Markdown Change Log
 Under development: version 3.3.5 (a bug-fix release).
 
 * Make the `slugify_unicode` function not remove diacritical marks (#1118).
+* Add support for custom Pygments lexers with Codehilite.
 
 Feb 24, 2021: version 3.3.4 (a bug-fix release).
 
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -43,6 +43,21 @@ def parse_hl_lines(expr):
         return []
 
 
+def get_custom_lexer_by_name(_alias, custom_lexers, options):
+    """Provides similar functionality to get_lexer_by_name, but
+    for custom lexers.
+
+    Returns the corresponding lexer if it exists, otherwise raising
+    a ValueError.
+    """
+
+    for lex in custom_lexers:
+        if _alias in lex.aliases:
+            return lex(**options)
+
+    raise ValueError(f'Lexer with alias "{_alias}" was not found.')
+
+
 # ------------------ The Main CodeHilite Class ----------------------
 class CodeHilite:
     """
@@ -69,6 +84,10 @@ class CodeHilite:
 
     * lang_prefix: Prefix prepended to the language when `use_pygments` is `False`.
       Default: "language-".
+
+    * custom_lexers: An optional list of custom Pygments lexers. The lexer should
+      be an extension of a valid Pygments lexer class such as `RegexLexer`.
+      Default: [].
 
     Other Options:
     Any other options are accepted and passed on to the lexer and formatter. Therefore,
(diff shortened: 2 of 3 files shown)


waylan:
So I suppose the use case for this is to allow a user to provide support for a 'language' which pygments does not support. Alternatively, a user could develop their own custom (different) support for a pre-existing 'language'. You're right that these are edge cases which I wouldn't have spent any time on. But since you've already done the work... I'll have to think about it.

facelessuser:
@CorvusPrudens I'm not sure what this is adding. Custom lexers are already possible as Pygments supports them, and Python Markdown will just call Pygments. If they are installed properly, they'll be used. This is a real example that uses some very simple [custom lexers](https://github.com/facelessuser/pymdown-lexers ) I use in my own documents:

```py
>>> markdown.markdown('\n```critic-markup\ntest {++add++}\n```\n', extensions=["fenced_code","codehilite"])
'<div class="codehilite"><pre><span></span><code>test<span class="w"> </span><span class="s">{++add++}</span>\n</code></pre></div>'
```

CorvusPrudens:
@facelessuser Yes you're absolutely right. I find the process of adding a plugin far more obtuse than simply passing in a class, however. The behavior of my PR is similar to how sphinx handles custom lexers, for example. 

I'll be deploying this on a server eventually so I'm shooting for simplicity of setup.

facelessuser:
> @facelessuser Yes you're absolutely right. I find the process of adding a plugin far more obtuse than simply passing in a class, however. The behavior of my PR is similar to how sphinx handles custom lexers, for example.

In what way is it obtuse? What part of it makes it difficult?

Installing a custom lexer pack doesn't seem that bad.

I ask this simply because there is already a way in which to add custom lexers, and it works, so it would be helpful to understand why Python Markdown should add more complexity (even if it isn't that much). What problems is this overcoming?

An alternative approach, just because it is subjectively preferred may not be a good reason, but if there is a real problem that is being overcome, it may make more sense. I'm just struggling to understand what this "problem" could be as I use custom lexers.

> I'll be deploying this on a server eventually so I'm shooting for simplicity of setup.

How does this simplify things for you?

CorvusPrudens:
@facelessuser I may be misunderstanding, but don't you have to add plugins with pip? Assuming that's true, then it would be one more thing to add to whatever environment I'm setting up. It seems simpler and more flexible to me to have that functionality right out of the box. Of course, it's true that it adds complexity, so feel free to close the pr. It's also relatively useless if plugins can be added dynamically. 

I wouldn't have bothered with a pr but I'll be using this code myself so I said why not haha.

facelessuser:
> @facelessuser I may be misunderstanding, but don't you have to add plugins with pip? Assuming that's true, then it would be one more thing to add to whatever environment I'm setting up. It seems simpler and more flexible to me to have that functionality right out of the box. Of course, it's true that it adds complexity, so feel free to close the pr. It's also relatively useless if plugins can be added dynamically.

That is how I do it 🤷🏻 . And yes it is one more step, but I don't think it is an unreasonable step. It's just the way Pygments is designed. With that said, I'm not saying I'm against it per se, I'm kind of a +0 and would personally just defer to @waylan's decision.

CorvusPrudens:
Essentially, to clarify:

This feature would improve code portability across environments and simplify the process of adding custom Pygments lexers. In my case, I have a little language I made that I want syntax highlighting for.

waylan:
As a reminder, we require Markdown extensions to be installed as Python packages. Therefore, making an exception for Pygments is counter intuitive. Just install them the way Pygments expects. The only reason I didn't close this immediately is because I figured that must not be possible, otherwise, why would someone even both with this.

PR #1149, opened 2021-07-01
Remove SO attributions to sidestep licensing greyness

### LINKED ISSUE #1148: Licensing issue with Stackoverflow's CC-BY-SA License
Hi, thank you for this library. 

We have recently been FOSS audited and this package was flagged with potential licensing issues. 

Code copied/distributed and/or adapted from [Stackoverflow requires](https://stackoverflow.com/help/licensing):
1. attribution and
2. **licensing under the same license as the original ([CC-BY-SA-3.0](https://creativecommons.org/licenses/by-sa/3.0/) or [CC-BY-SA-4.0](https://creativecommons.org/licenses/by-sa/4.0/)).** 
This repositories MIT License is not compatible with CC-BY-SA.

The code that is causing issue:

https://github.com/Python-Markdown/markdown/blob/faf73b08bb984669fcba044259f8d67fbfed69ab/markdown/util.py#L103
https://github.com/Python-Markdown/markdown/blob/faf73b08bb984669fcba044259f8d67fbfed69ab/markdown/util.py#L161

Consider changing the implementation of the function such that the attribution isn't required.

jarekwg:
Closes #1148

StackOverflow code uses copyleft licenses (CC-BY-SA-3.0 or CC-BY-SA-4.0), requiring that any usage be both attributed, and the license bubbled up.

In this case, we've got two functions that are basically standard python patterns/snippets, however by including an attribution link (which would sensibly seem like the honest thing to do), and having these snippets copypasted, the rules are technically being broken. The law uses broad brush strokes, so common sense arguments don't really apply unless the matter is in front of a judge. ;/

This PR, as silly as it seems, allays any of this greyness.

e: lemme know if my import sorter kicking in here is breaking the repo rules/standards.

Thanks for your time!

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -21,12 +21,13 @@ License: BSD (see LICENSE.md for details).
 
 import re
 import sys
-from collections import namedtuple
-from functools import wraps
 import warnings
 import xml.etree.ElementTree
-from .pep562 import Pep562
+from collections import namedtuple
+from functools import wraps
 from itertools import count
+
+from .pep562 import Pep562
 
 try:
     from importlib import metadata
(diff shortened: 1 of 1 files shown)


waylan:
I guess the lesson here is to avoid using SO for anything. Don't visit their site and certainly don't link to them.

jarekwg:
> I guess the lesson here is to avoid using SO for anything. Don't visit their site and certainly don't link to them.

Yeah basically.. I went down a rabbit hole of reading for this and saw they planned to swap everything to MIT back in 2016, but ended up backing out for some lame reason..

Thanks for merging! :)

facelessuser:
Yeah, I've used this same approach in some of my repos and will most likely have to strip out references to SO. I've only ever used basic examples like how to override PyYaml or quite literally this deprecation example. None of which are algorithms or things worthy of actual copyright, but just basic explanations of how to do something fairly basic.

>  I went down a rabbit hole of reading for this and saw they planned to swap everything to MIT back in 2016

Yeah, this is what they should have done as the current model is toxic.

PR #1151, opened 2021-07-03
document footnote's need for reset() between multiple runs

peterkaminski:
The footnote docs page doesn't mention the need to run `reset()` between multiple runs of the `markdown.Markdown` class.

This PR adapts and adds language taken from the `extensions/api.md` page to explain what to do and why.

### DIFF
--- a/docs/extensions/footnotes.md
+++ b/docs/extensions/footnotes.md
@@ -103,13 +103,21 @@ markdown.markdown(some_text, extensions=['footnotes'])
 Resetting Instance State
 -----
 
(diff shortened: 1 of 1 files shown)


waylan:
This assumes that the user will never want to have footnotes from multiple documents combined. However, that is why reset is an option. Sometimes a user may want to not reset. For example, the homepage of a blog might include the content from multiple documents. By not calling reset, all of the footnotes get included and they all have unique links/references. We should document both use cases, not just the one.

peterkaminski:
> We should document both use cases, not just the one.

Excellent point.  I have reworked the doc to include both use cases.

Issue #1154, opened 2021-07-15
Incorrect link 404 in admonition.md


IvanReznikov:
Broken link leads to 404 in admonition.md file

PR #1155, opened 2021-07-15
fixed link #1154

IvanReznikov:
Fixed link for https://github.com/Python-Markdown/markdown/issues/1154


waylan:
It would appear that the checklinks linter is [broken][1] in some way. As checking links is directly related to this change, I would like to get the link checker fixed and see this passing before accepting this change.

[1]: https://github.com/Python-Markdown/markdown/pull/1155/checks?check_run_id=3115635474

Issue #1160, opened 2021-07-27
Bug: nl2br and toc extensions removes line that come immediatly after [toc]


cbeauchesne:
## Issue

On this specific markdown (pay attention to the unique line return after the toc marker)

```markdown
[TOC]
where am I?
```

The `nl2br` and `toc` extensions used together remove the second line : 

```python
import markdown

parser = markdown.Markdown(extensions=['toc', 'nl2br'])

print(parser.convert("[TOC]\nWhere am I?",))
```

prints : 

```html
<div class="toc">
<ul></ul>
</div>
```

----

* Add another `\n` between the marker and the seconde line, and it's ok. 
* Do not use `nl2br`, and it's also ok (in this case, the toc is not created, but I guess it's the expected behavior, as `[toc]` should be a block).

## Expected behavior 

From what I understand, and regarding previous versions (tested on `Markdown==3.1.1`), the result should be:
 
```html
<p>[TOC]<br>
x</p>
```

## Context 

* Python 3.9.6 and Python 3.7.3
* Markdown==3.3.4
* both 5.10.25-linuxkit and MacOS

facelessuser:
While this may be considered a bug, you really shouldn't make the `[TOC]` part of another paragraph. Having `[TOC]` in its own paragraph would resolve your issue, but, technically, I would consider this a bug.

cbeauchesne:
Hi @facelessuser,

 > you really shouldn't make the [TOC] part of another paragraph

Yes I know 😅 , I expect this pattern not being populated as a toc.

I manage a wiki with more than 80,000 users and 1,000,000 documents, and almost all of them don't know markdown, and write it with a trial/error process. Any inconsistency in our markdown parser end up with terrible hacks in our base :/

A lead for the root cause: if I use `TocTreeprocessor.iterparent` and `TocTreeprocessor.replace_marker` from 3.3.1, I do not have the issue.

I'll try to dig it the next days.

cbeauchesne:
> A lead for the root cause: if I use TocTreeprocessor.iterparent and TocTreeprocessor.replace_marker from 3.3.1, I do not have the issue.

That's not true, sorry, it was a version I used locally. 

----

So, the answer. On line 198: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/toc.py#L195-L198

`c` is an element, and we check that `c.text == marker`. 

[...]

PR #1161, opened 2021-07-27
Better toc detection 

cbeauchesne:
Fix for issue #1160

See issue, and code comment for more details.

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -6,7 +6,7 @@ Python-Markdown Change Log
 Under development: version 3.3.5 (a bug-fix release).
 
 * Make the `slugify_unicode` function not remove diacritical marks (#1118).
-* Fix toc detection when used with nl2br extension (#1160)
+* Fix `[toc]` detection when used with `nl2br` extension (#1160)
 
 Feb 24, 2021: version 3.3.4 (a bug-fix release).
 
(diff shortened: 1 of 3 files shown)


facelessuser:
I'm not familiar with the "checklinks" failures, so I would ignore them as it seems to be a failure with a recent update to the checker.

I would, however, add a note in the changelog for the fix in order to resolve the "require_changelog" failure.

tests/test_extensions.py
```diff
@@ -310,6 +310,17 @@ def testRE(self):
...
 
+class TestTOCAndNl2br(unittest.TestCase):
+
+    def setUp(self):
+        self.md = markdown.Markdown(extensions=['toc', 'nl2br'])
+
+    def testNotToc(self):
+        """ Test TOC with a single newline is not populated. """
+        text = '[TOC]\ntext'
+        self.assertEqual(self.md.convert(text), '<p>[TOC]<br />\ntext</p>')
```
waylan: In addition to the points made by @facelessuser, the new test case should be using the newer [markdown.test_tools.TestCase][1] and be located in [tests/test_syntax/extensions/test_toc.py][2]

[2]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_syntax/extensions/test_toc.py

PR #1165, opened 2021-08-01
Require a local-part to be a valid email address in AUTOMAIL_RE.

nzlosh:
Valid email addresses require a _local-part_ before the `@` followed by _domain_.  This PR updates the `AUTOMAIL_RE` to use the `+` quantifier to require 1 or more characters in the _local-part_ instead of the `*` quantifier which allows 0 characters.

The current regex allows the invalid addresses in the form `<@domain>`.  My particular use case is formatting messages in Slack, where mentioned users take the form `<@Uxxxxx>`.  markdown is incorrectly processing the mention string form as an email address.  I still want markdown's automail feature for processing email addressing in the text.

The below code demonstrates the current and new behaviour:

```
>>> AUTOMAIL_RE = r'<([^<> !]*@[^@<> ]*)>'
>>> print(re.match(AUTOMAIL_RE, "<@localhost>"))
<re.Match object; span=(0, 12), match='<@localhost>'>
>>> print(re.match(AUTOMAIL_RE, "<bob@localhost>"))
<re.Match object; span=(0, 15), match='<bob@localhost>'>

>>> AUTOMAIL_RE = r'<([^<> !]+@[^@<> ]*)>'
>>> print(re.match(AUTOMAIL_RE, "<@localhost>"))
None
>>> print(re.match(AUTOMAIL_RE, "<bob@localhost>"))
<re.Match object; span=(0, 15), match='<bob@localhost>'>
```


mitya57:
Make it makes sense to also replace `*` with `+` in the domain part? With your fix it still allows for addresses like `bob@`:

```python
>>> AUTOMAIL_RE = r'<([^<> !]+@[^@<> ]*)>'
>>> re.match(AUTOMAIL_RE, "<bob@>")
<re.Match object; span=(0, 6), match='<bob@>'>
```

nzlosh:
I've updated the regex to require both `local-part` and `domain`.

mitya57 (approved):

nzlosh:
I've added tests for the missing local-part and domain.  I noticed that the `<>` characters are encoded in html form `&lt;&gt;` for the missing local-part, but not for the missing domain part.  The tests mirror this behaviour, but I don't know if it's expected behaviour or not.

docs/change_log/release-3.3.md
```diff
@@ -102,6 +102,7 @@ The following bug fixes are included in the 3.3 release:
 * Fix complex scenarios involving lists and admonitions (#1004).
 * Fix complex scenarios with nested ordered and unordered lists in a definition list (#918).
 * Fix corner cases with lists under admonitions.
+* Automail does not treat missing local-part or domain as a valid email addresses (#1165).
```
waylan: Version 3.3 has already been released. This note does not go here. Instead it should be added to the notes for 3.3.5 at [`docs/change_log/index.md`](https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/index.md)

waylan (changes requested):
All new tests should use the [new testing framework][1]. It appears that no changes have been made to the autolinks since introducing the new framework so we don't have a file for that yet. One should be added at `tests/test_syntax/inline/autolinks.py`. You only need to include the tests for the behaviors relevant to this PR. However, you may add additional tests if you would like.

Regarding the differing behaviors, I see that as a problem. If any part of the autolink does not match, then the behavior should be the same.

docs/change_log/release-3.3.md
waylan: Also, they are called "[Automatic Links][1]", not "Automail". Changing that will address the [failing spellchecker test][2]. You can ignore the failing `checklinks` test as it is broken.

[1]: https://daringfireball.net/projects/markdown/syntax#autolink
[2]: https://github.com/Python-Markdown/markdown/pull/1165/checks?check_run_id=3224274586

nzlosh:
Thanks for the pointers, I've updated the PR with the recommended changes.

tests/test_syntax/inline/test_autolinks.py
```diff
@@ -0,0 +1,55 @@
...
+        self.assertMarkdownRenders(
+            """Missing local-part <@domain>""",
+            """<p>Missing local-part &lt;@domain&gt;</p>"""
+        )
+
+    def test_invalid_email_address_domain(self):
+        self.assertMarkdownRenders(
+            """Missing domain <local-part@>""",
+            """<p>Missing domain <local-part@></p>"""
+        )
```
waylan: The different behavior still bothers me so I checked [Babelmark][1] and this matches the reference implementation. Although, I suspect that is a bug. In fact, no other implementation has this behavior.

I suspect the issue is elsewhere in the code and could be classified as a separate bug. However, I'm not comfortable codifying incorrect behavior in a test, so we either need to fix the issue here or remove this test. I'd rather fix the issue.

[1]: https://johnmacfarlane.net/babelmark2/?normalize=1&text=%3C%40domain%3E%0A%0A%3Clocal-part%40%3E

waylan: I see the issue. Note the inline HTML regex:

https://github.com/Python-Markdown/markdown/blob/c1724406ae518a4b295406cdb076a5804fee3ef3/markdown/inlinepatterns.py#L166

That correctly requires the first character after the opening bracket to be a letter or a slash (`[a-zA-Z/]`). Of course, `<@domain>` fails that test and is not treated an raw HTML and gets escaped. However, as anything other than angle brackets (`[^<>]`) are accepted for any character after the first character, then `<local-part@>` is treated as raw HTML and not escaped. And we can't just insert a `@` in the disallowed characters (`[^<>@]`) as the following is valid HTML: `<a href="mailto:email@example.com">email</a>`. Of course, an `@` outside of quotes could probably be disallowed, but that would require a much more sophisticated regex.

Outside of an easy fix, we may just leave it as-is. Thoughts anyone?

waylan: A simple solution may be to disallow some characters before a space character, but allow anything after an optional space. Then we don't need to get into the specifics of whether things are wrapped in quotes or not. We would simply be verifying that the tag name contains only valid characters. I don't see any reason to be concerned over whether any attributes are valid or not.

Perhaps something like this: `\/?[a-zA-Z][^<>@ ]*( [^<>]*)?`.

Note that I also moved the optional opening slash out of the first character That way we are still confirming that the first character after the opening slash is alpha only. For reference see the [HTML spec][1]. Of course, [XML][2] is more complicated, but XHTML would be the intersection of the two, so I think we are okay here. 

We could probably make some additional changes, but that is another issue. The above should address this issue alone. That is, if it doesn't break any of our existing tests.

[1]: https://html.spec.whatwg.org/multipage/syntax.html#syntax-tag-name
[2]: https://www.w3.org/TR/xml/#NT-S

mitya57: > A simple solution may be to disallow some characters before a space character, but allow anything after an optional space.

This sounds like the best thing to do.

docs/change_log/index.md
```diff
@@ -7,6 +7,7 @@ Under development: version 3.3.5 (a bug-fix release).
 
 * Make the `slugify_unicode` function not remove diacritical marks (#1118).
 * Fix `[toc]` detection when used with `nl2br` extension (#1160)
+* Automatic Links do not treat missing local-part or domain as a valid email address (#1165).
```
mitya57: I would suggest a small wording change:

```suggestion
* Automatic Links do not treat an email address with missing local part or domain as valid (#1165).
```

nzlosh: The rewording suggests that `<@Uxxxx>` was intended to be an email address.  How about:

`* Automatic Links improve email address validation when processing @ inside HTML tags <>.`

mitya57: That's even better.

waylan (approved):

PR #1166, opened 2021-08-04
Don't process shebangs in codehilite when processing fenced code

### LINKED ISSUE #1156: Don't eat "pathless shebang lines" that could be actual Rust code
Howdy!  The following syntax has surprising results with Python-Markdown, because the markdown processor tries to hide the "shebang line without path":

```
#!test
```

The feature causing this problem is https://python-markdown.github.io/extensions/code_hilite/#shebang-no-path, which attempts to create an empty code block.  I think that structurally you should be able to represent anything you like inside a code block, including syntax that could be a pathless shebang line, and thus the correct solution is to either remove the "shebang-no-path" feature, or at least make it optional (probably default-off), which makes the above string not appear properly in code blocks.

See https://johnmacfarlane.net/babelmark2/?text=%60%60%60%0A%23!test%0A%60%60%60%0A for what other implementations do, and https://github.com/zulip/zulip/issues/18591 for an example user experience with this feature (the original reproducer was actual Rust code).

facelessuser:
Fixes #1156

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -69,3 +69,6 @@ ENV/
 
 # MkDocs documentation
 site/
+
+# Mac files
+.DS_Store
--- a/docs/change_log/release-3.3.md
+++ b/docs/change_log/release-3.3.md
@@ -102,6 +102,7 @@ The following bug fixes are included in the 3.3 release:
 * Fix complex scenarios involving lists and admonitions (#1004).
 * Fix complex scenarios with nested ordered and unordered lists in a definition list (#918).
 * Fix corner cases with lists under admonitions.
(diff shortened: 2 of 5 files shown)


facelessuser:
Lint test can be ignored for now I guess. I have a fix for the failure in https://github.com/Python-Markdown/markdown/pull/1167. I'm unconcerned with caching and think having something not broken is more important in the short term than figuring out how to utilize npm caching correctly.

facelessuser:
GitHub was having lagging response time, so checklinks failed after the rebasing on the tip. Re-ran tests once GitHub was more responsive and tests are now passing. `require_changelog` bot (or action) seems to be hanging, but changelog was updated 🤷🏻.

PR #1152, opened 2021-07-05
ci(workflow): add cache to workflows using actions/setup-node

oscard0m:
## Description

Add `cache` to workflows using `actions/setup-node`

## Context

`setup-node` GitHub Action just released a new option to add cache to steps using it.

You can find the details here: https://github.blog/changelog/2021-07-02-github-actions-setup-node-now-supports-dependency-caching/

Fix #1153

---

🤖 This PR has been generated automatically by [this octoherd script](https://github.com/oscard0m/octoherd-script-add-cache-to-node-github-action), feel free to run it in your GitHub user/org repositories! 💪🏾

PR #1167, opened 2021-08-04
Fix cache error in CI

facelessuser:
If cache is desired, we should look into this in the future and do it
properly.


facelessuser:
#1155 can be rebased once this change is in

waylan:
Ah, that was added in #1152. It seemed to be working at the time. After merging it I forgot all about and never connected it with the current error. In any event, yes, let's back this out. Personally, I could care less about whether cache is enabled. If someone does care about it, then they need to provide a working solution that is not so brittle.

PR #1168, opened 2021-08-06
Add option for string output to md.convertFile

UtilityHotbar:
convertFile is unable to return a string. This patch should add the ability to remedy that.

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -308,7 +308,7 @@ class Markdown:
         Keyword arguments:
 
         * input: File object or path. Reads from stdin if `None`.
-        * output: File object or path. Writes to stdout if `None`.
+        * output: File object or path. Writes to stdout if `None`. Set to `str` to return a string containing generated html.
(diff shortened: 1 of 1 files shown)


waylan:
The `convertFile` method is only intended to handle the basic case. For anything else, it is assumed that the user will open her own files for reading any writing. In fact, we added the method back in Python 2.x when reading and writing with Unicode was less straightforward. Today, the method only exists for consistency. If I was to make any changes, I would simply remove it altogether.

Given the above, it is my suggestion to not accept this. Anyone else have any thoughts?

UtilityHotbar:
I don't see the harm in providing a utility function that handles an entire file automatically. However, I do agree that either such a function should be fully featured or it should not exist to prevent confusion. Eventually I solved the problem by making it write to a BytesIO object then extracting the data from that, but that feels like a workaround rather than a solution. As a further point, the top level utility function does not even return the Markdown object returned by the class method, making it even more of a strange implementation.

PR #1177, opened 2021-08-23
lxml support

arkadesOrg:
Some projects already rely on lxml features internally, but still want to make use of the python-markdown package. So it can be important to use the third-party implementation with python-markdown if one wants to deeper integrate with markdown library.

I did outsource the etree import mechanism into markdown/etree module as a try-except block.

### DIFF
--- a/markdown/blockparser.py
+++ b/markdown/blockparser.py
@@ -19,7 +19,7 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
-import xml.etree.ElementTree as etree
+from . import etree
 from . import util
 
 
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -32,7 +32,7 @@ as they need to alter how markdown blocks are parsed.
 
 import logging
 import re
-import xml.etree.ElementTree as etree
+from . import etree
 from . import util
 from .blockparser import BlockParser
 
--- a/markdown/etree.py
+++ b/markdown/etree.py
@@ -1,0 +1,9 @@
+try:
+  import lxml.etree as etree
+except:
+  import xml.etree.ElementTree as etree
+  
+def __getattr__(name):
+    """Get attribute."""
+    
+    return getattr(etree, name)
--- a/markdown/extensions/abbr.py
+++ b/markdown/extensions/abbr.py
@@ -21,7 +21,7 @@ from ..blockprocessors import BlockProcessor
 from ..inlinepatterns import InlineProcessor
 from ..util import AtomicString
 import re
-import xml.etree.ElementTree as etree
+from .. import etree
 
 
 class AbbrExtension(Extension):
--- a/markdown/extensions/admonition.py
+++ b/markdown/extensions/admonition.py
@@ -19,7 +19,7 @@ License: [BSD](https://opensource.org/licenses/bsd-license.php)
 
 from . import Extension
 from ..blockprocessors import BlockProcessor
-import xml.etree.ElementTree as etree
+from .. import etree
 import re
 
 
--- a/markdown/extensions/def_list.py
+++ b/markdown/extensions/def_list.py
@@ -17,7 +17,7 @@ License: [BSD](https://opensource.org/licenses/bsd-license.php)
 
 from . import Extension
 from ..blockprocessors import BlockProcessor, ListIndentProcessor
-import xml.etree.ElementTree as etree
+from .. import etree
 import re
 
 
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -22,7 +22,7 @@ from .. import util
 from collections import OrderedDict
 import re
 import copy
-import xml.etree.ElementTree as etree
+from .. import etree
 
 FN_BACKLINK_TEXT = util.STX + "zz1337820767766393qq" + util.ETX
 NBSP_PLACEHOLDER = util.STX + "qq3936677670287331zz" + util.ETX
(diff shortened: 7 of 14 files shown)


facelessuser:
What about 3rd party plugins interacting with a now unknown implementation? Now they aren't sure if it is an lxml or built-in XML.

arkadesOrg:
True, so the lxml-switch should go into the configuration right?

facelessuser:
You'll have a mix of 3rd party plugins. Most won't support this. TBH, I am doubtful @waylan will accept this.

What are you hoping to gain by having lxml support baked in?

facelessuser:
Also, keep in mind that @waylan  evaluated a number of parsers and such for HTML and may have evaluated lxml and passed on it for one reason or another.

arkadesOrg:
I recognized that the package architecture moved away from the global etree-variable toward a decentralized etree package import. So it became harder to switch parser without changing the python-markdown source.

I am currently working on a template engine using lxml's FunctionNamespace to extend the XPath functionality. So first I parse extended md files into etree, and then run the renderer (which also evaluates extended XPath and dpath on external resources from inside markdown, and also heavily relies on lxml's _getparent_) on it.

IMHO a custom parser configuration would be a nice feature. But I should call the branch experimental-lxml-support though, to make clear that non-standard parsers are not officially supported but possible.

I will try to come up with a reasonable Markdown config mechanism to switch parser optionally away from the python builtin parser.

waylan:
What value does this offer? We don't output the etree document to the end-user. We only use it internally and then output the rendered string. True, extensions can operate on the etree object and you can get the etree object as output using hacks, but even then, what benefit does lxml's implementation provide? 

If you want to combine the output of Markdown with an existing lxml etree object, then I would suggest running Markdown's output through lxml's parser. As a reminder, Markdown postprocessors take various actions on the rendered string which you will be missing if you make use of our internal etree object directly.

arkadesOrg:
In my case the benefit is to be able to use extended XPath evaluation from within my custom Markdown Extension. I already tried to back-and-forth parse the tree between _xml_ and _lxml_ inside the treeprocessors stage, but its horrible and just does not work, especially if I rely on references to nodes of the previous stage. In my case the best solution is to optionally switch the parser at the earliest stage possible, so the postprocessor won't miss anything but just run on lxml.

Of course I see the advantage of the builtin xml parser, most parties will be fine with lesser dependencies.
But lxml comes with extended XPath plus customizable XPath functions.
Now for example I can have a YAML data block inside Markdown and access it's values from within markdown text through {{ customFunc(/data/value1 + /data/value2) }} placeholder.

PS: thanks for both your quick and detailed answers though

waylan:
What would you do if we didn't use etree at all? I would suggest exploring that approach. For example, the codehilite extension inserts a placeholder in the etree object and stores the actual HTML in a stash. The placeholder is later swapped out for the stashed HTML in a postprocessor after the etree is rendered to a string. And highlighted code blocks are just plain HTML which would be relatively easy to insert as etree objects.

By the way, I have always been annoyed the lxml's etree object is not compatible with ElementTree. I see this as a bug with lxml and not something we need to accommodate. Therefore, I have always avoiding using it. I don't want to start now if there is any way to avoid it.

arkadesOrg:
The question of what would I do without etree is a bit out of scope. I guess I'd write a Markdown parser with ElementTree and experimental lxml-support in some of those parallel universes.
Anyway I agree that it is not the Python-Markdown-team's job to be compatible with multiple parsers.
But if I use an experimental parser through a parser-config-feature for my lxml-dependent Markdown-Extension, I will be in charge to commit ElementTree compatibility improvements to the lxml project. Basically lxml claims to be compatible with ElementTree, and not vice versa.

I'll implement a optional-parser-feature for Markdown kwargs-config-dict locally and do some further testing on it's compability. I'll let you know about another pull request.

What disadvantages/incompabilities of lxml did you encounter, except the unicode-string-thingy? I remember I did some hackish workaround locally several years ago, to make the codehilite placeholder work with lxml.

waylan:
> I guess I'd write a Markdown parser with ElementTree and experimental lxml-support

Interesting. Personally, in retrospect, I believe using ElementTree was the wrong choice for Markdown way back when. If I was to start over, I would instead generate an abstract syntax tree. The only reason we leave etree support in is to continue to support the large number of existing third party extensions. However, I'm not interested in expanding etree support.

arkadesOrg:
Interesting. So you'd rather print the XML-tree directly, right?
You want to develop the project away from common interfaces like ET or lxml?
So, hopefully you are not the only Developer to make the decision on this.
I am pretty sure the imagination of the community can follow my constructive
intention of enabling entended lxml functionality with almost no cost.

waylan:
> So you'd rather print the XML-tree directly, right?

Yes, that is correct.

> You want to develop the project away from common interfaces like ET or lxml?

Yes. Those are terrible HTML libraries IMO. They make sense for XML, but for HTML, they feel like a hack. I've discussed this at length elsewhere so I won't rehash it here.

That said, there are no plans to make any changes at this time because we want to continue supporting the existing library of third party extensions. I expect that if anything changes, it would be that this project gets put into maintenance mode. In other words, we would only consider/commit bug fixes and made no feature additions.

arkadesOrg:
I see. But someone who uses python to process Markdown will most probably need further transformations beyond bare XHTML output.
Anyways, could you lead me to the public discussion on ET/lxml as HTML libraries? I have to re-evaluate my own targets. Thank you!

waylan:
> But someone who uses python to process Markdown will most probably need further transformations beyond bare XHTML output.

Then they should be making those transformations after Markdown returns its output, not as part of the Markdown parser itself.

For the reasons stated here as well as in #1179, I am closing this.

PR #1179, opened 2021-08-24
Custom parser interface

arkadesOrg:
Regarding the [earlier pull request](https://github.com/Python-Markdown/markdown/pull/1177) I propose a _custom parser interface_ for better integration with third party projects using e.g. _lxml_.

### Purpose:
Help developers of third party projects to easily change the internal parser (etree) if their custom extensions do have need for extended functionality of e.g. _lxml_.

### Usage:
```python
import markdown
import lxml.etree
markdown.etree.set_parser(lxml.etree)
html = markdown.markdown("# Hello extended XPath World")
```

### Criticism:
Compatibility issues of third party parsers do not have to be targeted by Python-Markdown development team, as projects like _lxml.etree_ claim to be compatible with _xml.etree.ElementTree_

### Licensing:
There should be no legal conflicts (BSD/MIT) in case of lxml. :-D

**I hope you'll accept my idea to make my life a bit easier beyond python-markdown update cycles.**

Cheers
Dom

### DIFF
--- a/markdown/blockparser.py
+++ b/markdown/blockparser.py
@@ -19,7 +19,7 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
-import xml.etree.ElementTree as etree
+from . import etree
 from . import util
 
 
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -32,7 +32,7 @@ as they need to alter how markdown blocks are parsed.
 
 import logging
 import re
-import xml.etree.ElementTree as etree
+from . import etree
 from . import util
 from .blockparser import BlockParser
 
--- a/markdown/etree.py
+++ b/markdown/etree.py
@@ -1,0 +1,9 @@
+try:
+  import lxml.etree as etree
+except:
+  import xml.etree.ElementTree as etree
+  
+def __getattr__(name):
+    """Get attribute."""
+    
+    return getattr(etree, name)
--- a/markdown/etree.py
+++ b/markdown/etree.py
@@ -1,9 +1,45 @@
-try:
-  import lxml.etree as etree
-except:
-  import xml.etree.ElementTree as etree
+"""
+Python Markdown
+
+A Python implementation of John Gruber's Markdown.
+
+Documentation: https://python-markdown.github.io/
+GitHub: https://github.com/Python-Markdown/markdown/
+PyPI: https://pypi.org/project/Markdown/
+
+Started by Manfred Stienstra (http://www.dwerg.net/).
+Maintained for a few years by Yuri Takhteyev (http://www.freewisdom.org).
+Currently maintained by Waylan Limberg (https://github.com/waylan),
+Dmitry Shachnev (https://github.com/mitya57) and Isaac Muse (https://github.com/facelessuser).
+
+Copyright 2007-2021 The Python Markdown Project (v. 1.7 and later)
+Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)
+Copyright 2004 Manfred Stienstra (the original version)
+
+License: BSD (see LICENSE.md for details).
+"""
+
+import sys
+
+def set_parser(parser):
+  """ Set parser to alternative like e.g. lxml.etree
+  usage:
+
+  import markdown
+  import lxml.etree
+  markdown.etree.set_parser(lxml.etree)
+  """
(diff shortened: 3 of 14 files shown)


waylan:
My initial kneejerk reaction when seeing `markdown.etree.set_parser(lxml.etree)` was was 'but we aren't using etree as a parser so `set_parser` is the wrong name for the function.' I think that that gets to my underlying objection to adding any support for lxml. We are a Markdown parser. 

What harm is there in allowing third party extensions to inject a separate parser? Well, then we officially support it with all its edge cases. Now ever time some user runs into some obscure edge case with lxml, we need to deal with it. No thanks. I'm not interesting in the extra burden that the long term support will require.

Of course, this doesn't prevent you from accomplishing your end result. Just return a string from your parser which you then insert as a raw string into the Markdown output.

arkadesOrg:
I can't follow your argumentation because an interface for optional parsers doesn't oblige the maintainers to support any of those optional injections. It's the third party's obligation to be xml.etree.ElementTree compliant.

I don't see the point of forcing extension developers to rely on the built-in xml parser.

I won't inject lxml-trees into markdown results. it's cumbersome and excludes extended xpath operations on markdown-html result. let's fork.

facelessuser:
All of the basic built-in Markdown elements are built as internal plugins. They rely on etree. If we allow custom parsers, how do we avoid the inevitable questions of "I implemented my own parser now all of the basic Markdown things are broken?". Or "Everything isn't broken but with lxml this edge case is broken for Admontions, can we fix that so it isn't broken?".

This is what we don't want to deal with. I don't really know what you are doing with lxml and at what stage in the process, but I imagine there are ways to do what you need to without lxml. I realize that lxml is a preferred option for you, but I can envision all sorts of issues if we went this route.

And this isn't really a completely abstracted approach as there may be non-etree like parsers out there that someone would want to use. This really would only work for etree-like parsers, particularly lxml. What if I wanted to use html5lib? I don't think this would work. You'd have to completely abstract the interface. But again, you'd still have quirks maybe.

This is the bare minimum of abstraction needed to get lxml (and probably only lxml) shoehorned in, but not enough to abstract this for all custom parsers and it may introduce all sorts of quirks for anyone interfacing with the built-in Markdown parsing when using their own custom parser.

arkadesOrg:
Again, it is not the Python-Markdown's responsibility to guarantee every third party's parser compability. I know that there are some differences between lxml and ElementTree, but they can be solved without touching the Python-Markdown's source code by overriding affected methods from my own project's side for example.

> I don't think this would work.

Did you ever try to implement an abstract interface for html5lib? Programming is not a belief system.

>  it may introduce all sorts of quirks for anyone interfacing with the built-in Markdown parsing

Again, python is no voodoo magic. I use lxml and Python-Markdown and it works great on a great variety of Markdown source. Why would I commit code improvements then? Anyway I'll solve the issue with local patches. I see no point in supporting a closed source ElementTree cult without reasonable argumentation.

> inevitable questions of "I implemented my own parser now all of the basic Markdown things are broken?"

Seriously? Common I'll just fork it and add a Python-Markdown2 archlinux package...

My decision fell for Python-Markdown because of it's expandability with further tree processing. I quickly realized that my objectives need lxml functionality. There's no other reason to use it when markdown CLI tool is times faster.

facelessuser:
> Did you ever try to implement an abstract interface for html5lib? Programming is not a belief system.

I am involved in the [BeautifulSoup](https://launchpad.net/beautifulsoup) project as I provide the [CSS selector](https://github.com/facelessuser/soupsieve) library for that project. That project actually does create an abstract interface (for the purpose of extracting the content) around the built-in HTML lib, lxml, and html5lib, so I do know a little about this, and I don't appreciate the condescending tone. Let's try and keep this civil.

> Again, it is not the Python-Markdown's responsibility to guarantee every third party's parser compatibility.

I understand that, but it may introduce issues into existing pieces we do maintain. But it also may not. The point is that the maintainer has selected an HTML parser, and is not currently interested in opening the door for more "potential" maintenance burden. It may not offer any issues currently, but it could in the future, and once we start supporting this, we are committed to handling related workarounds etc. that may be required in the future.

> Again, python is no voodoo magic. I use lxml and Python-Markdown and it works great on a great variety of Markdown source. Why would I commit code improvements then? Anyway I'll solve the issue with local patches. I see no point in supporting a closed source ElementTree cult without reasonable argumentation.

You are free to do so as you feel is necessary as that is the benefit of open source, but just because something is open source does not mean they are obligated to accept every suggested feature or fix.

PR #1187, opened 2021-10-27
Support for custom Pygments formatter

sharat87:
Hi team!

This adds configuration support for using a custom Pygments formatter, either by giving the string name, or a custom formatter class (or callable).

Inspired by #490, which wasn't merged back then because v3 work was going on. I'm hoping we can merge this feature now please?

Thank you for your time.


waylan:
This looks like a great start. Before we merge this, at a minimum, we will need to address the failing tests.

1. As this is a new feature, it will require us to move up to a new MINOR version (version 3.4; see the [Contributing Guide][1]). Therefore, a new file will need to be created at `docs/change_log/release-3.4.md` for the release notes, and those release notes need to be added.
2. The word `formatters` should be added to [`.spell-dict`][2] so that it is recognized as a correctly spelled word.
3. Some tests should be added to [`tests/test_syntax/extensions/test_fenced_code.py`][3] which confirm that the new API works as advertised.

Finally, I wonder if perhaps we should have some error handling to deal with bad string names and/or objects (classes/callables) which don't properly match Pygments' API. As a reminder, a user passing text to Markdown should never result in an error. However, creating an instance of the Markdown class certainly can. So any error checking usually needs to happen  when the extension is loaded, not when it is passed user input. If that is not possible or practical (as I suspect is the case here), then perhaps we should fall back to a default formatter on error.

[1]: https://python-markdown.github.io/contributing/#versions
[2]: https://github.com/Python-Markdown/markdown/blob/master/.spell-dict
[3]: https://github.com/Python-Markdown/markdown/blob/master/tests/test_syntax/extensions/test_fenced_code.py

sharat87:
Understood. Thanks for detailed pointers @waylan. I checked the contributing guidelines that GitHub pointed me to, but the one you linked to has much more information. I'll get on this right away.

sharat87:
Tests now pass with Pygments installed and with Pygments uninstalled.

Screenshot 2021-10-28 at 11 45 12

sharat87:
Hey @waylan, I think I've covered all the feedback from your message. Please let me know if I've missed anything or if there's more changes you'd like before merging in.

Also, could you please run the workflows again so I can see if we get all green now? Thanks!

sharat87:
Fixed lint and spelling failures in the previous commit and pushed.

waylan:
This looks like it is ready to go. Thanks for doing the work. However, we have a bunch of bugfixes waiting for a bugfix release. Once we do that release, we can merge this. I've labeled this and added it to the correct milestone so that we don't forget about it.

sharat87:
Thank you for sharing the merging plan. I'll look forward to the 3.4 release then.

waylan:
I will merge this if the conflicts are resolved.

sharat87:
@waylan, thanks for the notification. I've just resolved merge conflicts.

PR #1189, opened 2021-11-07
CSS won't work here.

shitkoder:
CSS don't work in md files, it just gives a bad look.


facelessuser:
This CSS was specifically added so that when the documentation is rendered to HTML, the HTML document will use the CSS. It does work and does not give a "bad look", but instead applies the appropriate colors to the labels as desired.

shitkoder:
Better make a new html file but the person who will be going through docs can't have that css animation which you are trying to provide he'll just see all the code inside style tag below the docs.

facelessuser:
> Better make a new html file but the person who will be going through docs can't have that css animation which you are trying to provide he'll just see all the code inside style tag below the docs.

I'm not sure what you are trying to assert here, but let me be clear.  These Markdown pages are used to generate the final HTML pages of the official documentation. The CSS is needed and has been purposely added for this purpose to allow the labels to be colored: https://python-markdown.github.io/contributing/#issue-and-pull-request-labels.

It has a real purpose and will not be removed.

PR #1190, opened 2021-11-07
unwanted line

shitkoder:


facelessuser:
The line is unwanted by who?  This may be a personal preference of the author to specify the title explicitly with frontmatter instead of relying on the title being inferred from the first header. This is a very subjective change and adds no real value.

shitkoder:
There is no use of that line, you just gave the title below that line.

facelessuser:
Again, this is used in the MkDocs environment, which allows such notation to determine the title of the final documentation page. It is used and is used instead of whatever is specified by header 1. The author has specifically chosen to use this notation to specify the title regardless of whether the header happens to match the title or not. This is not about whether the page will or will not resolve to the same title, but how the documentation author has chosen to specify the title. 

For instance, see here: https://github.com/Python-Markdown/markdown/blob/master/docs/cli.md. Here the title does not match the first header. So instead of checking in each page if the header matches, they've chosen to "always" specify the title via frontmatter for consistency.

waylan:
This is being closed for the reasons given by @facelessuser.

PR #1208, opened 2022-01-06
Rename attr_list.md to Mgb.attr_list.md

MGBtrust:
Extension


mitya57:
Can you explain the rationale for this change?

waylan:
This does not fit the naming convention used by any of the included extensions. Additionally, it is not clear what `Mgb` could possibly stand for.

PR #1215, opened 2022-01-24
Convert % interpolation / .format to f-strings where feasible

akx:
Mostly mechanical conversion using pyupgrade, flynt, and PyCharm.


waylan:
While I don't have any objection to using f-strings over other options, this looks like a lot of change for changes sake. Generally, we only make updates like this only on the code which we are actively working on, not across the entire library. However, if you can demonstrate a clear benefit to making this change (like perhaps benchmarks demonstrating a notable increase in performance) then we may consider it.

PR #1216, opened 2022-01-25
Only enumerate extension entry points when required

akx:
On my machine, this saves 33 milliseconds off import time for the library, which could be significant for e.g. CLI applications.

```
(master) $ hyperfine 'python -m markdown'
Benchmark 1: python -m markdown
  Time (mean ± σ):     113.1 ms ±  11.5 ms    [User: 84.8 ms, System: 19.2 ms]
  Range (min … max):   100.1 ms … 151.7 ms    19 runs

(late-extensions) $ hyperfine 'python -m markdown'
Benchmark 1: python -m markdown
  Time (mean ± σ):      80.7 ms ±   4.7 ms    [User: 61.0 ms, System: 13.1 ms]
  Range (min … max):    73.9 ms …  94.8 ms    30 runs
```

There is also a small benefit to the overall test suite run time:

```
(master) $ hyperfine 'python -m unittest discover tests'
Benchmark 1: python -m unittest discover tests
  Time (mean ± σ):      1.177 s ±  0.074 s    [User: 1.073 s, System: 0.051 s]
  Range (min … max):    1.075 s …  1.250 s    10 runs

late-extensions) $ hyperfine 'python -m unittest discover tests'
Benchmark 1: python -m unittest discover tests
  Time (mean ± σ):      1.085 s ±  0.029 s    [User: 0.990 s, System: 0.046 s]
  Range (min … max):    1.034 s …  1.127 s    10 runs
```

Going forward, one could consider making the `importlib.metadata` backport an optional dependency – it's only required if one wants to load an extension by name.


waylan:
I will merge this if a note is added to the [release notes](https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/release-3.4.md).

PR #1218, opened 2022-02-03
Footnotes improvements

ysard:
Hi, I suggest in this PR 2 improvements for the footnotes module.

- Small refactor of the `BACKLINK_TITLE` option; The use of `format()` instead of "old" `%d` formatter allows
to specify text without the need to have the number of the footnote in it (like footnotes on Wikipedia for example).
The modification is backward compatible so no configuration change is required.

- Addition of a new option `SUPERSCRIPT_TEXT` that allows to specify a custom placeholder for the footnote itself in the text.
Ex: <sup>[1]</sup> or <sup>(1)</sup> or just by default, the current behavior: <sup>1</sup>.

Some writing rules specify these subtleties I think it could be useful.

Thanks for reading.


waylan:
These look like some good improvements/changes to me. Of course, we do need to include an update to the changelog. However, as this is adding a new feature, it will need to be in a new feature release, which means a new document will need to be added for the release notes. However, we already have PR #1187 which handles that. This can be finalized and merged after that.

waylan:
Oh, I just realized there are no tests for the new features. We also need some new tests added to [tests/test_syntax/extensions/test_footnotes.py](https://github.com/Python-Markdown/markdown/blob/master/tests/test_syntax/extensions/test_footnotes.py).

ysard:
Thanks for the review, tests are added and passing `(Ran 945 tests in 3.294s, OK (skipped=21))`. I also added the changelog hoping that it will make your work easier after the merge for 3.4.

PR #1223, opened 2022-02-21
fix: added None check in `convert` to prevent no attribute error

JoeHCQ1:
I'm getting a NoneType error when this [mkthedocstrings function](https://github.com/mkdocstrings/mkdocstrings/blob/master/src/mkdocstrings/handlers/base.py#L219) calls this function with an empty string. While I may submit a PR there to avoid this bug too, I expect this change will make this module better for everyone.

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -244,6 +244,9 @@ class Markdown:
 
         """
 
+        # Prevent `AttributeError: 'NoneType' object has no attribute 'strip'`
+        if source is None:
+            return ''
         # Fixup the source text
         if not source.strip():
             return ''  # a blank unicode string


waylan:
I can't think of any situation in which we should be accepting anything other than a string here. In fact, back in the Python 2 days (when the default string type was a byte string rather than Unicode), we only accepted Unicode strings. It was always the user's responsibility to ensure that a Unicode string be passed in. I don't see why we should change that policy now. In fact, this is clearly [documented][1].

> Python-Markdown expects a Unicode string as input (some simple ASCII binary strings may work only by coincidence) and returns output as a Unicode string. Do not pass binary strings to it! If your input is encoded, (e.g. as UTF-8), it is your responsibility to decode it. 

Given the above, I would consider the issue to be an upstream bug.

[1]: https://python-markdown.github.io/reference/#markdown

JoeHCQ1:
@waylan excellent points and thank you for responding. I'm new to both of these modules (this and the upstream one I linked) so I wasn't sure who's responsibility it was. Given your input I'm happy to close this and I'll open a new ticket upstream.

Thank you for your time!

Issue #1227, opened 2022-03-01
importib-metadata is too much constrainted for older environment than python 3.10


Krilivye:
Hello team, thanks for the work!

Regarding this https://github.com/Python-Markdown/markdown/issues/1203 issue, I also ran into a dependency problem regarding the latest markdown 3.3.6 pushing for using importlib-metadata>=4.4 and latest flake8 constraining it to importlib-metadata<4.3 💥 : conflict.

I'm a bit confused as I see that markdown support python 3.6->3.10.
According to importlib-metadata Changelog, https://importlib-metadata.readthedocs.io/en/latest/history.html markdown should need only 3.0.0 to support python 3.6.

So I would suggest restricting importlib-metadata to >=3.0.0 to keep backward dependencies compatible code with the ecosystem (who is not up to date with python 3.10 which requires 4.4 ;) ).

I'm downgrading markdown in my projects to 3.3.4 due to that but it makes me sad.

Theorie reasoning behind.
I'm a user of both flake8 and markdown (requirements.txt only contains flake8 and markdown).
constraints are free (no pinned version)
on a python3.6->3.9 I won't be able to install the latest markdown -> conflict
on python3.10 flake8 will be the bad guy... but that is expected as flake8 does not support python3.10 yet (nor 3.9).

With a lighter constraint, I'll be able to have markdown installed in every python3.6 -> 3.10 ecosystem

Also according to importlib-metadata 1.4 is used in python (stlib) 3.8.

I also discovered that pip6 support python version constraint so it could be something like that (didn't test it):

```
importlib-metadata >=1.4, <4.4; python_version <= '3.10'
importlib-metadata >=4.4 ; python_version >= '3.10'
```

_Originally posted by @Krilivye in https://github.com/Python-Markdown/markdown/issues/1203#issuecomment-1054619758_

waylan:
> I'm a bit confused as I see that markdown support python 3.6->3.10.

There is nothing to be confused about. So long as you don't have any third party libs installed which don't conflict, then Python-Markdown works in all of those Python versions. The problem is that some more popular third party packages have been slow to adopt Python 3.10. It is those packages which are causing the problem.

We are not going to revert our support of Python 3.10. However...

> ```
> importlib-metadata >=1.4, <4.4; python_version <= '3.10'
> importlib-metadata >=4.4 ; python_version >= '3.10'
> ```

That might be a reasonable way to approach this. I hadn't thought of that. A PR is welcome.

[...]

PR #1228, opened 2022-03-02
Release importlib-metadata constraints

Krilivye:
Solve #1227

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,7 @@ setup(
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=3.6',
-    install_requires=["importlib-metadata >=3.6,
+    install_requires=["importlib-metadata >=3.6",
                       "importlib-metadata >=4.4; python_version >= '3.10'"],
     extras_require={
         'testing': [
--- a/setup.py
+++ b/setup.py
@@ -74,7 +74,8 @@ setup(
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=3.6',
-    install_requires=["importlib-metadata>=4.4;python_version<'3.10'"],
+    install_requires=["importlib-metadata >=3.6,
+                      "importlib-metadata >=4.4; python_version >= '3.10'"],
     extras_require={
         'testing': [
             'coverage',


setup.py
```diff
@@ -74,7 +74,8 @@ def get_version():
     license='BSD License',
     packages=['markdown', 'markdown.extensions'],
     python_requires='>=3.6',
-    install_requires=["importlib-metadata>=4.4;python_version<'3.10'"],
+    install_requires=["importlib-metadata >=3.6",
+                      "importlib-metadata >=4.4; python_version >= '3.10'"],
```
waylan, line 78: After taking another look at this, I question the usefulness of this. Note this code:

https://github.com/Python-Markdown/markdown/blob/af38c42706f8dff93694d4a7572003dbd8b0ddc0/markdown/util.py#L32-L36

We only import `importlib-metadata` for `python_version < 3.10`. Therefore, we don't need `importlib-metadata` at all for `python_version >= 3.10`. Lets add `python_version < '3.10'` to the first line and remove the second line.

I now remember why we didn't do this earlier. We certainly cannot go back any further than 3.6 as we need to use `group`. As a reminder, in Python 3.10, `group` is the only supported method, so we have to use group to support Python 3.10. While support for `group` was first added in 3.6, my understanding is that it was in beta until 4.4, which is the first version which is guaranteed to have behavior which matches what is found in Python 3.10. Sure, maybe it all seems to work in our limited testing, but can we be confident that it will always work as expected in all cases for all users? I can only be reasonably sure if we require 4.4.

And finally, does 3.6 even give you what you want? I forget which version flake8 is pinned at.

Krilivye: 🤔 flake8 is limited to `importlib-metadata<4.3` 
So the >=3.6 constraints in a python<3.10 should be resolved by a 4.2 install with flake8 by dependencies resolver tools like pip/poetry.
I think that tox could solve the "can it work with all combinations of python version & importlib-metadata range library?" by empirical testing if we don't trust importlib-metadata documentation.

waylan: > if we don't trust importlib-metadata documentation.

The issue is that Python 3.10's [Release Notes][1] states that Python 3.10 has feature parity with import-metadata 4.6. Therefore, my assumption is that any earlier version of importlib-metadata will not behave exactly as expected in every edge case. And I'm not interested in testing every edge case. Therefore, we will not be supporting any earlier version.

As an aside, I see that we pinned 4.4. It appears we should have pinned 4.6. If we change anything, it will only be to pin 4.6 for `Python < 3.10`.

[1]: https://docs.python.org/3/whatsnew/3.10.html#importlib-metadata

PR #1241, opened 2022-04-01
extensions: copy config dict on each highlighted block

### LINKED ISSUE #1240: codehilite extension: Pygments style only applied to first block
_I have tracked down a bug and written a fix for this, wanted to explain the issue in an Issue. PR: https://github.com/Python-Markdown/markdown/pull/1241._

I'm running a [small static blog](https://blog.g3rt.nl/) using [Pelican](https://getpelican.com/), the [Elegant theme](https://elegant.oncrashreboot.com/) and I have enabled the Markdown plugin to do code highlighting for me.

My Markdown configuration via Pelican is like this (note `'pygments_style': 'native'`):

```
MARKDOWN = {
            'output_format': 'html5',
            'extension_configs': {
                'markdown.extensions.admonition': {},
                'markdown.extensions.codehilite': {
                                                   'css_class': 'highlight',
                                                   'noclasses': True,
                                                   'pygments_style': 'native',
                                                  },
                'markdown.extensions.extra': {},
                'markdown.extensions.meta': {},
                # Enable permalink on headings.
                # Note: value set to a space, so that the symbol doesn't become
                # part of the RSS feed as per recommendation.
                # https://elegant.oncrashreboot.com/permalinks-to-headings
                'markdown.extensions.toc': {'permalink': ' '},
            },
           }
```

With the upgrade of the Markdown package from 3.2.2 to 3.3._any_, the highlighting of any _subsequent_ blocks after the first would be 'default' instead of 'native': 🤔 

Screenshot_20220402_001102

The commit 10058fae6577e79b786f4e185218aebb1b53e937 (_Refactor fenced_code & codehilite options_) seems to have introduced a 'pop' rather than proper 'get' from the local configuration in the sequence of blocks, resulting in this behaviour.

https://github.com/Python-Markdown/markdown/blob/383de86c64101b8d14768d9a247c9efc97d703bd/markdown/extensions/codehilite.py#L238-L245

gertvdijk:
This fixes a bug where any subsequent highlighted block with codehilite would result in the omission of the style setting, falling back to default.

Fixes #1240

### DIFF
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -6,6 +6,7 @@ Python-Markdown Change Log
 (under development): version 3.3.7 (a bug-fix release).
 
 * Disallow square brackets in reference link ids (#1209).
+* Retain configured `pygments_style` after first code block (#1240).
 
 Nov 17, 2021: version 3.3.6 (a bug-fix release).
 
--- a/markdown/extensions/codehilite.py
+++ b/markdown/extensions/codehilite.py
@@ -237,11 +237,12 @@ class HiliteTreeprocessor(Treeprocessor):
         blocks = root.iter('pre')
         for block in blocks:
             if len(block) == 1 and block[0].tag == 'code':
+                local_config = self.config.copy()
                 code = CodeHilite(
                     self.code_unescape(block[0].text),
                     tab_length=self.md.tab_length,
-                    style=self.config.pop('pygments_style', 'default'),
-                    **self.config
+                    style=local_config.pop('pygments_style', 'default'),
+                    **local_config
                 )
                 placeholder = self.md.htmlStash.store(code.hilite())
                 # Clear codeblock in etree instance
--- a/tests/test_syntax/extensions/test_code_hilite.py
+++ b/tests/test_syntax/extensions/test_code_hilite.py
@@ -644,3 +644,35 @@ class TestCodeHiliteExtension(TestCase):
             expected,
             extensions=[CodeHiliteExtension(unknown='some value')],
         )
+
+    def testMultipleBlocksSameStyle(self):
+        if has_pygments:
+            # See also: https://github.com/Python-Markdown/markdown/issues/1240
+            expected = (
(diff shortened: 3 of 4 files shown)


mitya57:
Can you please add a test for multiple highlighted blocks?

gertvdijk:
@mitya57 Added them to the now amended commit. 😃

gertvdijk:
[CI error](https://github.com/Python-Markdown/markdown/runs/5807792601?check_suite_focus=true) appears unrelated to my change. 😕 

> 43 files checked.
> ERROR: 5 files with dead links found!

waylan:
The reason why we `pop` the setting is because we also pass the rest of the config items on to Pygments (see the very next line). As each of Pygmants' lexers/formatters have a different set of options they accept, we just pass all unknown keywords on. However, there is no need to pass the style on as we have already defined that. The only reason we don't get an error is because Pygments just ignores any unknown keywords passed to its lexers and formatters. Still, it would be best to not pass it on to avoid any weird conflicts in the future. 

Perhaps we should be making a copy of the config for each code block and then we can pop the setting off of the copy.

As an aside, we have the `pygments_style` setting for historical reasons (it existed before we accepted all Pygments options). If we were starting fresh today we would not define a special option for it and users could just use `style`  which would get passed it on with the rest of the Pygments' keyword options. Actually, a user could avoid this bug by using `style` now (without this fix). That said, we should still fix the bug. Although, we might want to deprecate the `pygments_style` option as it is duplicative and unnecessary.

waylan:
> [CI error](https://github.com/Python-Markdown/markdown/runs/5807792601?check_suite_focus=true) appears unrelated to my change. 😕
> 
> > 43 files checked.
> > ERROR: 5 files with dead links found!

Yes, I see that. I have opened a separate issue for that in #1243. You can ignore that failure here.

gertvdijk:
@waylan Thanks for the explanation! 😃 
It took a bit longer to get back to this, but I've pushed an amendment.

PR #1250, opened 2022-05-05
Add markdown-link-check guthub action.

waylan:
This is experimental to compare to running the CLI which we run form tox. It gives us the additional option of only running against changed files, which would eliminate the frustration contributors face when a file they didn't touch results in a failure.


waylan:
[This][1] is with quiet mode off and [this][2] is with it on. I'm leaving it on as we only need to see the problems, not what checks out okay.

[1]: https://github.com/Python-Markdown/markdown/runs/6310892561?check_suite_focus=true
[2]: https://github.com/Python-Markdown/markdown/runs/6310925356?check_suite_focus=true

Issue #668, opened 2018-06-17
attr_list for table and ul/ol ?


ikus060:
I see multiple issues (#312, #227, #505, #594, #600) related to `attr_list` applied to table, ol or ul and I can't figure out if we settle on a specific syntax. Looking at the code it's not obvious if it's possible to assign class to ol/ul.

If not supported, I'm wondering why. Clearly, it's a highly requested feature and I see multiple proposal for syntax and implementation.

If supported, what is the right syntax ? and may we get this documented on the web site.

Thanks !

waylan:
I don't know what is so confusing. Sure, various non-members have proposed various solutions on the cited issues. However, my position has remained constant. Regarding [lists][1] I explain that:

> ...the markdown syntax has no concept of a ul or ol. They are simply implied by the fact that we have list items.

And [tables][2]:

> ...how would one differentiate between attributes defined on a row, or on a table. The syntax only defines the cells, the rows and table are inferred.

It is not possible to assign attributes to elements which don't actually exist in the document. Therefore, we will never attempt to support such a feature, regardless of how many requests or proposals we receive. Of course, anyone is free to create their own extensions which have any behavior they want.

This is being closed as another duplicate of #227 and #312.

[1]: https://github.com/Python-Markdown/markdown/issues/227#issuecomment-20223738
[2]: https://github.com/Python-Markdown/markdown/issues/312#issuecomment-44227093

Issue #227, opened 2013-06-28
attr_list on ul/ol elements?


heston:
I'm not sure if this is a bug, or just a lack of documentation, but it seems impossible to get att_list to apply to `<ul>` or `<ol>` elements. It can only be applied to individual `<li>` elements.

```
s = """
* Item 1
* Item 2
{: .some_list }
"""

markdown.markdown(s, extensions=['extra'])
```

Results in:

```
<ul>
    <li>Item 1</li>
    <li class="some_list">Item 2</li>
</ul>
```

I would have expected:

```
<ul class="some_list">
    <li>Item 1</li>
    <li>Item 2</li>
</ul>
```

Is there a way to get an attribute to apply to the whole list?

waylan:
> ```
> s = """
> * Item 1
> * Item 2
> {: .some_list }
> """
> ```

If that results in the class being assigned to the `ul`, then how would you propose assigning a class to a list item? Remember that secondary lines of a list item are not required to be indented (unless they are nested block items--which we don't have here). That is a basic markdown concept defined in the original syntax rules.

The fact is, the markdown syntax has no concept of a `ul` or `ol`. They are simply implied by the fact that we have list items. How are we supposed to assign attributes to something that basically doesn't exist in the syntax?

So to answer your question: no, it is not possible to assign attributes to the list itself. Granted, being able to do so would certainly make for a better hook for CSS and Javascript. I just don't see a clear way to do so that will not veer too far from Markdown's philosophy.

The basic syntax was inspired by [Maruku](http://maruku.rubyforge.org/proposal.html#attribute_lists)'s Attribute List feature. How does that implementation handle lists? It isn't documented and I don't recall offhand, but that might shed some light on things.

heston:
Intuitively, I would expect the attribute to reside on the `<li>` if it was defined on the same line (or on the same line as the last line of the `<li>` block):

```
* Item 1
Item 1a{: .class_on_li }
* Item 2{: .class_on_li }
```

However, this could conflict with inline syntax, and it does get ambiguous with nested `<p>` tags. Not to mention it changes the existing grammar.

Another idea is to leave the `<li>` attribute placement as-is, and introduce a new placement for `<ul>`. Something like this:

```
* Item 1
Item 1a
* Item 2

{: .class_on_ul }
```

Notice the extra blank line after the last `<li>`. I'm not sure of this is feasible, but I don't think it conflicts with existing syntax. It could also be used with the existing `<li>` placement:

```
[...]

PR #1252, opened 2022-05-05
Allow setting attributes on <ol> and <ul> tags with attr_list extension

paulmelis:
This adds a way for the attr_list extension to lift attribute values set on list items to their parent list element (i.e. `<li>` attribute to `<ul>` or `<ol>`). This allows list-wide attributes, such as list start number or class to be set by setting them on list items with the appropriate syntax.

Contains additions to the docs describing it, as well extra test input (plus all existing unit tests still pass)

Should be able to solve #227, #668 and many of the other requests for a similar feature, including my own quest to get this feature.

### DIFF
--- a/docs/extensions/attr_list.md
+++ b/docs/extensions/attr_list.md
@@ -169,7 +169,7 @@ can be set. To still be able to set attributes on the full list itself the follo
 1. Item1
 {^ start=4 }
 2. Item2
-{: #item2 }
+{ #item2 }
 3. Item3
 {^ .inline }
 
@@ -181,7 +181,7 @@ can be set. To still be able to set attributes on the full list itself the follo
 {^ #myid }
 ```
 
-The use of `^` instead of `:` at the start of the attribute definition indicates that the attributes values are
+The use of `^` at the start of the attribute definition indicates that the attributes values are
 to be set on the parent of the list item, i.e. the implicitly generated list start tag (either `<ol>` or `<ul>`
 in the output HTML, depending on the type of list).
 
--- a/docs/extensions/attr_list.md
+++ b/docs/extensions/attr_list.md
@@ -158,6 +158,54 @@ assigned to the inline element (`<em>`) which immediately preceded it.
 
 Attribute lists may also be defined on table header cells (`<th>` elements) in the same manner.
 
+### Setting attributes on lists
+
+Unlike HTML there is no explicit start and end tag of a list in Markdown. A list begins when the first item
+is parsed, closed after the last item. So the list items are the only places on which an attribute can be set.
+To still be able to set attributes on the full list itself the following syntax can be used:
+
+
+```text
+1. Item1
+{^ start=4 }
+2. Item2
+{: #item2 }
+3. Item3
+{^ .inline }
+
+* Item1
+{^ .firstclass }
+* Item2
+{^ .secondclass }
+* Item3
+{^ #myid }
+```
+
+The use of `^` instead of `:` at the start of the attribute definition indicates that the attributes values are
+to be set on the parent of the list item, i.e. the implicitly generated list start tag (either `<ol>` or `<ul>`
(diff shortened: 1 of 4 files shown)


docs/extensions/attr_list.md
```diff
@@ -158,6 +158,54 @@ assigned to the inline element (`<em>`) which immediately preceded it.
...
+
+* Item1
+{^ .firstclass }
+* Item2
+{^ .secondclass }
+* Item3
+{^ #myid }
+```
+
+The use of `^` instead of `:` at the start of the attribute definition indicates that the attributes values are
```
Andre601: `:` isn't even needed anymore. It's a relict of the old syntax and is from what I know only still supported for backwards compatibility.
Either way, the new (and recommended) syntax is `{ <stuff> }`

paulmelis: Fine, updated the docs

facelessuser:
I'm not commenting on whether this feature should or should not be included, but how would you specify both a class on an `<li>` and on `<ol>` or  `<ul>` if you must do it on the same line?

paulmelis:
> I'm not commenting on whether this feature should or should not be included, but how would you specify both a class on an `<li>` and on `<ol>` or `<ul>` if you must do it on the same line?

As mentioned in the doc update this isn't possible with this patch. I initially started out with a `parent:` prefix (e.g. `parent:attr=value`) to be able to more precisely specify which attributes should be lifted to the parent, but decided against it as it needed quite more code changes. Plus I wonder how many times you would really want to be able to set attributes on both list items and the list itself. For the latter you only need to use an attribute on a single list item, while all the other list items can still have their own attributes.

paulmelis:
An alternative could be to have a single ^ as a marker in the attribute list to indicate that all attributes following it should be set on the parent. That would allow both list and table attributes with minimal code changes.

paulmelis:
I updated the pull request to use `^` as a marker to signal when attributes should be lifted. Solves the issue of not being to define both list element attributes and parent attributes in the same attribute definition

waylan:
As I have stated on every request we have received for this, we will not be implementing this feature. Of course, you are free to provide this functionality in a third party extension.

paulmelis:
> As I have stated on every request we have received for this, we will not be implementing this feature. Of course, you are free to provide this functionality in a third party extension.

I have seen a number of similar comments about not wanting to implement it, but I'm really struggling to understand why? This feature is currently not available, yet is obviously valuable to many, given all the requests over the many *years* it has shown up as an issue here. The way I implement it here does not seem to interfere with any existing feature, as far as I can tell. And if the special `^` marker isn't used at all then there's no difference in behaviour, so it should be fully backwards compatible. It adds value, not only to me, yet it is coldly being rejected for unclear reasons other than "we will not be implementing this feature". 

Please help me understand, because this all does not make any sense...

waylan:
This has been discussed and explained many times over. You even linked to some of those discussions. In short, I am not interested in supporting a syntax that is does not make sense to me. It should be no surprise that it was rejected. Of course, you are free to disagree with me on this. If this is important to you, then you can provide and support your own third-party extension which provides the feature.

Andre601:
> In short, I am not interested in supporting a syntax that is does not make sense to me.

From me as an outside-viewer I can see this response as essentially "I don't like this, so I don't want it" response.
You put your own bias over something that the majority of people may like to have for once. I saw this problem in MkDocs and see it here again. You seem to have a general issue with accepting things that you don't like, but that the community could benefit from.

waylan:
I don't accept things that I'm not willing to support long term. Long after the person who proposes it is gone, I'm left defending it and fixing issues with it. That's (one of the reasons) why we provide a public extension API. Anyone can provide their own implementation which works how they want. Then they can support it without any additional burden on me. I am not limiting anyone at all except for myself.

PR #1253, opened 2022-05-10
Added type hints for deprecated decorator.

lucas-six:


facelessuser:
So, I'm curious, as this project does not support type hints in any way, why specifically the pull for the deprecation decorator?

lucas-six:
> So, I'm curious, as this project does not support type hints in any way, why specifically the pull for the deprecation decorator?

Well, I want to determine if PR for type hints would be accepted. If yes, I plan to cover all step by step (if time permits)

waylan:
I'm not opposed to type hints. However, I generally dislike a wholesale change across the entire code base without a clear benefit. As type hints do not affect how the code works in any way, there is no clear benefit here. In the past, when adopting some new style, we have usually adapted it piecemeal by only implementing it in the sections of code that are being worked in for other reasons. However, I am also not interested in receiving a new PR every so often solely to add type hints to some new section of code.

facelessuser:
Well, this could be implemented with `pyi` files. Ideally it wouldn't impact the entire code base because it would be kept separate. This is of course assuming you wanted to go in this direction.

waylan:
I was going to mention [stub files][1], but couldn't remember what they were called. So thanks @facelessuser.

By the way, a separate stub package (see [PEP 561][3]) already exists for Python-Markdown at [types-Markdown][2]. If a user wants type hints, they can install that package. Alternatively, we could include the stub files with our own package. As the work is already done, I would suggest asking to copy the existing work first before redoing it all from scratch.

Given the above, I think that we will not be modifying the code to include type hints. Instead, all type hints should be implemented by stub files. That said, we have two options:

1. Leave things as-is. Users can install `types-Markdown` if they want type hints.
2. Copy the stub files from `types-Markdown` into our own source (with permission of course).

Therefore, I am closing this PR as we will not be taking this route. Of the other options I'm fine with either. Option 1 means it is not an additional maintenance issue we need to be concerned with (the maintainers of that package are responsible for keeping it up-to-date). On the other hand, by bringing it in-house, we can ensure it doesn't get out-of-sync.

[1]: https://peps.python.org/pep-0484/#stub-files
[2]: https://pypi.org/project/types-Markdown/
[3]: https://peps.python.org/pep-0561/#stub-only-packages

lucas-six:
> 

> Well, this could be implemented with `pyi` files. Ideally it wouldn't impact the entire code base because it would be kept separate. This is of course assuming you wanted to go in this direction.

I think the `pyi` solution is just fine.

Thanks @facelessuser @waylan.

PR #1254, opened 2022-05-12
Use `@deprecated` for Registry.add()

lucas-six:
Clean the redundancy codes.

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -423,6 +423,7 @@ class Registry:
         else:
             raise KeyError('Cannot delete key {}, not registered.'.format(key))
 
+    @deprecated('Use the `register` method instead.')
     def add(self, key, value, location):
         """ Register a key by location. """
         if len(self) == 0:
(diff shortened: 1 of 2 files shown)


markdown/util.py
```diff
@@ -459,12 +460,6 @@ def add(self, key, value, location):
             raise ValueError('Not a valid location: "%s". Location key '
                              'must start with a ">" or "<".' % location)
         self.register(value, key, priority)
-        warnings.warn(
-            'Using the add method to register a processor or pattern is deprecated. '
-            'Use the `register` method instead.',
```
waylan, line 464: You removed the first half of the message. Let's add it back in.

mitya57: `deprecated` decorator itself includes the method name and words that it is deprecated:

https://github.com/Python-Markdown/markdown/blob/9d90d47157f4348ad70f1a070571de87efbc6cbc/markdown/util.py#L117

Also, I wonder if we can do the same for the other deprecated methods, like `__setitem__` and `__delitem__`.

waylan: Ah right. Thanks for the reminder.

PR #1256, opened 2022-05-13
Use `@deprecated` for __setitem__ and __delitem__ of Registry.

lucas-six:


markdown/util.py
```diff
@@ -387,15 +387,10 @@ def _sort(self):
 
     # Deprecated Methods which provide a smooth transition from OrderedDict
 
+    @deprecated('Use the `deregister` method instead.')
```
mitya57:
```suggestion
    @deprecated('Use the `register` method instead.')
```

mitya57 (changes requested):

mitya57 (approved):

PR #1259, opened 2022-05-24
Drop support for PY36

waylan:
Python dropped support on 2021-12-23. Our policy (#760) is to drop
support on the next point release after Python does.


.github/workflows/tox.yml
```diff
@@ -33,7 +31,7 @@ jobs:
           - tox-env: py310
             python-version: '3.10'
           - tox-env: pypy3
-            python-version: pypy3
+            python-version: pypy-3.7
```
waylan, line 34: If we wait for the next release of actions/setup-python, it will presumably include the changes in actions/setup-python#349. Then we can exclude the `-` from the version (like this `pypy3.7`), which is more consistent with all other version indicators.

Additionally, I wonder if we should perhaps run pypy tests for all supported versions (3.7, 3.8, 3.9).

waylan:
As PEP562 was added in PY37, we can also remove the backport. This should resolve the failing coverage check.

PR #1260, opened 2022-05-24
Remove previously deprecated objects.

waylan:
This completely removes all objects which were deprecated back on version 3.0. We are now working on 3.4 so I think perhaps we can move forward with this. The one hesitation I have is that it is possible some older (unmaintained) extensions could stop working. However, given the time that has passed, and the fact that those extensions probably don't even support the new minimum Python version, I expect this to mostly be a non-issue.

So far, I have only made changes to the code. I still need to add something to the release notes.

### DIFF
--- a/docs/change_log/release-3.4.md
+++ b/docs/change_log/release-3.4.md
@@ -32,7 +32,7 @@ In addition, tests were moved to the modern test environment.
 
 ### Previously deprecated objects have been removed
 
-Various objects were dreprecated in version 3.0 and began raising deprecation
+Various objects were deprecated in version 3.0 and began raising deprecation
 warnings (see the [version 3.0 release notes] for details). Any of those object
 which remained in version 3.3 have been removed from the codebase in version 3.4
 and will now raise errors. A summary of the objects are provided below.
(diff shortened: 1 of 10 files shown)


waylan:
After having made these changes it occurs to me that we could perhaps have raised some sort of DeprecationError for a version before completely removing these objects. Any thoughts?

docs/change_log/release-3.4.md
```diff
@@ -30,6 +30,34 @@ markdown.markdown(src, extensions=[TableExtension(use_align_attribute=True)])
 
 In addition, tests were moved to the modern test environment.
 
+### Previously deprecated objects have been removed
+
+Various objects were dreprecated in version 3.0 and began raising deprecation
```
mitya57:
```suggestion
Various objects were deprecated in version 3.0 and began raising deprecation
```

mitya57 (commented):
Looks good, just one typo noticed.

mitya57:
> After having made these changes it occurs to me that we could perhaps have raised some sort of DeprecationError for a version before completely removing these objects. Any thoughts?

You mean one intermediate step between DeprecationWarning and complete removal? I don't see any sense, warnings are quite visible and I don't remember any other projects doing this. Some projects have distinct PendingDeprecationWarning and DeprecationWarning, though.

mitya57 (approved):

Issue #1263, opened 2022-05-26
PrettifyTreeprocessor stripping whitespace in `<pre><code>`


fourpoints:
Again this small code section [markdown.treeprocessor.py:L432](https://github.com/Python-Markdown/markdown/blob/96d27f191bf31e76fca8c4294a3238519d13a631/markdown/treeprocessors.py#L432). I hope I'm not too annoying hitting these edge cases :sweat_smile:

My element looks like this; it's a table where each line is wrapped in `<pre><code>`.

```html
<table>
  <tbody>
    <tr>
      <td class="lineno">1</td>
      <td><pre><code><span>def</span> <span>foo</span><span>()</span><span>:</span></code></pre></td>
    </tr>
    <tr>
      <td class="lineno">2</td>
      <td><pre><code>    <span>return</span> <span>2</span></code></pre></td>
    </tr>
  </tbody>
</table>
```
which looks like this
```python
1 def foo():
2    return 2
```

The Prettifier turns this into something like this
```html
<table>
  <tbody>
    <tr>
      <td class="lineno">1</td>
      <td><pre><code>
<span>def</span> <span>foo</span><span>()</span><span>:</span></code></pre></td>
    </tr>
    <tr>
      <td class="lineno">2</td>
      <td><pre><code>
<span>return</span> <span>2</span></code></pre></td>
    </tr>
  </tbody>
</table>
```
which looks like this
```python
1
def foo():
2
return 2
```

Looking at [Babelmark](https://johnmacfarlane.net/babelmark2/?normalize=1&text=%60%60%60%0A++++return+5%0A%60%60%60) only pandoc and Python-Markdown appears to be stripping whitespace in `<pre><code>`, though fenced code isn't supported by default.

Possible fixes:
* Don't change anything, keep it as it is.
* Remove section.
* Don't prettify if `pre[0].text.isspace()` is `True` (would fix my case).

waylan:
When you made your previous report my first thought was that it is strange the we are altering whitespace in a code block. But then I realized that the intent was only to remove extra blank lines at the end of a code block. Perhaps the earlier code which parses the code block should be removing those blank lines instead of addressing it here. 

[...]

PR #1264, opened 2022-05-26
Update PrettifyTreeprocessor `<pre><code>` handling (attempt 2)

fourpoints:
Fixes issue #1263 

Added some tests near the other `PrettifyTreeprocessor` tests, not sure if this is the correct place for them.

Let me know if additional edge cases should be included, or if the added ones should be changed.


waylan:
Great work. Thank you.

PR #1270, opened 2022-06-14
fixed some spelling mistakes,

venthur:
that's all :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,5 +60,5 @@ You may report bugs, ask for help, and discuss various other issues on the [bug 
 Code of Conduct
 ---------------
 
-Everyone interacting in the Python-Markdown project's codebases, issue trackers,
+Everyone interacting in the Python-Markdown project's code bases, issue trackers,
 and mailing lists is expected to follow the [Code of Conduct].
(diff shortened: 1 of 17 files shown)


waylan:
I recently noticed [this failure](1) elsewhere. It would appear that I introduced the error in #1260 and then failed to notice the failing test. We should probably fix it here. We have two options.

1. Change `codebase` to `code base` (two words).
2. Add `codebase` to the [.spell-dict][2] as a single word.

A quick search shows that a few dictionaries have added the single word, however they all appear to be the community editable type. Therefore, I am more inclined to go with two words unless someone can provide a good reason not to.

[1]: https://github.com/Python-Markdown/markdown/runs/6900216611?check_suite_focus=true#step:5:26
[2]: https://github.com/Python-Markdown/markdown/blob/master/.spell-dict

waylan:
I just noticed that this PR is changing the `php` and `pl` tests. Those tests have been copied from their respective projects and are to be used as-is. In other words, we do not make any edits to those files. All edits to those files will need to be reverted for this to be accepted.

venthur:
done and done :)

PR #1272, opened 2022-07-13
Move backslash unescaping to treeprocessor

### LINKED ISSUE #1131: Smarty extension causes generation of invalid HTML syntax
This works as expected:

```console
$ echo '![x"x](x)' | python3 -m markdown -x smarty
<p><img alt="x&quot;x" src="x" /></p>
```

but this unexpectedly generates invalid HTML syntax:

```console
$ echo '![x\"x](x)' | python3 -m markdown -x smarty
<p><img alt="x"x" src="x" /></p>
```

I would expect the second output to be the same as the first. Backslash should suppress special interpretation by Markdown; it should not also *induce* special interpretation by HTML.

waylan:
By unescaping backslash escapes in a treeprocessor, the text is properly
escaped during serialization. Fixes #1131.

As it is recognized that varous third-party extensions may be calling the
old class at `postprocessors.UnescapePostprocessor` the old class remains
in the codebase, but has been deprecated and will be removed in a future
release. The new class `treeprocessors.UnescapeTreeprocessor` should be
used instead.

### DIFF
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -16,7 +16,7 @@ License: [BSD](https://opensource.org/licenses/bsd-license.php)
 from . import Extension
 from ..treeprocessors import Treeprocessor
 from ..util import code_escape, parseBoolValue, AMP_SUBSTITUTE, HTML_PLACEHOLDER_RE, AtomicString
-from ..postprocessors import UnescapePostprocessor
+from ..treeprocessors import UnescapeTreeprocessor
 import re
 import html
 import unicodedata
@@ -84,8 +84,8 @@ def stashedHTML2text(text, md, strip_entities=True):
 
 def unescape(text):
     """ Unescape escaped text. """
-    c = UnescapePostprocessor()
-    return c.run(text)
+    c = UnescapeTreeprocessor()
+    return c.unescape(text)
 
 
 def nest_toc_tokens(toc_list):
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -289,10 +289,10 @@ class TocTreeprocessor(Treeprocessor):
                     toc_tokens.append({
                         'level': int(el.tag[-1]),
                         'id': el.attrib["id"],
-                        'name': unescape(stashedHTML2text(
+                        'name': stashedHTML2text(
                             code_escape(el.attrib.get('data-toc-label', text)),
                             self.md, strip_entities=False
-                        ))
+                        )
                     })
 
                 # Remove the data-toc-label attribute as it is no longer needed
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -19,6 +19,7 @@ Copyright 2004 Manfred Stienstra (the original version)
 License: BSD (see LICENSE.md for details).
 """
 
+import re
 import xml.etree.ElementTree as etree
 from . import util
 from . import inlinepatterns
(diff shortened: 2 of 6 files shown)


tests/basic/backlash-escapes.html
```diff
@@ -9,7 +9,7 @@
 <p>Right bracket: ]</p>
 <p>Left paren: (</p>
 <p>Right paren: )</p>
-<p>Greater-than: ></p>
+<p>Greater-than: &gt;</p>
```
waylan, line 12: This is the one and only change in behavior in the existing tests. I'm okay with this, however, as technically this results in valid output. The reason for the change is that the angle bracket gets escaped during serialization. Previously, a placeholder was there during serialization, which was swapped out for the actual character later. The whole point of this change was to better ensure valid HTML output, so this is an acceptable change in behavior.

markdown/treeprocessors.py
```diff
@@ -429,3 +431,28 @@ def run(self, root):
...
+        return chr(int(m.group(1)))
+
+    def unescape(self, text):
+        return self.RE.sub(self._unescape, text)
+
+    def run(self, root):
+        """ Loop over all elements and unescape all text. """
+        for elem in root.iter():
+            # Unescape text content
+            if elem.text and not elem.tag == 'code':
```
waylan, line 451: I'm not sure we actually need to skip `code` tags, In fact, if I remove the check, the tests all pass. In fact, the previous code did not have a way to distinguish between code and other content. However, there is always a possibility that `code` could intentionally contain what looks like a placeholder. In that case, the content should not be altered. Therefore, I have left the check in.

markdown/extensions/toc.py
```diff
@@ -16,7 +16,7 @@
 from . import Extension
 from ..treeprocessors import Treeprocessor
 from ..util import code_escape, parseBoolValue, AMP_SUBSTITUTE, HTML_PLACEHOLDER_RE, AtomicString
-from ..postprocessors import UnescapePostprocessor
+from ..treeprocessors import UnescapeTreeprocessor
```
waylan, line 19: I excepted to not need to explicitly call the unescaping code from the TOC extension. After all, the TOC extension updates the Tree prior to the `UnescapeTreeprocessor` being run. However, when I remove the explicit calls to escape, the tests which specifically test this code fail. I haven't worked out why.

waylan (commented):
Below are a few comments and concerns I have about this change. Feedback is welcome.

facelessuser:
I'll try and look at this soon. I'd like to pull it and see how it impacts some of my things.

markdown/extensions/toc.py
waylan: I figured out the reason. The text of a TOC item is passed to `slugify` for the `id`. However, it needs to be unescaped beforehand. Otherwise, `slugify` mucks up the placeholder so that the new Treeprocessor doesn't recognize it and fails to make the swap in the `id`. However, the text in the body of the TOC item does not need unescaped within the TOC extension as it will be handled by the new Treeprocessor. I have removed that call in acb0c31.

tests/basic/backlash-escapes.html
mitya57: Having unescaped `>` in HTML was a bug, so good that you fixed it.

mitya57 (approved):
As all tests pass and the bug is fixed, I am happy with this change.

facelessuser (approved):
I see nothing breaking on my end. Seems good.

PR #1288, opened 2022-08-28
Change Treeprocessor to TreeProcessor and create an alias for backwar…

### LINKED ISSUE #815: alias TreeProcessor to Treeprocessor
In the interest of consistency with `BlockProcessor` and most other `*Processor`. I suggest to provide an alias `TreeProcessor` to the class `Treeprocessor`.

This would reduce the friction for extension writers. At least a bit.

venthur:
…ds compatibility.

Closes: #815

### DIFF
--- a/docs/extensions/api.md
+++ b/docs/extensions/api.md
@@ -218,18 +218,17 @@ Some block processors in the Markdown source tree include:
 [Admonition]: https://python-markdown.github.io/extensions/admonition/
 [b4]: https://github.com/Python-Markdown/markdown/blob/master/markdown/extensions/admonition.py
 
-### Tree processors {: #treeprocessors }
+### Tree Processors {: #treeprocessors }
 
 Tree processors manipulate the tree created by block processors.  They can even create an entirely new ElementTree
 object. This is an excellent place for creating summaries, adding collected references, or last minute adjustments.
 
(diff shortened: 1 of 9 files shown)


venthur:
Not sure which of the other classes matching `*TreeProcessor` need an alias too, just let me know, and I'll update accordingly.

waylan:
I'm inclined to not merge this as the current capitalization matches the other classes (preprocessor, postprocessor, etc.). Anyone else have any thoughts?

facelessuser:
FYI, BlockProcessor is an outlier: https://github.com/Python-Markdown/markdown/blob/master/markdown/blockprocessors.py#L59. I'm not sure which way you'd like to normalize class names 🤷🏻.

waylan:
So it would appear that for the least amount of change we would change `BlockProcessor => Blockprocessor`, which would result in a consist style.

If we went the other way, we would need to change everything except `BlockProcessor`. Although, I can see an argument for not camel casing `Preprocessor` as that would be a single work in normal English prose. I suspect that is where the current style of not camel casing any of them case from. `BlockProcessor` was certainly a later addition and perhaps we forgot about that earlier style choice.

venthur:
I'm fine either way, I've now also created https://github.com/Python-Markdown/markdown/pull/1290 that changes spelling for the `BlockProcessor` to `Blockprocessor`

PR #1290, opened 2022-08-30
Change BlockProcessor to Blockprocessor

venthur:
this makes it consistent with the spelling of the other processors.

This PR also adds an alias of the old BlockProcessor to the new
Blockprocessor for backwards compatibility.

### DIFF
--- a/markdown/blockparser.py
+++ b/markdown/blockparser.py
@@ -60,7 +60,7 @@ class State(list):
 class BlockParser:
     """ Parse Markdown blocks into an ElementTree object.
 
-    A wrapper class that stitches the various BlockProcessors together,
+    A wrapper class that stitches the various Blockprocessors together,
     looping through them and creating an ElementTree object.
     """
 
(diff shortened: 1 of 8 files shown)


facelessuser:
I imagine the old name should be marked as deprecated via [PEP 562](https://peps.python.org/pep-0562/).

waylan:
While I understand the motivation for making this change I do have some apprehension about accepting it. Specifically, by proposing this you have ignored the suggestion in our [contributing guidelines][1]:

In other words, I would be more emendable to this change if we were making a change to the class itself. Of course, there would need to be a real need for the change, not just an arbitrary change.

Regardless, before this change could be accepted, a note would need to be made in the release notes and the old class would need to be marked as deprecated (note that a `deprecated` decorator exists at [markdown.util.deprecated][2]). Also, a change like this would need to be part of a new point release (rather than a bug-fix release), which I don't expect to happen for a while (we just released 3.4.0). Therefore, even if this PR was accepted, it wouldn't likely be merged until we were ready to start development in earnest on 3.5.0.

[2]: https://github.com/Python-Markdown/markdown/blob/master/markdown/util.py#L91

PR #1294, opened 2022-09-11
Fix #1293 Add customizable TOC title class

### LINKED ISSUE #1293: Allow TOC element to be fully customizable
```
[TOC]
# Header 1
```

will currently generate the following HTML:

```
<div class="toc">
  <ul>
    <li><a href="#header-1">Header 1</a></li>
  </ul>
</div>
<h1 id="header-1">Header 1</h1>
```

The `div` element cannot be changed and so it is for the title `span` element and its `toctitle` class.

I suggest that other config parameters (`toc_tag`, `title_class`, `title_tag`) that defaults to their current defaults are introduced.

This would be very useful for producing semantic HTML, using specific HMTL elements.

For example to use the [details disclosure element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/details) as toc, the extension could be called like so:

```
md = markdown.Markdown(extensions=[TocExtension(
    title='Summary',
    title_class="",
    title_tag="summary",
    toc_class="",
    toc_tag="details"
)]]
```

```
<details>
<summary>Summary</summary>
<ul>
<li><a href="#header">Header</a></li>
</ul>
</details>
<h1 id="header">Header</h1>
```

This will render an initially closed details element with a summary:

--- 

<details>
<summary>Summary</summary>
<ul>
<li><a href="#header">Header</a></li>
</ul>
</details>
<h1 id="header">Header</h1>

pauloxnet:


waylan:
This appears to be an implementation of the feature requested in #1293. What happens here depends on what decision is made there.

waylan:
Given the decision made in #1293, I will not be accepting this as-is. However, I am not opposed to including the new config option to allow the user to define the class assigned to the title. After all, we already allow various other classes to be set by the user. Might as well include this one too.

If this is updated (with all checks passing), then I would be happy to accept it. If no action is taken, it will likely be closed.

pauloxnet:
I'll try to work on that in may because tomorrow I'm leaving to go to PyCon US.

pauloxnet:
@waylan I rebased this PR and left only the customizable TOC title class.
I renamed this PR because is changed a lot from my original idea.

waylan:
The tests will need to all be passing before we can accept this. Also, the release notes and the documentation need to be updated appropriately.

pauloxnet:
> The tests will need to all be passing before we can accept this. Also, the release notes and the documentation need to be updated appropriately.

@waylan I fixed test failures and updated the documentation

waylan (approved):

PR #1303, opened 2022-11-14
Improve standalone * and _ parsing.

### LINKED ISSUE #1300: stand-alone * or _ lead to incorrect results
A text with several stand-alone underscores is not converted correctly.

**input**
```
**_ _ _ N _ _ _ _ _ _ _ R S _ _ N _**
```

**output**
```
 <p><strong><em> _ </em> N _ <em> _ </em> _ <em> _ R S _ </em> N _</strong></p>
```

**expected**
```
<p><strong>_ _ _ N _ _ _ _ _ _ _ R S _ _ N _</strong></p>
```

It seems that underscores with the surrounding spaces are replaced by the `placeholder`.
In the original on [https://daringfireball.net/projects/markdown/dingus](https://daringfireball.net/projects/markdown/dingus) it is converted correctly.

waylan:
The `NOT_STRONG_RE` regex matchs 1, 2, or 3 * or _ which are surrounded by white space to prevent them from being parsed as tokens. However, the surrounding white space should not be consumed by the regex, which is why lookhead and lookbehind assertions are used. As `^` cannot be matched in a lookbehind assertion, it is left outside the assertion, but as it is zero length, that should not matter.

Tests added and/or updated to cover various edge cases. Fixes #1300.

### DIFF
--- a/tests/test_syntax/inline/test_emphasis.py
+++ b/tests/test_syntax/inline/test_emphasis.py
@@ -36,6 +36,42 @@ class TestNotEmphasis(TestCase):
             '<p>_</p>'
         )
 
+    def test_standalone_asterisks_consecutive(self):
+        self.assertMarkdownRenders(
+            'Foo * * * *',
+            '<p>Foo * * * *</p>'
+        )
+
+    def test_standalone_understore_consecutive(self):
+        self.assertMarkdownRenders(
+            'Foo _ _ _ _',
+            '<p>Foo _ _ _ _</p>'
+        )
+
+    def test_standalone_asterisks_pairs(self):
+        self.assertMarkdownRenders(
+            'Foo ** ** ** **',
+            '<p>Foo ** ** ** **</p>'
+        )
+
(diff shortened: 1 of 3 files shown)


tests/test_syntax/inline/test_emphasis.py
```diff
@@ -72,10 +108,16 @@ def test_standalone_understores_with_newlines(self):
...
+            '_ foo_ bar',
+            '<p>_ foo_ bar</p>'
+        )
+
+    def test_standalone_asterisk_at_end(self):
         self.assertMarkdownRenders(
-            'foo * bar *',
-            '<p>foo * bar *</p>'
+            'foo *bar *',
+            '<p>foo *bar *</p>'
```
waylan, line 120: Note that there was no change in behavior for this test. This change ensures that the test is actually testing what it was intended to test. By making the first `*` not standalone, the only way the test will pass is if the second one is standalone.

facelessuser (approved):

PR #1304, opened 2022-11-15
modernized syntax allowed for py37+

venthur:
This pr modernizes the syntax for things that are allowed for pyhton3.7+. It is mostly

* ''.format -> f''
* set() -> {}
* removal of explicit utf8 encoding of files

this is related to #760

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -259,7 +259,7 @@ class CodeBlockProcessor(BlockProcessor):
             code = sibling[0]
             block, theRest = self.detab(block)
             code.text = util.AtomicString(
-                '{}\n{}\n'.format(code.text, util.code_escape(block.rstrip()))
+                f'{code.text}\n{util.code_escape(block.rstrip())}\n'
             )
         else:
             # This is a new codeblock. Create the elements and insert text.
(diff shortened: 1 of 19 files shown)


waylan:
Please note that the [Contributing Guide][1] states the following:

That said, `.format -> f'` could be classified as a performance improvement, especially if we were provided some numbers to back it up. However, that should be a separate PR by itself. And removal of explicit utf8 encoding of files could be considered code cleanup I suppose (again as a separate PR. However, `set() -> {}` is clearly a style thing and the sort of change our Contributing Guide discourages.

venthur:
Alright, I thought it might help to cleanup the code a bit, but feel free to reject the PR if it is not needed.

waylan:
As I said, if you want to do the work to show that the `.format -> f'` change provides us with a clear performance improvement, then we would be happy to accept the work.

However, I don't actually see any changes which remove explicit utf8 encoding of files. So if you want to reclassify the PR to be about the first issue you can. Just remove the set changes.

venthur:
> As I said, if you want to do the work to show that the .format -> f' change provides us with a clear performance improvement, then we would be happy to accept the work.

No not really, actually. If you don't like the proposal you can just go ahead and close it. No hard feelings.

> However, I don't actually see any changes which remove explicit utf8 encoding of files. So if you want to reclassify the PR to be about the first issue you can. Just remove the set changes.

It is at the very bottom of the diff:

https://github.com/Python-Markdown/markdown/pull/1304/files#diff-57787454b56d61377fe6ec8e971df1e1e1ec9d6f1f0dd4a13209bf5bfb55633bL1

The idea of the PR was to make the code more readable, I was not aware of that rather strict guideline not to touch code unless it fixes something else. Sorry for the noise.

akx:
Seems to be a revisit of my #1215, fwiw.

waylan:
I'm closing this. If an update to provided which addresses the concerns raised, we can reopen.

PR #1309, opened 2022-11-30
Consider <html> a block-level HTML element

### LINKED ISSUE #1308: <html> tag wrapped in <p>
Since v3.3 (more precisely, https://github.com/Python-Markdown/markdown/pull/803 / https://github.com/Python-Markdown/markdown/commit/b701c34ebd7b2d0eb319517b9a275ddf0c89608d), converting the following test document _bug-test.txt_:

```markdown
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <link rel="stylesheet" href="style.css"/>
  </head>
  <body>

!!! header "foo"
    bar
```
by running e.g.
```bash
markdown_py -o html5 -x extra -x admonition bug-test.txt
```
results in 
```html
<!DOCTYPE html>
<p><html>
  <head>
    <meta charset="utf-8"/>
    <link rel="stylesheet" href="style.css"/>
  </head>
  <body></p>
<div class="admonition header">
<p class="admonition-title">foo</p>
<p>bar</p>
</div>
```

Prior to v3.3, the result was as expected:
```html
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8"/>
    <link rel="stylesheet" href="style.css"/>
  </head>
  <body>

<div class="admonition header">
<p class="admonition-title">foo</p>
<p>bar</p>
</div>
```
Does the new HTML parser added in https://github.com/Python-Markdown/markdown/pull/803 perhaps not account for `<html>` HTML tag?

kernc:
Fixes https://github.com/Python-Markdown/markdown/issues/1308

I'm obviously :﻿+1: on adding `<html>` to block-level elements. Literal HTML is valid syntax in markdown, and the [specification for `DocumentFragment`](https://dom.spec.whatwg.org/#interface-documentfragment) doesn't mention any limits wrt what particular nodes the fragment is or isn't allowed to consist of ...

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -43,7 +43,7 @@ BLOCK_LEVEL_ELEMENTS = [
     # Other elements which Markdown should not be mucking up the contents of.
     'canvas', 'colgroup', 'dd', 'body', 'dt', 'group', 'html', 'iframe', 'li', 'legend',
     'math', 'map', 'noscript', 'output', 'object', 'option', 'progress', 'script',
-    'style', 'tbody', 'td', 'textarea', 'tfoot', 'th', 'thead', 'tr', 'video'
+    'style', 'summary', 'tbody', 'td', 'textarea', 'tfoot', 'th', 'thead', 'tr', 'video'
 ]
 
 # Placeholders
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -41,7 +41,7 @@ BLOCK_LEVEL_ELEMENTS = [
     'h4', 'h5', 'h6', 'header', 'hgroup', 'hr', 'main', 'menu', 'nav', 'ol',
     'p', 'pre', 'section', 'table', 'ul',
     # Other elements which Markdown should not be mucking up the contents of.
-    'canvas', 'colgroup', 'dd', 'body', 'dt', 'group', 'iframe', 'li', 'legend',
+    'canvas', 'colgroup', 'dd', 'body', 'dt', 'group', 'html', 'iframe', 'li', 'legend',
     'math', 'map', 'noscript', 'output', 'object', 'option', 'progress', 'script',
     'style', 'tbody', 'td', 'textarea', 'tfoot', 'th', 'thead', 'tr', 'video'
 ]
--- a/docs/change_log/index.md
+++ b/docs/change_log/index.md
@@ -6,6 +6,7 @@ Python-Markdown Change Log
 *under development*: version 3.4.1 (a bug-fix release).
 
 * Improve standalone * and _ parsing (#1300).
+* Consider `<html>` HTML tag a block-level element (#1309).
 
 July 15, 2022: version 3.4.1 (a bug-fix release).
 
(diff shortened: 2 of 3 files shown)


markdown/util.py
```diff
@@ -41,9 +41,9 @@
     'h4', 'h5', 'h6', 'header', 'hgroup', 'hr', 'main', 'menu', 'nav', 'ol',
     'p', 'pre', 'section', 'table', 'ul',
     # Other elements which Markdown should not be mucking up the contents of.
-    'canvas', 'colgroup', 'dd', 'body', 'dt', 'group', 'iframe', 'li', 'legend',
+    'canvas', 'colgroup', 'dd', 'html', 'body', 'dt', 'group', 'iframe', 'li', 'legend',
     'math', 'map', 'noscript', 'output', 'object', 'option', 'progress', 'script',
-    'style', 'tbody', 'td', 'textarea', 'tfoot', 'th', 'thead', 'tr', 'video'
+    'style', 'summary', 'tbody', 'td', 'textarea', 'tfoot', 'th', 'thead', 'tr', 'video'
```
facelessuser, line 46: This change is probably above and beyond what was originally discussed. I think `summary` was purposely left out of the block-level list.

kernc: Summary was added already in d1ed480d. I'm just proposing a move for consistency.

kernc:
FTR, some older third parties seem to use `BLOCK_LEVEL_ELEMENTS` list incompatibly:
https://github.com/HEInventions/docnado/blob/7e474a34036248fa773f9487be21252db03a7b5f/docnado/docnado.py#L348-L356

markdown/util.py
facelessuser: Ah, that makes sense. It wasn't updated in both places.

facelessuser:
> FTR, some older third parties seem to use `BLOCK_LEVEL_ELEMENTS` list incompatibly:

We can't really control how people monkey patch the lib. Once you start monkey patching like they do, you are on your own.

markdown/util.py
```diff
@@ -41,9 +41,9 @@
     'h4', 'h5', 'h6', 'header', 'hgroup', 'hr', 'main', 'menu', 'nav', 'ol',
     'p', 'pre', 'section', 'table', 'ul',
     # Other elements which Markdown should not be mucking up the contents of.
-    'canvas', 'colgroup', 'dd', 'body', 'dt', 'group', 'iframe', 'li', 'legend',
+    'canvas', 'colgroup', 'dd', 'html', 'body', 'dt', 'group', 'iframe', 'li', 'legend',
```
waylan: This is a nitpick, but `html` is the only item in the list which is not located alphabetically.

waylan (approved):

waylan (changes requested):
We need to add a comment to the changelog indicating that the list of block-level-elements was changed.

facelessuser (approved):

waylan (approved):

PR #1313, opened 2022-12-09
Officially support Python 3.11

waylan:
Python 3.11 was released 2023-10-02.

I've added test here. I will merge if/when the tests pass.


pspacek:
> Python 3.11 was released 2023-10-02.

Wow, version from future! :joy:

PR #1323, opened 2023-03-06
Add "&gt;" support for blockquote

alberic89:
This commit adds the possibility to use “`&gt;`” instead of “>” to introduce a blockquote. This prevents blockquotes from not being rendered when Markdown is escaped.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -276,7 +276,7 @@ class CodeBlockProcessor(BlockProcessor):
 
 class BlockQuoteProcessor(BlockProcessor):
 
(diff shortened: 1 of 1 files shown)


waylan:
Are there any other implementations which employ this behavior? More specifically, does the reference implementation?

If not, then I would suggest that you can implement this as a third-party extension.

facelessuser:
> Are there any other implementations which employ this behavior? 

Absolutely not. When `>` is HTML escaped, it is not treated as Markdown by any implementation that I know of: 
 https://babelmark.github.io/?text=%26gt%3B+test

PR #1324, opened 2023-03-10
Switch from setup.py to pyproject.toml

waylan:
This completely replaces `setup.py` with `pyproject.toml`.

@facelessuser and @mitya57 if either of you would like to be listed in the `maintainers` entry, I would be happy to add you. You can be listed with or without an email address. If you would also like your email address listed, provide it and I'll include it as well.


pyproject.toml
```diff
@@ -2,3 +2,88 @@
 # Minimum requirements for the build system to execute.
 requires = ["setuptools>=36.6", "wheel"]
```
mitya57: I think you now need to bump minimum setuptools version to 61 (or even 61.2, as some recommend).

mitya57 (commented):
I don't do much these days, so it's okay to not list me in maintainers.

facelessuser:
Yeah, you can add me.

pyproject.toml
waylan: Good call. Thanks.

waylan:
@facelessuser I have added your name. If you can confirm I spelled it correctly, etc. then I will merge this.

PR #1330, opened 2023-03-27
Use pyspelling to check spelling.

waylan:
This is currently only running on `--group docs` which is consistent with that we were checking before. We can add `--group source` later when we have time to fix any detected mistakes (I'm sure there are many).

Unfortunately, I am not able to get apsell installed locally (for uninteresting reasons) so I'm testing this on the CI server for the first time. I don't know what results to expect.

@facelessuser as pyspelling is your project, any input for suggestions for improvement are welcome.


waylan:
It seems strange that there are multiple failed words which are in the custom wordlist, but the failed occurrence ends with an `'s`. Why were those not failing with our previous direct use of aspell?

facelessuser:
Yeah, feel free to ask any questions you have. It looks like it's running, so I'd just add anything to your dictionary that you want to ignore. If you want to avoid certain HTML elements, just use CSS selectors.

Anyways, just let me know if you get stuck on something as I'm not sure what to suggest until you run into a problem.

facelessuser:
As far as `'s` is concerned, it's a good question. There are probably plenty of questions on the internet related to that. Usually, I just add them to the list. As Pyspelling is just a wrapper around Aspell or Hunspell, it just sends the content to them. We apply some of our own filters to give us more control than is offered in them, but ultimately, Aspell and Hunspell do the spell checking.

facelessuser:
I'm sure there are ways we could further process such things as `'s`. Strip them out? Force a word break by replacing `'` with a space. Is there a way to tell Aspell to do that already? Maybe. Is it some dictionary option? 🤷🏻 That's why I just add them as it wasn't worth my time and I haven't seen any suggestions otherwise. Maybe if you use the C API directly? 🤷🏻

facelessuser:
> Why were those not failing with our previous direct use of aspell?

I just reread and noticed this. If this was working before, I'd love to know why as well. It is possible there is something I'm still missing. I'll take a look at what you were doing previously and see if I can find how we deviate.

facelessuser:
Using Aspell directly, `'s` does not work for me, so I do not know why it worked for you, maybe an older version?

I place `gldjsl's` in `test.txt`. I place `gldjsl` in `dict.txt` and I run `aspell` as shown below:

```console
$ aspell --lang=en create master ./dict.dic < dict.txt
$ aspell list --lang=en --extra-dicts=./dict.dic < test.txt
gldjsl's
```

waylan:
I've cleanup up  what needed to be cleanup up. Turns out there were some problems we were not catching so that is good now. However, there are two problems that were were not encountering before and I'm not sure what to do about them.

1. Some headers contain code spans and the text from those code spans are being inserted in the page's toc outside of a code span. This is normal behavior for the toc extension. However, it is causing spelling failures on words that should not be in the dictionary because they should always be in a code span. What is curious is that this was not causing a failure before and I don't understand why.

    A possible workaround would be to only check the spelling within a subset of each page. For example, only the text of elements which are children of a specified parent. However, I don't think pyspelling (or aspell for that matter) offers such a feature. We could possibly use a more basic template when building the docs which only includes the main body text. I'll have to think about that.

2. We have a bunch of instances `[name]'s` failing where `[name]` is in the dictionary. This was definitely not failing prior to using pyspelling and I an not inclined to add all of them to the dictionary. I did a search and found some comments about tokenization for spellchecking. The tokenizer needs to properly handle the `'s` for aspell to properly check these words. Presumably pyspelling is tokenizing the words differently that we were previously, but I have no idea how or why. I think I need to find some time to play with this on a local device and see what kind of behavior I can replicate in aspell.

Until I can resolve these two issues, this PR is on hold. If I decide to abandon it, I will likely extract the spelling fixes which have already been made here.

facelessuser:
@waylan 

1. If you can be more specific as to what you don't want to spellcheck and what you do, I imagine this can be worked out.
2. Did you see my example earlier that isn't even using PySpelling? What is it that you suggest is done? I don't know exactly what version of Aspell you were using previously, but all I did was what your shell script did: build a dictionary, and redirect the target file's input into Aspell. How were you tokenizing differently?

facelessuser:
PySpelling doesn't do any tokenization, it simply applies whatever filters you give it. As far as I see, Aspell doesn't tokenize the content either. English is not the only language in which this is used, so I don't want to get into the guessing game of how to handle special symbols in other languages. With that said, if you absolutely refuse to place possessive words in your list, you can provide a rule to ignore them. This is an easy fix. Feel free to tweak it as needed.

```yaml
  - pyspelling.filters.context:
      context_visible_first: true
      delimiters:
      # Ignore possessive endings
      - open: '(?<=\w)''s(?!\w)'
        close: '\b'
```

As for the other issue, if you are clear on what you wish to scan and not scan, I am fairly positive we can get a working setup.

facelessuser:
This may help you with your second issue. Let's assume I only wanted to spell check content under the `<article>` tag in my document, I could adjust the HTML filter `captures` rules to capture text under any element with any namespace under article, but not script or style.

```yaml
  - pyspelling.filters.html:
      comments: false
      attributes:
      - title
      - alt
      captures:
      - 'article *|*:not(script,style)'
      ignores:
      - 'code, pre'
      - 'a:is(.magiclink-compare, .magiclink-commit, .magiclink-repository)'
      - 'span.keys'
      - '.md-social__link'
```

waylan:
Awesome! Thanks for the help. I will take a look at this sometime in the next few days.

waylan:
> 2\. Did you see my example earlier that isn't even using PySpelling? What is it that you suggest is done? I don't know exactly what version of Aspell you were using previously, but all I did was what your shell script did: build a dictionary, and redirect the target file's input into Aspell. How were you tokenizing differently?

Yes I did see that. The current implementation is using the default version available on GitHub as it is running in a workflow. See the script in [checkspelling.sh](https://github.com/Python-Markdown/markdown/blob/be1c2839dd587a858f91c710e56667cba9f5329d/checkspelling.sh) which is called from a [tox env](https://github.com/Python-Markdown/markdown/blob/be1c2839dd587a858f91c710e56667cba9f5329d/tox.ini#L28-L33) via a [workflow](https://github.com/Python-Markdown/markdown/blob/be1c2839dd587a858f91c710e56667cba9f5329d/.github/workflows/tox.yml#L68-L90). The script just loops through the files and feeds them to a basic call to aspell. As that works but pyspelling doesn't, I'm assuming the problem is with pyspelling. But I'll need to play with things locally to be sure. Unfortunately, I haven't had the time or opportunity to do so yet.

facelessuser:
Since I've been able to take PySpelling out of the equation and demonstrate the exact scenario with possessive words, I'm fairly certain this is not a problem with PySpelling. With that said, with the workaround I posted, I think the problem becomes moot.

waylan:
I just figured out the apostrophe issue. The preexisting tests were rendering the Markdown to HTML such that the apostrophes are being rendered as html entities. In fact, the `smarty` extension is in our `mcdocs.yml` config file. So the real question is why actual apostrophes are appearing in the docs when using pyspelling?

And, of course, we do need to deal with the apostrophes in the other documents (the preexisting test only checked the docs, not RAEDME, etc).

Anyway, that is all I have time for tonight. I'll take another look at this later.

facelessuser:
Beautifulsoup is used to navigate the HTML. IIRC, it converts everything to friendly Unicode while your working with it. We don't convert/encode that back.

waylan:
> Beautifulsoup is used to navigate the HTML. IIRC, it converts everything to friendly Unicode while your working with it. We don't convert/encode that back.

That is what I suspected. This is a nonstarter for me. HTML entities should stay HTML entities. Any other behavior is a bug IMO.

Although, I wonder why converting to Unicode also converts HTML entities to unicode characters. That seems really weird to me. Is there a way to not have BS do that conversation specifically?

facelessuser:
> That is what I suspected. This is a nonstarter for me. HTML entities should stay HTML entities. Any other behavior is a bug IMO.

Well, that's just how BS4 works. It lets you work with the content of tags in plain text (attributes may be a different story, I don't recall). Since we extract it from the structure once it's decoded, we have no idea what was once entities and what was once not. At that point, it's not HTML anymore, it is text, which IMO is what I care about and what I want to spellcheck. I don't want to spell check `don` and `t` I want to spell check `don't` (which Aspell handles fine). Personally, I actually like that unnecessary quote entities are turned into real text.

I'm not sure why Aspell doesn't have an auto possessive handler. I guess it's possible it doesn't know context so it doesn't know when a word should be allowed to be possessive. It also works for other languages, so maybe they also didn't want to hardcode in English logic as a general approach, I don't know. And as pointed out earlier, possessive endings can be ignored already if that is what you want.

Also, keep in mind that you do not have to use the HTML filter provided with Pyspelling. You can forego it and enable Aspell's own HTML handler if you prefer that instead and have Pyspelling just give Aspell the full HTML content. I personally like being able to filter tags with CSS selectors, though. It was something I was invested in and actually wrote the CSS selector implementation used by BeautifulSoup. That is why you can capture the elements you want and ignore the elements you don't want to.

I'm not here though to convince anyone to use certain filters or even the application itself. People can also write their own filters if they like. They can create one that uses etree or whatever suits them best.

> Although, I wonder why converting to Unicode also converts HTML entities to unicode characters. That seems really weird to me. Is there a way to not have BS do that conversation specifically?

BS4 can re-encode the HTML, and it will encode absolutely necessary entities on output, but you have to configure it to encode extra entities beyond `&lt;`,`&gt;`, and `&amp;` (see [docs](https://www.crummy.com/software/BeautifulSoup/bs4/doc/#output-formatters)). Since the HTML filter provided by PySpelling is meant to extract text content from HTML, that is what it does, so it does not re-encode back to HTML.

waylan:
What exactly does pyspelling pass to aspell? Are you saying it extracts the plain text from the HTML and only passes that on? I have been assuming it passed on HTML and used aspell's `html` mode. No wonder I'm confused.

For that matter, regarding the README, LICENSE and similar files, I have been thinking that we should just be using aspell's `markdown` mode. Those files don't use any extension syntax and there is no reason to convert them to HTML. However, what I find strange is that that simple case is not even available from pyspelling without building my own filter.

In the end I am finding that while pyspelling is an amazingly powerful tool, it misses the mark on the simple cases. I had an easier time getting aspell to work directly. I'm not sure if I want to continue pursuing it. I'll have to think about it some more.

facelessuser:
Understand, feel free to use what fits you best. Personally, I needed a way to target specific content and ignore other content. Aspell requires you to wrap nonsense tags around content in HTML that you don't want to spell check (outside of ignoring code blocks). It was just too limited for my needs. Yes, flexibility comes with more complexity.

For me, it is a very powerful tool that allows me to capture spelling mistakes in the documentation and source code comments. It provides more flexibility than Aspell with the cost of some added complexity. For what I need, it fills the gap I had, but I understand it may not be for everyone.

If you have more questions, feel free to ask, if you move in a different direction, that is fine too.

waylan:
> However, what I find strange is that that simple case is not even available from pyspelling without building my own filter.

Turns out this is easy, although, as far as I can tell, not documented. I was looking through the source and noticed the `Aspell.spell_check_no_pipeline` method and realized that if I didn't define any pipelines at all and set `mode: markdown` under the `aspell` config option, then pyspelling would simply pass the file on to aspell in `markdown` mode; no rendering to HTML needed. This is exactly what I wanted for the README, LICENSE, etc files as they only use basic Markdown syntax without any extensions. See 91d09f2 for the change. This should be prominently documented IMO.

And then I added the suggested context filter to skip possessive S for the docs. And now everything is passing. Originally I was thinking that ignoring the possessive S would ignore the entire word. But then I realized it was only ignoring the apostrophe and the S immediately following, but would check the rest of the word. This seems like a reasonable compromise, which I have confirmed with testing.

facelessuser:
> Originally I was thinking that ignoring the possessive S would ignore the entire word.

Yeah, I should have made that more clear.

> This should be prominently documented IMO.

That's fair. If you create an issue with any doc confusions you ran into, I'll try to update them to be more clear when I get some time.

waylan:
I've been going through the source code and addressing issues with the code comments.

However, there are a few situations that I'm not sure how to handle. Specifically, multi-line comments typically follow Markdown formatting more-or-less. In most cases, they are just plain paragraphs with various code spans (so far, most of the edits have involved wrapping variable/function/class names in backticks). But there are a few occasions were we have what amounts to indented code blocks. Because we are not actually doing Markdown processing, and those code blocks contain CamelCased words, etc., they are failing the spelling check. Ideally, they would be ignored. What's the best way to do that? @facelessuser do you have any suggestions?

Here are a couple examples. In the first example we see an indented code block in a string-literal comment:

https://github.com/Python-Markdown/markdown/blob/9c95ec44021bf4abc4b030e2d7f156be3ad77a8f/markdown/inlinepatterns.py#L24-L35

In this specific example, the comment is not nested and therefore is not indented. But if such a comment was in a function/class/method, then the entire thing would be indented by at least however many levels deep the function/class/method is nested) . Which means we would need to dedent before determining indentation. This could get complicated fast. Or maybe we could use fenced code blocks and ignore content between the fences. That would eliminate the indentation problem. 

But I don't think fenced code blocks will work so well in the next example, where we have line comments (each line is prefaced by a hash).

https://github.com/Python-Markdown/markdown/blob/9c95ec44021bf4abc4b030e2d7f156be3ad77a8f/markdown/blockprocessors.py#L379-L383

While we do see indentation for the entire block here, we can ignore everything before the hashes. So I would imagine we could use a regex to check for indentation after the hash and ignore the line. Just wondering if there is a better approach. I suspect fenced code blocks would actually be more difficult to handle here.

I should note that if we need to limit what we allow, that is fine. We can always document it in our contributing guide. We just need to find a workable solution for each case above.

facelessuser:
I'd create a fenced code block rule like I use in my docs and fence them instead of indenting, though you can indent then as well they are fenced.

Basically, it's that (some kind of delimiter) or a special filter that parts the content.

facelessuser:
https://github.com/facelessuser/pyspelling/blob/master/.pyspelling.yml#L124

facelessuser:
Just an FYI, the fenced rule isn't strict Markdown fenced, it doesn't care that a hash comes in front.

facelessuser:
I take that back, but comment hashes get stripped

waylan:
> comment hashes get stripped

This is helpful. I wondered about that, but it wasn't clear to me. Thanks for all of the help.

waylan:
Not only do comment hashes get stripped, but the comments are dedented for string literal comments. Therefore, in both cases, it is easy to check for indented lines.

```
- pyspelling.filters.context:
   context_visible_first: true
   delimiters:
   # Ignore lines which are indented at least one level (Markdown code block)
   - open: '^ {4,}'
     close: '$'
```

This works as the context filter looks at each line one-at-a-time. The `open` regex matches the beginning of an indented line and `close` matches the end of the line. The only real concern here is that this could also match a nested list item. However, we do not appear to be using list items in our comments, so I'm not too worried about it.

That appears to have addressed the last of my concerns. I believe this is ready to go.

facelessuser:
Nice. I'm glad it is all working.

PR #1343, opened 2023-05-12
Hidden tags with content don't get removed

shahmirj:
We ran into an issue when doing the following:

```
Hello <param>world</param>
```

Which through markdown produced:

```
Hello <param />
```

This is because by default the output format is set to `xhtml` which is going through serialisation and removing any tag with content that is in the `HTML_EMPTY`, so we set the `ouput_format` to `html` and then the following happens:

```
Hello <param>world
```

So this fix now fixes this issues and produced valid html when `output_format='html`':

```
Hello <param>world</param>
```

### DIFF
--- a/markdown/serializers.py
+++ b/markdown/serializers.py
@@ -153,7 +153,7 @@ def _serialize_html(write, elem, format):
             write(' xmlns="%s"' % (_escape_attrib(namespace_uri)))
         if format == "xhtml" and tag.lower() in HTML_EMPTY:
             write(" />")
-        else:
+        elif text or len(elem) > 0:
             write(">")
             if text:
                 if tag.lower() in ["script", "style"]:
@@ -162,8 +162,13 @@ def _serialize_html(write, elem, format):
                     write(_escape_cdata(text))
             for e in elem:
                 _serialize_html(write, e, format)
-            if tag.lower() not in HTML_EMPTY:
-                write("</" + tag + ">")
+            write("</" + tag + ">")
+        elif tag.lower() in HTML_EMPTY:
+            if format == "xhtml":
+                write(" /")
+            write(">")
+        else:
+            write("></" + tag + ">")
     if elem.tail:
         write(_escape_cdata(elem.tail))
 
--- a/markdown/serializers.py
+++ b/markdown/serializers.py
@@ -164,8 +164,6 @@ def _serialize_html(write, elem, format):
                 _serialize_html(write, e, format)
             write("</" + tag + ">")
         elif tag.lower() in HTML_EMPTY:
-            if format == "xhtml":
-                write(" /")
             write(">")
         else:
             write("></" + tag + ">")
--- a/docs/change_log/release-3.4.md
+++ b/docs/change_log/release-3.4.md
@@ -111,3 +111,5 @@ The following bug fixes are included in the 3.4 release:
 * Extension entry-points are only loaded if needed (#1216).
 * Added additional checks to the `<pre><code>` handling of
   `PrettifyTreeprocessor` (#1261, #1263).
+* When running in html mode hidden tags with content show
+  up when they have content (#1343)
--- a/tests/test_apis.py
+++ b/tests/test_apis.py
@@ -506,6 +506,9 @@ class testSerializers(unittest.TestCase):
         p.text = 'foo <&escaped>'
         p.set('hidden', 'hidden')
         etree.SubElement(el, 'hr')
+
+        param = etree.SubElement(el, 'param')
+        param.text = "Param text"
         non_element = etree.SubElement(el, None)
         non_element.text = 'non-element text'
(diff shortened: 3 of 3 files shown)


shahmirj:
I should point out that the tests also fully pass:

```
$ python -m unittest discover tests
................................................................................................................................................s....................................................................................ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss...........................................................................ss..............................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................................
----------------------------------------------------------------------
Ran 980 tests in 2.106s

OK (skipped=71)
```

waylan:
I haven't yet taken a close look at your changes (I'll get to the when I have some time), but I did notice that you have modified some legacy tests. Note that any changes should be tested using the newer [markdown.test_tools.TestCase][1]. If the existing tests are in the legacy tests, then we move and reformat them to use the new framework as part of the change. Note that all new tests are under [tests/test_syntax](https://github.com/Python-Markdown/markdown/tree/master/tests/test_syntax). The only changes made to any other tests is to remove them.

shahmirj:
> I haven't yet taken a close look at your changes (I'll get to the when I have some time), but I did notice that you have modified some legacy tests. Note that any changes should be tested using the newer [markdown.test_tools.TestCase](https://python-markdown.github.io/test_tools/#markdowntest_toolstestcase). If the existing tests are in the legacy tests, then we move and reformat them to use the new framework as part of the change. Note that all new tests are under [tests/test_syntax](https://github.com/Python-Markdown/markdown/tree/master/tests/test_syntax). The only changes made to any other tests is to remove them.

Got it, I can look at the new test and make sure the new behaviour is added there as well, I will update my PR

markdown/serializers.py
```diff
@@ -162,8 +162,13 @@ def _serialize_html(write, elem, format):
                     write(_escape_cdata(text))
             for e in elem:
                 _serialize_html(write, e, format)
-            if tag.lower() not in HTML_EMPTY:
-                write("</" + tag + ">")
+            write("</" + tag + ">")
+        elif tag.lower() in HTML_EMPTY:
+            if format == "xhtml":
+                write(" /")
```
waylan: If you look at the coverage report, this line (168) never gets run. In fact, this case is caught in line 154. Perhaps the `html` case could be moved up there as well and we can remove this repeat check of `HTML_EMPTY`

waylan (changes requested):

markdown/serializers.py
shahmirj: I have removed this line, I ran coverage as you said and this was no longer needed.

shahmirj:
@waylan I can't find any test in the new framework that test the `output_format`, so I see the following options:

1. I have missed where this is being tested in the new framework, if so please let me know where to add the tests.
2. A test does not exist in the new framework, let me know where to add it and I can do that. If you choose this option, do you want me to delete the tests I added in legacy?
3. The test in legacy is fine as no tests exists in the new framework, and there is nothing to do.

Please let me know which direction you want to go and I am happy to oblige.

waylan:
I just reread your initial report and realize I misunderstood what you are trying to accomplish here.

> removing any tag with content that is in the `HTML_EMPTY`

By definition a tag in `HTML_EMPTY` cannot contain any content. Therefore, it should always render with any content removed. This is the correct expected behavior. I now realize that you are trying to change the behavior to have the content rendered. That would be incorrect and we will not be making that change.

Now if you have a reasonable argument for why a specific tag should not be in the `HTML_EMPTY` list, then we could consider removing a tag from the list. However, the tags in the list were put there based in the HTML spec, so you would need to point to a good reference to support any proposed change.

shahmirj:
That is fair, but my use case is producing XML, can we instead add an `output_format=xml` that allows tags to be tags, rather than HTML? Would this then be an acceptable solution?

And in xml tags with contents are written out, and HTML_EMPTY is not used when output_format=xml.

waylan:
We do not provide an XML serializer and we have no intention of doing so. However, if you want to provide your own you are welcome to with no need to alter the Markdown library at all. A quick hack might look something like this:

```python
md = markdown.Markdown()
md.output_formats['xml'] = your_serializer_function
md.set_output_format('xml')
xml = md.convert(src)
```

Be careful though, some extensions may change behavior based on which output format is selected. As you are changing the format after extensions have loaded, this could cause trouble. YMMV. Probably the better approach would be to subclass `markdown.Markdown` and redefine the `output_formats` attribute directly.

```python
class YourSubclass(Markdown):
    output_formats = {
        'html':   to_html_string,
        'xhtml':  to_xhtml_string,
        'xml': your_serializer_function
    }

md = YourSubclass(output_format='xml')
xml = md.convert(src)
```

shahmirj:
That works, happy with your proposal, I will close the PR. If I run into issues I can come back to you.

PR #1345, opened 2023-05-15
Add Badges / Tags extension

BitwiseAndrea:
## What
Add support for Badges (or tags - i'm not married to the name, but think it should be named after the function of the element in docs, not the general UI element like chip or pill) to python-markdown.

Associated Issue: https://github.com/Python-Markdown/markdown/issues/1346

## Why
Badges are very common UI elements in docs. They can be used for many purposes:

Showing relevant tags
image

Showing version numbers
image

Showing attributes of something in an easy-to-recognize way
image

Showing the status of an element
image

## How
Pattern matched to admonitions a bit, but updated to be an `InlineProcessor` since badges should be inline elements.

The format for making a badge is
```md
{{ badgeType "Badge Title" }}
```

and this will render the following html
```html
<span class="badge badgeType"><span class="badge-title>Badge Title</span></span>
```

## Testing
I could not for the life of me figure out how to test this locally other than in an interpreter so i just pushed to master every time and relied on the CI checks. Check out the contents of the test files to see whats covered, happy to add more cases.

### Mockup
I'm using mkdocs + mkdocs-material + admonitions for the project i would like to use this for and here is a mockup of what i would ideally like this to look like for that consumer, just to tie it all together.
image

This is the structure of the html - same as whats generated here, using a ::before element for the icon and just having two spans, one inside another.
image


BitwiseAndrea:
please don't judge my commit messages thank you

waylan:
I think this would be a great candidate for a third party extension. In other words, you are free to release this as an extension yourself. There is no need for this to be included with the Markdown library directly. Just be sure to add it to the [list][1] so others will find it.

[1]: https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions

PR #1350, opened 2023-06-04
Fixed Python 3.12 DeprecationWarning

vedranmiletic:
Testing an element's truth value will raise an exception in future versions.

### DIFF
--- a/markdown/extensions/admonition.py
+++ b/markdown/extensions/admonition.py
@@ -76,7 +76,7 @@ class AdmonitionProcessor(BlockProcessor):
             # to be under it, then the content's sibling is in the list.
             last_child = self.lastChild(sibling)
             indent = 0
-            while last_child:
+            while last_child is not None:
                 if (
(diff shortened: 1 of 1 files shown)


waylan:
Thanks for this.

For the record, the relevant section of the Python 3.12 release notes is [here][1].

> The [xml.etree.ElementTree](https://docs.python.org/3.12/library/xml.etree.elementtree.html#module-xml.etree.ElementTree) module now emits [DeprecationWarning](https://docs.python.org/3.12/library/exceptions.html#DeprecationWarning) when testing the truth value of an [xml.etree.ElementTree.Element](https://docs.python.org/3.12/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element). Before, the Python implementation emitted [FutureWarning](https://docs.python.org/3.12/library/exceptions.html#FutureWarning), and the C implementation emitted nothing.

[1]: https://docs.python.org/3.12/whatsnew/3.12.html#deprecated

waylan (approved):

PR #1354, opened 2023-06-29
Fix #1353: Update list of empty HTML tags.

### LINKED ISSUE #1353: Incomplete list of empty HTML tags leads to invalid HTML
I have made an extension to replace `` tags with `<figure>` and `<picture>` tags. This is the expected output:

```html
<picture>
    <source ...>
    <img ...>
</picture>
```

Unfortunately, the `<source>` element has a closing `</source>` tag. That's invalid HTML. I traced it to [this line](https://github.com/Python-Markdown/markdown/blob/1337d0891757e192165668d2606db36cf08e65a9/markdown/serializers.py#L44) in the code. The source element is missing from Python Markdown's internal list of empty elements.

The [full list of empty elements](https://developer.mozilla.org/en-US/docs/Glossary/Void_element) includes: 

```
<area>
<base>
<br>
<col>
<embed>
<hr>
<img>
<input>
<link>
<meta>
<param> (deprecated)
<source>
<track>
<wbr>
```

nicbou:
I left existing tags in, even the deprecated ones. This fixes #1353


waylan (changes requested):
The changes look good to me. However, we need a note added to the change log. 

I don't think we need to add a test. Current tests are [here](https://github.com/Python-Markdown/markdown/blob/1337d0891757e192165668d2606db36cf08e65a9/tests/test_apis.py#L498). We are currently only testing one empty tag (`hr`). That is probably sufficient. 

Note that we can ignore the failing pypy37 test env as that failure is related to the tooling (a tox call is raising an error). Our code never even runs. We will need to track that down separately.

waylan:
The [3.5 release notes](https://github.com/Python-Markdown/markdown/blob/master/docs/change_log/release-3.5.md) are now available. If a note is added there, this will be merged.

nicbou:
Thank you, I will try to fix it today. I was simply a bit confused about the requirement (the actual how-to-that part), and got distracted.

nicbou:
There you go! I updated the changelog. Thanks for your work and your patience.

waylan (approved):

PR #1356, opened 2023-07-17
Fix XML deprecations

### LINKED ISSUE #1355: DeprecationWarning: Testing an XML element's truth value will raise an exception in future versions
From a pytest run:

```python
              last_child = self.lastChild(sibling)
              indent = 0
  >           while last_child:
  E           DeprecationWarning: Testing an element's truth value will raise an exception in future versions.  Use specific 'len(elem)' or 'elem is not None' test instead.

  __pypackages__/3.12/lib/markdown/extensions/admonition.py:79: DeprecationWarning
```

From https://docs.python.org/3.12/whatsnew/3.12.html#deprecated:

> The [xml.etree.ElementTree](https://docs.python.org/3.12/library/xml.etree.elementtree.html#module-xml.etree.ElementTree) module now emits [DeprecationWarning](https://docs.python.org/3.12/library/exceptions.html#DeprecationWarning) when testing the truth value of an [xml.etree.ElementTree.Element](https://docs.python.org/3.12/library/xml.etree.elementtree.html#xml.etree.ElementTree.Element). Before, the Python implementation emitted [FutureWarning](https://docs.python.org/3.12/library/exceptions.html#FutureWarning), and the C implementation emitted nothing.

Also from https://docs.python.org/3.12/whatsnew/3.12.html#pending-removal-in-python-3-14:

> Testing the truth value of an [xml.etree.ElementTree.Element](https://docs.python.org/3.12/library./xml.etree.elementtree.html#xml.etree.ElementTree.Element) is deprecated and will raise an exception in Python 3.14.

Python Markdown version: 3.3.7.

Not sure if this is fixed in later versions.

facelessuser:
Resolves #1355

### DIFF
--- a/docs/change_log/release-3.4.md
+++ b/docs/change_log/release-3.4.md
@@ -111,3 +111,4 @@ The following bug fixes are included in the 3.4 release:
 * Extension entry-points are only loaded if needed (#1216).
 * Added additional checks to the `<pre><code>` handling of
   `PrettifyTreeprocessor` (#1261, #1263).
+* Fix XML deprecation warnings.
--- a/markdown/extensions/admonition.py
+++ b/markdown/extensions/admonition.py
@@ -78,14 +78,14 @@ class AdmonitionProcessor(BlockProcessor):
             indent = 0
             while last_child is not None:
                 if (
-                    sibling and block.startswith(' ' * self.tab_length * 2) and
-                    last_child and last_child.tag in ('ul', 'ol', 'dl')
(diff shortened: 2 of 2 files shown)


facelessuser:
pypy seems upset and cannot build...I'm not sure this is related to these changes as the tests never seem to run.

waylan (approved):

waylan:
Yeah, one of the pypy envs was failing previously. Now it seems that all of them are. Sigh.

Issue #1367, opened 2023-08-02
Footnotes numbering


alexvoss:
The numbering of footnotes follows the order in which the footnotes are defined rather than the order in which they are referenced in the document. This is somewhat surprising. 

To reproduce:

```python
#!/usr/bin/env python
import markdown

text = """
# Footnotes ordering

It seems that footnotes appear in the order in which they are
defined[^definedsecond] rather than in the order of their 
occurrence[^definedfirst] in the text.

[^definedfirst]: blah
[^definedsecond]: blurb
"""

print(markdown.markdown(text, extensions=['footnotes']))
```

(Naively) expected behaviour: the footnotes are numbered 1 then 2 but they come out as 2 then 1.

I do realize that changing the behavior could break things, so perhaps documenting this would be the way to go?

waylan:
Our [documentation](https://python-markdown.github.io/extensions/footnotes/#syntax) states:

> Python-Markdown’s Footnote syntax follows the generally accepted syntax of the Markdown community at large and almost exactly matches [PHP Markdown Extra](http://michelf.com/projects/php-markdown/extra/#footnotes)’s implementation of footnotes.

We have no intention of changing our behavior when it matches the generally accepted behavior across multiple implementations.

Regarding documenting the behavior, we don't actually document it. Instead we rely on PHP Markdown Extra's documentation to describe the behavior. Whether that should change is a question I have asked from time to time. In the end, we haven't made a change because no-one was willing to do the work. However, if someone was to submit a pull request which included documentation of the full behavior of our extension, I would be willing to review and guide it toward acceptance.

alexvoss:
Dear Waylan,

thanks for the prompt reply. I appreciate that changing the behavior might not be on the cards. The PHP Markdown Extra documentation states that 

—8<— 
Footnote definitions can be found anywhere in the document, but footnotes will always be listed in the order they are linked to in the text.
—8<— 

but it does not say anything about the order of the numbering. Going further down the chain to DaringFireball does not help either. 

I originally intended to file an issue with MkDocs Material but since I knew that this uses Python Markdown, I thought I’d go one step down in the list of dependencies. 

I can do a pull request that documents the current behavior.

Cheers,

Alex

---
Dr Alexander Voss
Freelance IT consultant and independent researcher
Rights-respecting software engineering
***@***.***
www.corealisation.com

radiofan:
Hello!
Numbering is really wrong...

I try convert this MD to HTML:
```
[...]

PR #1368, opened 2023-08-02
Alexvoss footnote numbering doc #1367

alexvoss:
As discussed, here is a small addition to the documentation that documents the way that the numbering of footnotes works. Fixed one small grammar issue as well.

### DIFF
--- a/docs/extensions/footnotes.md
+++ b/docs/extensions/footnotes.md
@@ -60,7 +60,7 @@ is indented consistently and any errors are more easily discernible by the autho
     A final paragraph.
 ```
 
-In the output, footnote are represented by superscript text in the places where 
+In the output, footnotes are represented by superscript text in the places where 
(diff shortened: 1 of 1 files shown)


waylan:
You seem to have misunderstood me. I had intended to either add full documentation of do nothing. I did not intend to simply add a single note on this one issue.

alexvoss:
Right, I did miss the word "full". Can you point me to any resources that help me understand what "full" means? Am happy to help. If there is nothing else to go by then I will need to go an study the implementation before I write anything - I guess?

waylan:
Essentially we would be mirror PHP's documentation of the syntax and behavior. Of course, we wouldn't want plagiarize their documentation so we would need to rewrite it differently. However, the information conveyed would essentially be the same with adjustments for any differences between the two implementations. And we could pull ideas from the documentation for other implementations as well.

PR #1371, opened 2023-08-07
Add PY3.12 & PYPY3.10 and drop 3.7

### LINKED ISSUE #1357: Python 3.12 tests currently fail
The error is:

```
  File "/Users/facelessuser/Code/github/markdown/tests/test_meta.py", line 20, in test__version__IsValid
    import packaging.version
ModuleNotFoundError: No module named 'packaging'

During handling of the above exception, another exception occurred:

  File "/Users/facelessuser/Code/github/markdown/tests/test_meta.py", line 22, in test__version__IsValid
    from pkg_resources.extern import packaging
ModuleNotFoundError: No module named 'pkg_resources'
```

This will require more investigation. 3.12 is still in beta. We will need to be fixed before we officially state support for 3.12.

waylan:
Also update default version for non-version specific tests (linters, etc) to PY-3.11 (previously used 3.7). Fixes #1357.

### DIFF
--- a/.github/workflows/version_check.yml
+++ b/.github/workflows/version_check.yml
@@ -1,0 +1,25 @@
+name: version_check
+
+on:
+  pull_request:
+    branches:
+    - '**'
+    paths:
+    # Only run when these files have been edited.
+    - 'markdown/__meta__.py'
+
+jobs:
+  check_version:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.12.0-rc.1'
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip packaging
+    - name: Run tests
+      run: |
+        python -m unittest tests.test_meta.TestVersion.test__version__IsValid
--- a/.github/workflows/version_check.yml
+++ b/.github/workflows/version_check.yml
@@ -16,7 +16,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v2
       with:
-        python-version: '3.12.0-rc.1'
+        python-version: '3.11'
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip packaging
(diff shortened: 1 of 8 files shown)


pyproject.toml
```diff
@@ -51,6 +51,7 @@ classifiers = [
 testing = [
     'coverage',
     'pyyaml',
+    'packaging'
```
waylan: I have initially added the `packaging` dependency. I just wanted to verify that that resolves the issue. I am not committed to this resolution. We may still drop the one test which requires it, in which case the dependency will be removed.

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,25 @@
+name: version_check
+
+on:
+  pull_request:
+    branches:
+    - '**'
+    paths:
+    # Only run when these files have been edited.
+    - 'markdown/__meta__.py'
```
waylan, line 9: This appears to be working correctly. In [this run][1], the job was defined but did not run as no changes had been made to `markdown/__meta__.py`. And then [the job ran successfully][2] after 9ac6fd3 included a change to `markdown/__meta__.py`.

[1]: https://github.com/Python-Markdown/markdown/actions/runs/5799096176/job/15718320332
[2]: https://github.com/Python-Markdown/markdown/actions/runs/5799199515/job/15718664775?pr=1371

.github/workflows/version_check.yml
```diff
@@ -0,0 +1,25 @@
...
+
+jobs:
+  check_version:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - name: Setup Python
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.12.0-rc.1'
```
waylan: I set this to `python-version: '3.12.0-rc.1'` to ensure it works correctly as that is the only version with the dependency issue. However, it doesn't really make sense to leave it this way. We could wait until 3.12 is released and update it then or we could set it to 3.11 for now. Thoughts?

waylan: I've reverted to 3.11 for now.

PR #1372, opened 2023-08-08
Update changelog workflow

waylan:
Update `mskelton/changelog-reminder-action` to v3 which removes the `include` option. Now we must define that option in GitHub's `path` option. This requires us to split the docs out into a separate workflow file.


waylan:
As far as I can tell, the `permissions` option is being ignored here, which is resulting in the action failing to even run. Not sure what to make of this.

waylan:
I am closing this in favor of #1373.

PR #1373, opened 2023-08-10
Refactor changelog

waylan:
I've been bothered by the format of our changelog for a while now. Finally was motivated to do something about it.

New changelog is at `docs/changelog.md` and uses the format defined at
    <https://keepachangelog.com/>. It only contins releases in the 3.x series.
    The old changelog at `docs/change_log/` remains in its old format as an
    archive and only contains older releases. The archive is removed from
    the site navigation and is only accessable from links in the new changelog.
    It is anticipated that only one major release (and its minor releases) will
    ever be in the new changelog at a time. Therefore, with the release of a
    future version 4.0, the 3.x changelog would be archived as well.

We could perhaps use the [Changelog Validate](https://github.com/marketplace/actions/changelog-validate) action, but it appears to require the filename to be all uppercase and in the root dir, which doesn't make sense for our documentation. A more versatile action would be needed. There is also the [Changelog Parser](https://github.com/marketplace/actions/changelog-parser) action which we could use to extract the notes for the newest release and insert them in the body of a new GitHub release in the deploy workflow. I have not tried that yet.

I did include the changes in #1372 (with adjustments), but they are subject to change. If we go this route, this would replace that PR.


waylan:
Some potentially better actions than the ones I mentioned above might be:

- https://github.com/marketplace/actions/changelog-enforcer
- https://github.com/mindsers/changelog-reader-action

docs/changelog.md
```diff
@@ -0,0 +1,718 @@
+title: Changelog
+toc_depth: 2
+
+# Python-Markdown Changelog
+
+All notable changes to this project will be documented in this file.
+
+The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
```
mitya57: That page says:

> A new version is available: [English 1.1.0](https://keepachangelog.com/en/1.1.0)

mitya57 (commented):
One minor comment, otherwise LGTM.

mitya57 (approved):

Issue #1374, opened 2023-08-11
Any tips on how to speed up build of a _very_ large table? (~1500 rows)


flynneva:
First off - thanks for the awesome project 🙏🏼 

I use the mkdocs-material + python-markdown integration quite heavily with the docs I build and I am running into a corner case which I'm not quite sure how best to improve / optimize.

We have a super large markdown table (~1500 rows with 3 columns, with md links in each cell) that is auto-generated and when trying to build with `mkdocs build`, it hangs specifically on the `python-markdown` bit and takes ~30min just to build that one file 😅 

The last debug print I see before the ~30min wait until the next file are some `Successfully imported extension` and `Successfully loaded extension` from `core.py` within this package:

```
 DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.toc.TocExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.tables.TableExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.fenced_code.FencedCodeExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.meta.MetaExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.admonition.AdmonitionExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.attr_list.AttrListExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.footnotes.FootnoteExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "markdown.extensions.md_in_html.MarkdownInHtmlExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:163  ] Successfully imported extension module "pymdownx.highlight".
[ DEBUG    ][ MARKDOWN ][ core.py:126  ] Successfully loaded extension "pymdownx.highlight.HighlightExtension".
[ DEBUG    ][ MARKDOWN ][ core.py:163  ] Successfully imported extension module "pymdownx.snippets".
[...]

Issue #954, opened 2020-05-05
reduce import/load logging


valentijnscholten:
Hi,

Can I suggest to adjust the logging of importing/loading extensions? When debugging issues, about 80% of my log files look like the below. Might it be sufficient to only log failed imports/loading of extensions? 

```
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.nl2br".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.nl2br.Nl2BrExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.sane_lists".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.sane_lists.SaneListExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.codehilite".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.codehilite.CodeHiliteExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.fenced_code".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.fenced_code.FencedCodeExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.toc".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.toc.TocExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.tables".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.tables.TableExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.nl2br".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.nl2br.Nl2BrExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.sane_lists".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.sane_lists.SaneListExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.codehilite".
uwsgi_1_820da895ad7a | Successfully loaded extension "markdown.extensions.codehilite.CodeHiliteExtension".
uwsgi_1_820da895ad7a | Successfully imported extension module "markdown.extensions.fenced_code".
[...]

PR #1375, opened 2023-08-11
Add some useful debug prints

flynneva:
- Changed a bunch of the code comments to debug prints instead to help when debugging

Relates to #1374

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -233,7 +233,7 @@ class Markdown:
         5. The output is written to a string.
 
         """
-
+        logger.debug("Converting markdown file to serialzed XHTML or HTML")
         # Fix up the source text
         if not source.strip():
             return ''  # a blank Unicode string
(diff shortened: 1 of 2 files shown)


waylan:
Not sure how this helps.

waylan:
Still not sure how this helps. In what workflow would I use this?

waylan:
Now that I have looked at #1374, I realize that you were using MkDocs and the debug messages were being printed by MkDocs logger. That is not the correct way to debug something like that would be to call Markdown directly passing in your sample Markdown source. And there are existing Python tools which can help to isolate which part of the code is taking all the time. Therefore, I am closing this as wontfix.

flynneva:
@waylan confused by your comment here about "using the mkdocs logger". The changes in this PR use the MARKDOWN logged already used in this file. The prints I copied to the issue were just formatted in a different way using the typical python logging API. No relation to mkdocs at all.

waylan:
Yes, you were using the Markdown logger, but from MkDocs', which is the point. You should have separated out the Markdown call and tested that individually. And using the method you used is like stabbing in the dark until you maybe hit your target. And in fact you did not hit the target. You hit the wrapper around your target and many others, which didn't really help. It was my guess based on years of experience dealing with these sorts of reports that narrowed down the issue.

In comparison, look at the discussion for #161, specifically [this comment](https://github.com/Python-Markdown/markdown/issues/161#issuecomment-10808038). With one short line of code I narrowed it down to the one method which was taking long. And then I was able to quickly build a test which narrowed the issue down further. The first line of code isn't provided, but I did mention [cProfile](https://docs.python.org/3/library/profile.html), which is in the Python Standard Library. I'm assuming I used some variation of `cProfile.run(markdown.markdown(some_input))`. That is a much quicker and more precise way of finding out what is causing a slowdown.

flynneva:
@waylan not really sure what you are saying here in relation to this PR. All it does is adds some debug prints....nothing more.

If you don't think it's helpful, that's ok with me 👍

waylan:
I meant to mention this before, but we already have users requesting less verbose output (see #954), so it would be worse to add more from their perspective. As there are other ways to debug the issue you had, then the best approach is to use those other ways, not to add more debug messages.

flynneva:
@waylan thats fine, and totally get it. Just that's a very different answer than what you said at first. I'll remove the debug prints from my other PR then, no worries! 👍

PR #1377, opened 2023-08-15
Add test case for large markdown table, add useful debug prints

flynneva:
- Adds a test case for a large markdown table with an HTML link before it
- Adds useful debug prints

By removing the blank space in the test between the HTML link and the markdown table, the test fails to complete without a very helpful warning / error.

I'd be up for improving the error that is reported if you are, just let me know if this is interesting for the project or not.

Relates to #1374

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -233,7 +233,7 @@ class Markdown:
         5. The output is written to a string.
 
         """
-
+        logger.debug("Converting markdown file to serialzed XHTML or HTML")
         # Fix up the source text
         if not source.strip():
             return ''  # a blank Unicode string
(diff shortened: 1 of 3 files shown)


waylan:
How does this test fail? My understanding of you previous report was that it simply took a long time for the parser to complete. That's not really the sort of test that makes sense to automate.

Yes, we could alter the code to raise an error, except that is against our policy.  See the first [goal][1] listed in our homepage. The only way we indicate to a document author that they provided bad input is by outputting unexpected output. That is how all Markdown implementations work.

[1]: https://python-markdown.github.io/#goals

flynneva:
@waylan this test fails when you remove the blank line between the link and the table. It passes just fine otherwise.

By having a test with a large table with valid formatting you can point future devs like me who say "does this package support super large tables?" to the test and say yes it does and builds fine. If this test existed when I made my original issue, you could just sent me a link to it and I would've been able to pretty quickly see what's different between the table source in the test and mine (the blank line just before the table).

I'm not saying you have to fix anything or merge this PR if you don't want to - you are the maintainer 😅 I am not.

waylan:
> you can point future devs like me who say "does this package support super large tables?" to the test and say yes it does and builds fine

I'm not likely to do that because I'm not likely to ever recall that such a test case exists. Our tests exist to ensure we don't introduce regressions when we make changes to the code. This test case does not seem to serve that function. Therefore, it is of no value and simply adds to the maintenance burden (it anything). While I appreciate your willingness to contribute, I will not be accepting this contribution.

PR #1380, opened 2023-09-02
Fix admonition corner case with indented code blocks

### LINKED ISSUE #1329: admonition extension returns **invalid** html code without blank line.
```python
In [8]: import markdown
In [9]: some_text = """
   ...: !!! danger "Danger"
   ...:     don't try this at home.
   ...:
   ...: """
In [10]: markdown.markdown(some_text, tab_length=2, extensions=['admonition'])
Out[10]: '<div class="admonition danger">\n<p class="admonition-title">Danger<p>don\'t try this at home.</p>\n</p>\n</div>'
```

1. A paragraph inside a paragraph is **invalid** html and *etree* parser complains!
2. the simple solution of turning the title *paragraph* into a *div* will work, but it wraps the inner paragraph, resulting in a CSS mess.

Recommended working solution:

```diff
     CLASSNAME_TITLE = 'admonition-title'
+    CLASSNAME_BODY = 'admonition-body'
     RE = re.compile(r'(?:^|\n)!!! ?([\w\-]+(?: +[\w\-]+)*)(?: +"(.*?)")? *(?:\n|$)')

@@ -130,9 +134,11 @@ class AdmonitionProcessor(BlockProcessor):
             div = etree.SubElement(parent, 'div')
             div.set('class', '{} {}'.format(self.CLASSNAME, klass))
             if title:
-                p = etree.SubElement(div, 'p')
+                p = etree.SubElement(div, 'div')
                 p.text = title
                 p.set('class', self.CLASSNAME_TITLE)
+                p = etree.SubElement(div, 'div')
+                p.set('class', self.CLASSNAME_BODY)

```

> **Note:**   the `parse_content()` method of your admonition extension is an unreadable code-mess ;-)

facelessuser:
Fix a corner case in admonitions where if an indented code block was provided as the first block, the output would be malformed.

Fixes #1329

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -28,12 +28,10 @@ links to the header, placing them after all other header content.
 * Update the list of empty HTML tags (#1353).
 * Add customizable TOC title class to TOC extension (#1293).
 
-## [3.4.5] -- ????-??-??
-
 ### Fixed
 
 * Fix a corner case in admonitions where if an indented code block was provided
-  as the first block, the output would be malformed.
+  as the first block, the output would be malformed (#1329).
 
 ## [3.4.4] -- 2023-07-25
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -28,6 +28,13 @@ links to the header, placing them after all other header content.
 * Update the list of empty HTML tags (#1353).
 * Add customizable TOC title class to TOC extension (#1293).
 
+## [3.4.5] -- ????-??-??
+
+### Fixed
+
+* Fix a corner case in admonitions where if an indented code block was provided
+  as the first block, the output would be malformed.
+
 ## [3.4.4] -- 2023-07-25
 
 ### Fixed
--- a/markdown/extensions/admonition.py
+++ b/markdown/extensions/admonition.py
@@ -69,7 +69,7 @@ class AdmonitionProcessor(BlockProcessor):
 
         sibling = self.lastChild(parent)
 
-        if sibling is None or sibling.get('class', '').find(self.CLASSNAME) == -1:
+        if sibling is None or sibling.tag != 'div' or sibling.get('class', '').find(self.CLASSNAME) == -1:
             sibling = None
         else:
             # If the last child is a list and the content is sufficiently indented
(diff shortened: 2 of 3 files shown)


docs/changelog.md
```diff
@@ -28,6 +28,13 @@ links to the header, placing them after all other header content.
 * Update the list of empty HTML tags (#1353).
 * Add customizable TOC title class to TOC extension (#1293).
 
+## [3.4.5] -- ????-??-??
```
waylan: I realize the changelog format is new for everyone here. But my understanding of the spec is that we never specify a version number until the actual release is made. Therefore, this line should be removed. The thinking is that there is no reason to be concerned over what the next version will be. Just add to the changelog as you make changes. Then, when release time comes, we do either a bug-fix, point, or major release depending on what we have accumulated. I'm not entirely sure we want to be that freeform about it.

Of course, I understand that what you were trying to accomplish is perhaps get this bugfix out before the next major release. But to do that we should fork master from the tag for 3.4.4 (to the 3.4 branch) and apply the patch to that. Of course, this would remain to get the patch applied to 3.5 as well.

That said, I am hopeful of getting 3.5 out immediately after Python 3.12 is released. As that is scheduled for October, there shouldn't be much wait. So if your okay waiting for 3.5, then remove this line and combine the item below with the list above.

docs/changelog.md
```diff
@@ -28,6 +28,13 @@ links to the header, placing them after all other header content.
...
+
+### Fixed
+
+* Fix a corner case in admonitions where if an indented code block was provided
+  as the first block, the output would be malformed.
```
waylan: Can we please note the issue where this was raised? Thank you.

waylan (changes requested):

docs/changelog.md
facelessuser: That's fine, I honestly didn't have any real plan for when it got released. I don't have an immediate need for this to be released, others may desire this fix sooner. I'll let you decide when to release it. I will update this.

docs/changelog.md
facelessuser: Yep

waylan (approved):

PR #1381, opened 2023-09-04
Deleting Redundant Newline Characters in the Python Syntax

shuishu:


facelessuser:
This completely alters the intention of the rule. The rule is meant to remove meaningless whitespace from an empty line, not collapse empty lines. This can alter and break other syntax. This should not be merged.

Issue #541, opened 2017-01-30
Wikilinks extension should support custom link text


clarfonthey:
For example, doing either `[Link][[Page]]` (Markdown style) or `[[Link|Page]]` (wiki style).

waylan:
Then you can just use a normal markdown link. I'm not interested in adding any features to this extension (If anything I would remove it). Of course, you (or anyone) are free to fork it and modify it however you like as long as you following the licence.

clarfonthey:
That's fair! I feel that removing/deprecating it is best if it doesn't support features like this, because otherwise it's not too useful. Thanks for the quick response.

PR #1382, opened 2023-09-25
Added support for alternative link text for wikilinks extension.

talamus:
This is a really simple change to the `wikilinks` extension that adds support for the alternative link text: \
<https://community.telligent.com/community/10/w/user-documentation/62141/how-do-i-use-wiki-link-syntax#Using_alternative_link_text>

The Finnish language uses inflection heavily, so most links need to have an alternative name:
```markdown
Lue täältä lisää [[ meidän koiramme | meidän koristamme! ]]
```

(If a thing like this looks suitable for a merge, I can add test cases and the documentation.)

### DIFF
--- a/markdown/extensions/wikilinks.py
+++ b/markdown/extensions/wikilinks.py
@@ -43,7 +43,7 @@ class WikiLinkExtension(Extension):
         self.md = md
 
         # append to end of inline patterns
-        WIKILINK_RE = r'\[\[([\w0-9_ -]+)\]\]'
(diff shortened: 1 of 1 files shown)


waylan:
This was previously proposed in #541 and rejected. I believe there are [existing third-party extensions](https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions#additional-tags) which support this feature already. I would suggest exploring those for one which meets your needs.

PR #1384, opened 2023-10-04
Add `set_css_classes` configuration option to the table extension

mikev-db:
A new boolean option `set_css_classes` adds css classes to the table generation. These classes include row and columns counts and parity and will allow for more granular table styling.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -20,6 +20,10 @@ will cause the links to be inserted at the start of the header, before any other
 header content. The default behavior for `permalink` is to append permanent
 links to the header, placing them after all other header content.
 
+#### Add `set_css_classes` configuration option to the table extension 
+
+A new boolean option `set_css_classes` adds css classes to the table generation. These classes include row and columns counts and parity and will allow for more granular table styling.
+
 ### Changed
 
 * Add support for cPython version 3.12 (and PyPy 3.10) and drop support for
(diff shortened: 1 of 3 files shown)


waylan:
Interesting suggestion. However, I'm curious why it has taken so many years for anyone to ask for this. Makes me wonder if it is necessary. Is there a way to accomplish what you want without these classes? If so they are redundant and unnecessary.

facelessuser:
Using things like `:nth-child()` for styling is really all that is needed. I guess adding classes could make it "easier", but I'd argue not necessary.

mikev-db:
@facelessuser I am using this in email templates, and gmail, outlook, etc do not support `:nth-child()`. I could not find any other way to accomplish this.  Thanks!

waylan:
If you search through past feature requests you will see that I am generally resistant to adding features to the tables extension. The extension provides the most basic table support. For anything more than that, a third party extension should be used (of which there are many). As this is not necessary in the common case, I am not inclined to add it.

mikev-db:
Alright. I'll hunt around for third-party extensions that might fit the bill. (Prefer not to maintain this as I'll be handing it off at some point). If either of you happen to know of one or a way around this issue, happy to hear it. I tried all the css I could think of but couldn't really make it happen.  Thanks @waylan, @facelessuser.

PR #1388, opened 2023-10-06
Update changelog validation

waylan:
My first attempt to release version 3.5 [failed](https://github.com/Python-Markdown/markdown/actions/runs/6435539821/job/17477010868) due to an invalid changelog file. We need to validate its format every time we confirm an edit is needed to ensure edits don't break the format and prevent a release.

### DIFF
--- a/.github/workflows/changelog.yml
+++ b/.github/workflows/changelog.yml
@@ -17,17 +17,17 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


waylan:
Note that the release succeeded to PYPI but failed to deploy to Github because of the changelog validation.

waylan:
Success! We are now validating the changelog. Unfortunately, it fails so I need to fix it.

waylan:
It works now!

Issue #1066, opened 2020-11-12
unclosed tag in code span


iamvdo:
Hello,

I'm facing a problem, not sure where the problem really is (I'm not a Python developer), but I managed to create a little use case showing it.

Using fresh install mkdocs (1.1.2) and markdown 3.3.3 (problem is not here with markdown 3.2.2) 

This `.md` file works
```md
A text

Another text `<?php`.

<div>
hello
</div>
```

Adding a new line, and it no longer works with error `AttributeError: 'NoneType' object has no attribute 'end'`

```md
A text

New line

Another text `<?php`.

<div>
hello
</div>
```

It also fails with other combinations of `code` tags containing HTML and/or PHP. But this one is the smaller use case I've found.

Thanks!

facelessuser:
There seem to be two problems here.

1. The line offset function expects to always match something, but if it doesn't, it fails. There are probably a couple of ways to fix this, but one way is simply to check for `None`:

    ```py
    @property
    def line_offset(self):
        """Returns char index in self.rawdata for the start of the current line. """
        if self.lineno > 1:
            m = re.match(r'([^\n]*\n){{{}}}'.format(self.lineno-1), self.rawdata)
            return 0 if m is None else m.end()
        return 0
    ```

2. After fixing this, another problem becomes apparent, at least with the second example. Usint this source:

    ```py
    import markdown

    src1 = """\
    A text

    Another text `<?php`.

    <div>
    hello
    </div>
    """

    src2 = """\
    A text

    New line

    Another text `<?php`.

    <div>
    hello
    </div>
    """


    def test(md_text):
        html = markdown.markdown(
            md_text,
            # Same result with or without md_in_html
            extensions=["markdown.extensions.md_in_html"],
        )
        print(html)


    test(src1)
    print('------')
    test(src2)
    ```

    We can see that in the second output the `<?php` content is posted twice.

    ```
    <p>A text</p>
    <p>New line</p>
    <p>Another text <code>&lt;?php</code>.</p>
    <p><div>
    hello
    &lt;?php`.</p>
    <div>
    ```

I'm not sure what the issue is here as I don't have time to dig further right now. It looks like we may be attempting to handle `<?php`, but we probably shouldn't be as it isn't at the start of a line, etc.

I would definitely argue this is a bug.

facelessuser:
@waylan while I believe that preprocessing instructions need some tweaks, but looking at this inline code case, I am slightly curious about your thoughts on this case:

```
[...]

PR #1392, opened 2023-10-20
Speedup line_offset property

eanorige:
When using `mkdocs`, I noticed huge runtimes and tracked these down to `HTMLExtractor.line_offset` taking all the time.  Before this patch, a run of `mkdocs` could take 500+ seconds, with this patch, it takes ~10 seconds.

Flamegraph before optimization:
mkdocs2

Flamegraph after optimization:
mkdocs2a

Note: It may be the case that the full cache of start-positions of all lines isn't needed, and just a single entry is sufficient; I didn't explore this optimization.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -8,6 +8,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html). See the [Contributing Guide](contributing.md) for details.
 
+## [unreleased]
+
+### Fixed
+
+* Fix a performance problem with HTML extraction where large HTML input could trigger quadratic line counting behavior (PR#1392).
+
 ## [3.5] -- 2023-10-06
 
 ### Added
--- a/markdown/htmlparser.py
+++ b/markdown/htmlparser.py
@@ -83,6 +83,9 @@ class HTMLExtractor(htmlparser.HTMLParser):
         # Block tags that should contain no content (self closing)
         self.empty_tags = set(['hr'])
 
+        self.lineno_start_cache = [0]
+
+
         # This calls self.reset
         super().__init__(*args, **kwargs)
         self.md = md
@@ -94,6 +97,8 @@ class HTMLExtractor(htmlparser.HTMLParser):
         self.stack = []  # When `inraw==True`, stack contains a list of tags
         self._cache = []
         self.cleandoc = []
+        self.lineno_start_cache = [0]
+
         super().reset()
 
     def close(self):
@@ -114,6 +119,15 @@ class HTMLExtractor(htmlparser.HTMLParser):
     @property
     def line_offset(self) -> int:
         """Returns char index in `self.rawdata` for the start of the current line. """
+        for ii in range(len(self.lineno_start_cache)-1, self.lineno-1):
+            last_line_start_pos = self.lineno_start_cache[ii]
+            lf_pos = self.rawdata.find('\n', last_line_start_pos)
+            if lf_pos == -1:
+                # No more newlines found. Use end of rawdata.
+                lf_pos = len(self.rawdata)
+            self.lineno_start_cache.append(lf_pos+1)
+
+        return self.lineno_start_cache[self.lineno-1]
         if self.lineno > 1 and '\n' in self.rawdata:
             m = re.match(r'([^\n]*\n){{{}}}'.format(self.lineno-1), self.rawdata)
             if m:
--- a/markdown/htmlparser.py
+++ b/markdown/htmlparser.py
@@ -85,7 +85,6 @@ class HTMLExtractor(htmlparser.HTMLParser):
 
         self.lineno_start_cache = [0]
 
-
         # This calls self.reset
         super().__init__(*args, **kwargs)
         self.md = md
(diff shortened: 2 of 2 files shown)


eanorige:
Thanks for triggering CI.  Based on its feedback, I have 

1. added the auto-requested changelog entry
2. removed the extra blank line accidentally introduced
3. reworded a comment.  

In rewording this comment, I realize that there is a slight change in behavior of this function when `self.lineno` is past the last line; the old implementation would return the position of the last '\n' in the input, while this implementation returns the length of the input (i.e. EOF).  If the exact backwards compatible behavior is needed, `lf_pos = len(self.rawdata)` should be changed to `= self.rawdata.rfind('\n')`.  Let me know what you think.

facelessuser:
@waylan I am not as familiar with the HTML parsing that you implemented, can a buffer in this situation change on you invalidating the cached offset?

> If the exact backwards compatible behavior is needed, lf_pos = len(self.rawdata) should be changed to = self.rawdata.rfind('\n'). Let me know what you think.

Generally, I would say that to have a better chance of getting merged quickly, backwards compatible would be better. When something backwards incompatible is introduced, we would need to be even more cautious to ensure this doesn't inadvertently break some other behavior that is going to cause us more time to debug later.

While the speedup is certainly nice, I would definitely want to ensure this hasn't introduced some unexpected parsing behavior that is going to break 3rd party plugins that aren't immediately obvious.

facelessuser:
I guess if the change could be somehow identified as inadvertent behavior, then it would make sense to not retain backwards compatibility.

waylan:
My memory is very hazy about this. That said, the code that this PR replaces is discussed in #1066 and/or #1068 and was committed in #1069. It appears that we added tests to cover the various edge cases.

> the old implementation would return the position of the last '\n' in the input, while this implementation returns the length of the input (i.e. EOF). If the exact backwards compatible behavior is needed, `lf_pos = len(self.rawdata)` should be changed to `= self.rawdata.rfind('\n')`. Let me know what you think.

Yes, we want the exact same behavior. It is possible that a file may not end with a newline and that would have been taken into consideration with the existing implementation. Therefore, we would need the method to always return the position of a `\n` exactly. Returning any other position would be an error. Perhaps we need to add a test for that.

waylan:
> can a buffer in this situation change on you invalidating the cached offset?

In short yes. See [this comment](https://github.com/Python-Markdown/markdown/issues/1066#issuecomment-726290648) (and the few that follow) for an explanation. As a reminder, we are using the standard library HTML parser to extract HTML from non-HTML and need to hack it to make it work with any non-HTML which contains angle brackets.

> Therefore, we would need the method to always return the position of a `\n` exactly.

Actually, what I should have said is that the method always must return the beginning of a line. EOF is only the beginning of a (empty) line if the buffer ends with a newline. If it does not end with a newline, the EOF is not the beginning of a line and it would be an error to return that position. In fact, looking at the example given in the above linked comment, it is possible that returning EOF could fail to address the purpose of this hack altogether.

facelessuser:
> In short yes. See https://github.com/Python-Markdown/markdown/issues/1066#issuecomment-726290648 (and the few that follow) for an explanation. As a reminder, we are using the standard library HTML parser to extract HTML from non-HTML and need to hack it to make it work with any non-HTML which contains angle brackets.

This poses a problem then. If the cache can be invalidated without us knowing, then caching is likely a bad idea.

waylan:
Sorry, I think I misread your question. No, the buffer doesn't change. However, the parser jumps too far ahead and we need to back it up. What is being cached is the last position. The new position we back it up to should always be after the last cached position. And now I'm wondering if I am wrong about what to do if no more newlines are found.

eanorige:
> Actually, what I should have said is that the method always must return the beginning of a line. 

The old behavior is actually a bit more complex than that.  Here's a sketch of the current (before my patch) implementation:

    When there's no newlines in the input file or lineno is <= 1, return 0
    Else if lineno is a valid line number, return the position of the first character *after* the newline
    Else return the position of the last newline.  

I suspect that there's a bug in the existing code and that we should return `rfind('\n')+1` in the last case to be consistent with the first two.  

ed: remove caching invalidation discussion

waylan:
> I suspect that there's a bug in the existing code and that we should return `rfind('\n')+1` in the last case to be consistent with the first two.

What is curious is why no tests are failing. Unless the character at the `+1` position is insignificant whitespace or something. I wondered if maybe the tests weren't triggering that code path, but the coverage report says they are. I guess I need to take a closer look at the tests. One or two might need to be tweaked to be sure it fails if we are off by one character.

waylan:
I just took another look at this. Specifically, the behavior when no newlines are found. I confirmed that we have no tests for this edge case. In fact, the existing code has a `# pragma: no cover` comment on that code block. If I remove the comment, that line shows up in the coverage report as missing.

I don't recall if I added the comment because I couldn't come up with a scenario which triggered that code path, or because I was focusing on other things and expected to come back to it later. In any event, we should probably try to come up with a test for this. If we do, then that should dictate how it should behave. If not, then I'm not going to be too concerned about it.

waylan:
So, the concern (how the HTML parser could get into the state where the line number is greater than the total number of lines) is described in [this comment](https://github.com/Python-Markdown/markdown/issues/1066#issuecomment-726290648). However, among various similar tests, the following test exists:

https://github.com/Python-Markdown/markdown/blob/4f0b91abe1a954a72ab3c99c3e2b880ab36631fa/tests/test_syntax/blocks/test_html_blocks.py#L1119-L1139

I just tried many other variations and I cannot get that condition. I expect that it would take a very unusual edge case (with very poorly formed HTML) to ever get that condition. Therefore, I'm not going to sweet it.

waylan (approved):

Issue #1389, opened 2023-10-09
Full type annotations


pawamoy:
Now that you added type annotations in many places, I was wondering if you would consider annotating the whole package, and mark it as typed with a `py.typed` file? For now users who want to type-check their code using Python-Markdown can install [`types-markdown`](https://pypi.org/project/types-Markdown/) which comes from [typeshed](https://github.com/python/typeshed/tree/main/stubs/Markdown/markdown).

waylan:
I am indifferent to this. I was willing to add type annotations for documentation purposes because I see the value. However, I see no value in type annotations for annotations sake. Probably because I'm an old-school Python dev who never had any annotations for many years, and well, they don't actually do anything... that is unless you use some tools that I don't use.

The point is that to me they are just noise in my code. But as that noise now has value for the documentation, I am willing to overlook it. However, if the noise isn't going to contribute to the documentation, then it's just an annoyance. Which means I am not likely to maintain and update it if/when the code changes. But if someone else wants to do the work, then they are certainly welcome to. 

I suppose that was a long-winded way of saying a PR is welcome. But I'm not putting any more effort into this personally.

pawamoy:
I understand, thanks for your answer :) I have no plans to work on this myself in the short term, but it's good to know you'd accept PRs and maintenance!

waylan:
A lot happened in other PRs since this issue was last updated. Of significance is [this comment](https://github.com/Python-Markdown/markdown/pull/1399#issuecomment-1791031687) of mine where I laid out more specific direction about my expectations regarding type annotations and whether this library should be fully type checked. In the ensuing discussion and updates to the PR it became clear that the type checker being used (mypy) could not be configured to meet my expectations (it was too strict) and the PR was closed as rejected.

In that same discssion I became convinced that it was also impossible for this library to be type complete (have a py.typed file). However, I have continued to question if that was correct. So today, I did some more research and found this [typing documentation](https://typing.readthedocs.io/en/latest/source/libraries.html#) (linked from the [typeshed](https://github.com/python/typeshed) project) which, among other things, spells out what constitutes [type completeness](https://typing.readthedocs.io/en/latest/source/libraries.html#id1) (see also the earlier section [How much of my library needs types?](https://typing.readthedocs.io/en/latest/source/libraries.html#how-much-of-my-library-needs-types). Significantly, my review of this document does not seem to suggest that my previous position was off the mark. Noteworthy among the things listed there are the following:

[...]

PR #1394, opened 2023-10-25
Improve type annotations (add more and fix wrong ones)

oprypin:
The origins of these are three-fold:

* Merging in stubs from https://github.com/python/typeshed/tree/main/stubs/Markdown using "merge-pyi"
   - Note: we can consider these annotations to be the important ones because it's what people have been adding according to their own need
* Double-checking around places where stubs were already added from the above, particularly conflicts with annotations that got added in this repository already
   + Taking the opportunity to declare a generic "Registry of T" class
* Running "mypy" and eliminating the most glaring errors it reported

### DIFF
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -317,11 +317,11 @@ class Registry(Generic[_T]):
         return iter([self._data[k] for k, p in self._priority])
 
     @overload
-    def __getitem__(self, key: str | int) -> _T:
+    def __getitem__(self, key: str | int) -> _T:  # pragma: no cover
         ...
 
     @overload
-    def __getitem__(self, key: slice) -> Registry[_T]:
+    def __getitem__(self, key: slice) -> Registry[_T]:  # pragma: no cover
         ...
 
     def __getitem__(self, key: str | int | slice) -> _T | Registry[_T]:
--- a/markdown/util.py
+++ b/markdown/util.py
@@ -27,13 +27,14 @@ from __future__ import annotations
 import re
 import sys
 import warnings
-from collections import namedtuple
 from functools import wraps, lru_cache
 from itertools import count
-from typing import TYPE_CHECKING, Any
+from typing import TYPE_CHECKING, Generic, Iterator, NamedTuple, TypeVar, overload
 
 if TYPE_CHECKING:  # pragma: no cover
     from markdown import Markdown
+
+_T = TypeVar('_T')
 
 
 """
@@ -108,7 +109,7 @@ def get_installed_extensions():
     return metadata.entry_points(group='markdown.extensions')
 
 
-def deprecated(message, stacklevel=2):
+def deprecated(message: str, stacklevel: int = 2):
     """
     Raise a [`DeprecationWarning`][] when wrapped function/method is called.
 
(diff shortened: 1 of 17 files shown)


waylan:
So there appears to be a couple code changes included in this (I don't mean adjustments to whitespace/line length). Could you please remove those? They should be submitted separately.

oprypin:
I only recognized one change that's clearly not about type *annotations* and reverted it.

I'm not sure if you also mean the NamedTuple change (which I think is on topic, as it's just the way to spell the same thing but with type annotations) or if there was something else.

markdown/util.py
```diff
@@ -298,39 +301,47 @@ class Registry:
...
         return item in self._data.values()
 
-    def __iter__(self):
+    def __iter__(self) -> Iterator[_T]:
         self._sort()
         return iter([self._data[k] for k, p in self._priority])
 
-    def __getitem__(self, key):
+    @overload
+    def __getitem__(self, key: str | int) -> _T:
```
waylan: Could you please add a ` # pragma: no cover` comment to these overloaded methods?

waylan:
I was thinking of the NameTuples, but you are correct, they do have type annotations. So nevermind on that.

markdown/util.py
oprypin:
```suggestion
    def __getitem__(self, key: str | int) -> _T:  # pragma: no cover
```

markdown/util.py
```diff
@@ -298,39 +301,47 @@ class Registry:
...
+        ...
+
+    @overload
+    def __getitem__(self, key: slice) -> Registry[_T]:
```
oprypin:
```suggestion
    def __getitem__(self, key: slice) -> Registry[_T]:  # pragma: no cover
```

waylan (approved):

waylan:
Thanks for this. Some of the wrong annotations I completely missed. 

I'll merge this after a changelog entry is added.

oprypin:
That is done

oprypin:
For future reference:

1. The main reason that nobody complained about the current type annotations is probably that they aren't actually used by type checkers. Unless a file named `py.typed` exists, the dominant type checker "mypy" ignores the types and asks you to install https://pypi.org/project/types-Markdown/ anyway.

2. If you are interested in a fully correct type-checked codebase (according to "mypy") as well as official external-facing type annotations support (by adding `py.typed`), let me know anytime, I can take up that task and help maintain this going forward. Otherwise that's all for now.

waylan:
Note that you have a conflict with the changelog (presumably with #1392).

Regarding a fully correct type-checked codebase (according to "mypy")  that is probably a reasonable goal. Personally, I haven't really explored what's involved given that this is a rather new thing for Python. So contributions are welcome.

oprypin:
> Note that you have a conflict with the changelog

Hm? I had already merged to resolve the conflict. There was just the spelling check failing

> Regarding a fully correct type-checked codebase (according to "mypy") that is probably a reasonable goal.

Great, so I'll continue looking into this soon (in a separate PR)

pawamoy:
This PR almost closes #1389 :slightly_smiling_face:

oprypin:
Oh I didn't see that!
No there are still like 100 mypy errors so maybe not  "almost" 😅

pawamoy:
Oh OK I thought there was just the `py.typed` file missing :sweat_smile:

Issue #1383, opened 2023-10-01
Language Server does not find extension attribute: `Cannot access member "Meta" for type "Markdown" Member "Meta" is unknown`


PlexSheep:
When using the meta extension, the `Meta` field of the `Markdown` class is not found by `Pyright`. Perhaps I missed something, can we import a stub file to make this error go away?

For clarity: The code works, but the language server does not find the `Meta` Member for type `Markdown`.

This might also happen with the attributes added by other extensions, but I've found it for the `meta` extension specifically.

**Minimal example**:
`test.py`:
```python
import markdown
MD = markdown.Markdown(extensions=['meta'])
with open("foo.md", "r") as f:
    content: str = f.read()
    html: str = MD.convert(content)
    meta = MD.Meta    # Cannot access member "Meta" for type "Markdown"    Member "Meta" is unknown
    print(meta)
```

`foo.md`:
```md
key: 19

this is the content
```

executing the script:
```
$ python test.py           
{'key': ['19']}
```

**Proposed Fix**
Add a python stub file for the `Meta` attribute.

waylan:
I can't replicate this. In fact, notice below that I have not even called `convert` yet. Simply loading the extension ensures that the `Meta` attribute exists.

```python
>>> md = markdown.Markdown(extensions=['meta'])
>>> md.Meta
{}
>>> markdown.__version__
'3.4.4'
```

facelessuser:
I don't think the argument is that `md.Meta` doesn't exist, only that LSP (Language Server Provider) in their editor can't automatically recognize `md.Meta` and its type.

As Python Markdown does not provide type annotations, and any that are available are provided by 3rd parties, I'm not sure if this is an issue for Python Markdown.  If we were currently providing type annotations, then maybe it could be argued we should do something, but currently, this is not a typed library, even if someone else has provided typing that you are using in conjunction with your LSP.

waylan:
Now I have reread your report and you note that things are working. But apparently some tool is inspecting the code and not finding the attribute. That is correct. Many extensions add their own attributes (most of them being third-party) and there is no way that we could possible anticipate or stub every possible attribute added by every extension. So, no we will not be giving this one extension special treatment.

That said, if anyone has suggestions for a general purpose solution, I am willing to consider it.

waylan:
> As Python Markdown does not provide type annotations

[...]

PR #1399, opened 2023-11-01
Fix all mypy typechecking errors, add a lot of type annotations

### LINKED ISSUE #1389: Full type annotations

oprypin:
* Closes #1389

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -167,18 +167,18 @@ class ListIndentProcessor(BlockProcessor):
     LIST_TYPES = ['ul', 'ol']
     """ Types of lists this processor can operate on. """
 
-    def __init__(self, *args):
+    def __init__(self, *args) -> None:
         super().__init__(*args)
         self.INDENT_RE = re.compile(r'^(([ ]{%s})+)' % self.tab_length)
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return block.startswith(' '*self.tab_length) and \
             not self.parser.state.isstate('detabbed') and \
             (parent.tag in self.ITEM_TYPES or
-                (len(parent) and parent[-1] is not None and
+                (len(parent) > 0 and parent[-1] is not None and
                     (parent[-1].tag in self.LIST_TYPES)))
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         level, sibling = self.get_level(parent, block)
         block = self.looseDetab(block, level)
@@ -251,10 +251,10 @@ class ListIndentProcessor(BlockProcessor):
 class CodeBlockProcessor(BlockProcessor):
     """ Process code blocks. """
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return block.startswith(' '*self.tab_length)
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         sibling = self.lastChild(parent)
         block = blocks.pop(0)
         theRest = ''
@@ -286,10 +286,10 @@ class BlockQuoteProcessor(BlockProcessor):
 
     RE = re.compile(r'(^|\n)[ ]{0,3}>[ ]?(.*)')
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return bool(self.RE.search(block)) and not util.nearing_recursion_limit()
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         m = self.RE.search(block)
         if m:
@@ -353,10 +353,10 @@ class OListProcessor(BlockProcessor):
         self.INDENT_RE = re.compile(r'^[ ]{%d,%d}((\d+\.)|[*+-])[ ]+.*' %
                                     (self.tab_length, self.tab_length * 2 - 1))
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return bool(self.RE.match(block))
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         # Check for multiple items in one block.
         items = self.get_items(blocks.pop(0))
         sibling = self.lastChild(parent)
@@ -462,10 +462,10 @@ class HashHeaderProcessor(BlockProcessor):
     # Detect a header at start of any line in block
     RE = re.compile(r'(?:^|\n)(?P<level>#{1,6})(?P<header>(?:\\.|[^\\])*?)#*(?:\n|$)')
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return bool(self.RE.search(block))
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         m = self.RE.search(block)
         if m:
@@ -493,10 +493,10 @@ class SetextHeaderProcessor(BlockProcessor):
     # Detect Setext-style header. Must be first 2 lines of block.
     RE = re.compile(r'^.*?\n[=-]+[ ]*(\n|$)', re.MULTILINE)
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return bool(self.RE.match(block))
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         lines = blocks.pop(0).split('\n')
         # Determine level. `=` is 1 and `-` is 2.
         if lines[1].startswith('='):
@@ -519,7 +519,7 @@ class HRProcessor(BlockProcessor):
     # Detect hr on any line of a block.
     SEARCH_RE = re.compile(RE, re.MULTILINE)
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         m = self.SEARCH_RE.search(block)
         if m:
             # Save match object on class instance so we can use it later.
@@ -527,7 +527,7 @@ class HRProcessor(BlockProcessor):
             return True
         return False
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         match = self.match
         # Check for lines in block before `hr`.
@@ -547,10 +547,10 @@ class HRProcessor(BlockProcessor):
 class EmptyBlockProcessor(BlockProcessor):
     """ Process blocks that are empty or start with an empty line. """
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return not block or block.startswith('\n')
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         filler = '\n\n'
         if block:
@@ -577,10 +577,10 @@ class ReferenceProcessor(BlockProcessor):
         r'^[ ]{0,3}\[([^\[\]]*)\]:[ ]*\n?[ ]*([^\s]+)[ ]*(?:\n[ ]*)?((["\'])(.*)\4[ ]*|\((.*)\)[ ]*)?$', re.MULTILINE
     )
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return True
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> bool:
         block = blocks.pop(0)
         m = self.RE.search(block)
         if m:
@@ -603,10 +603,10 @@ class ReferenceProcessor(BlockProcessor):
 class ParagraphProcessor(BlockProcessor):
     """ Process Paragraph blocks. """
 
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return True
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> None:
         block = blocks.pop(0)
         if block.strip():
             # Not a blank block. Add to parent, otherwise throw it away.
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -121,7 +121,7 @@ class BlockProcessor:
             parent: An `etree` element which will be the parent of the block.
             block: A block of text from the source which has been split at blank lines.
         """
-        pass  # pragma: no cover
+        raise NotImplementedError()  # pragma: no cover
 
     def run(self, parent: etree.Element, blocks: list[str]) -> bool | None:
         """ Run processor. Must be overridden by subclasses.
@@ -147,7 +147,7 @@ class BlockProcessor:
             parent: An `etree` element which is the parent of the current block.
             blocks: A list of all remaining blocks of the document.
         """
-        pass  # pragma: no cover
+        raise NotImplementedError()  # pragma: no cover
 
 
 class ListIndentProcessor(BlockProcessor):
@@ -417,7 +417,7 @@ class OListProcessor(BlockProcessor):
 
     def get_items(self, block: str) -> list[str]:
         """ Break a block into list items. """
-        items = []
+        items: list[str] = []
         for line in block.split('\n'):
             m = self.CHILD_RE.match(line)
             if m:
@@ -426,7 +426,9 @@ class OListProcessor(BlockProcessor):
                 if not items and self.TAG == 'ol':
                     # Detect the integer value of first list item
                     INTEGER_RE = re.compile(r'(\d+)')
-                    self.STARTSWITH = INTEGER_RE.match(m.group(1)).group()
+                    int_match = INTEGER_RE.match(m.group(1))
+                    assert int_match is not None
+                    self.STARTSWITH = int_match.group()
                 # Append to the list
                 items.append(m.group(3))
             elif self.INDENT_RE.match(line):
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -23,7 +23,7 @@ import codecs
 import sys
 import logging
 import importlib
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Mapping, Sequence, TextIO
+from typing import TYPE_CHECKING, Any, BinaryIO, Callable, ClassVar, Mapping, Sequence
 from . import util
 from .preprocessors import build_preprocessors
 from .blockprocessors import build_block_parser
@@ -85,7 +85,7 @@ class Markdown:
     callable which accepts an [`Element`][xml.etree.ElementTree.Element] and returns a `str`.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs: Any):
         """
         Creates a new Markdown instance.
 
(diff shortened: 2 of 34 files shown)


markdown/core.py
```diff
@@ -85,7 +86,11 @@ class Markdown:
     callable which accepts an [`Element`][xml.etree.ElementTree.Element] and returns a `str`.
     """
 
-    def __init__(self, **kwargs):
+    toc_tokens: list[TocToken]
+    toc: str
+    Meta: dict[str, Any]
```
waylan: Why are these here? They only exist via extensions. The core knows nothing of these extensions and should under no circumstances define these.

This is important because third party extensions might want to do something similar and it it not reasonable for us to provide the same for them. If this is a problem, then we need a different solution. This was discussed in depth elsewhere (can't seem to find it now).

markdown/blockprocessors.py
```diff
@@ -426,7 +426,9 @@ def get_items(self, block: str) -> list[str]:
                 if not items and self.TAG == 'ol':
                     # Detect the integer value of first list item
                     INTEGER_RE = re.compile(r'(\d+)')
-                    self.STARTSWITH = INTEGER_RE.match(m.group(1)).group()
+                    int_match = INTEGER_RE.match(m.group(1))
+                    assert int_match is not None
```
waylan, line 430: Not sure what you are trying to accomplish here. Does this make it possible for Markdown content to cause an error? If so, that would be unacceptable.

markdown/core.py
```diff
@@ -442,18 +445,14 @@ def convertFile(
                 output_file.close()
             else:
                 writer = codecs.getwriter(encoding)
-                output_file = writer(output, errors="xmlcharrefreplace")
-                output_file.write(html)
+                output_writer = writer(output, errors="xmlcharrefreplace")
+                output_writer.write(html)
```
waylan, line 444: Not sure why this change was made at all. Note that the [Contributing Guide](https://python-markdown.github.io/contributing/#code-style-guide) states:

I realize that in this instance the variable name is not exposed outside of this method so there is no concern over a backward incompatible change, but the general principle remains. Please, let's refrain from making unnecessary changes just for personal preference.

markdown/core.py
```diff
@@ -442,18 +445,14 @@ def convertFile(
...
-            html = html.encode(encoding, "xmlcharrefreplace")
-            try:
-                # Write bytes directly to buffer (Python 3).
-                sys.stdout.buffer.write(html)
-            except AttributeError:  # pragma: no cover
-                # Probably Python 2, which works with bytes by default.
-                sys.stdout.write(html)
+            html_bytes = html.encode(encoding, "xmlcharrefreplace")
+            # Write bytes directly to buffer (Python 3).
+            sys.stdout.buffer.write(html_bytes)
```
waylan: Again, I agree that this is a reasonable update to old code, but it is out-of-scope for this PR.

markdown/core.py
```diff
@@ -489,7 +488,13 @@ def markdown(text: str, **kwargs: Any) -> str:
...
 
 
-def markdownFromFile(**kwargs: Any):
+def markdownFromFile(
+    *,
+    input: str | BinaryIO | None = None,
+    output: str | BinaryIO | None = None,
+    encoding: str | None = None,
+    **kwargs: Any
+) -> None:
```
waylan, line 491: I don't recall why the specific keyword parameters were removed some years ago, but why did you feel the need to add them back in? I'm not opposed to it if there is a good reason related to this scope of this PR, but I would like to here the reason.

However, I am more concerned about why you added position arguments (*)?

markdown/extensions/admonition.py
```diff
@@ -70,11 +75,11 @@ def parse_content(self, parent, block):
 
         # We already acquired the block via test
         if self.current_sibling is not None:
-            sibling = self.current_sibling
+            prev_sibling = self.current_sibling
             block, the_rest = self.detab(block, self.content_indent)
             self.current_sibling = None
             self.content_indent = 0
-            return sibling, block, the_rest
+            return prev_sibling, block, the_rest
```
waylan, line 82: Again, we are making unnecessary out-of-scope changes.

markdown/extensions/admonition.py
```diff
@@ -143,6 +148,7 @@ def run(self, parent, blocks):
                 p.text = title
                 p.set('class', self.CLASSNAME_TITLE)
         else:
+            assert sibling is not None
```
waylan, line 151: Can this raise an error based on Markdown input? If so, this is unacceptable.

markdown/extensions/codehilite.py
```diff
@@ -258,14 +266,16 @@ def code_unescape(self, text):
...
 
-    def run(self, root):
+    def run(self, root: etree.Element) -> None:
         """ Find code blocks and store in `htmlStash`. """
         blocks = root.iter('pre')
         for block in blocks:
             if len(block) == 1 and block[0].tag == 'code':
                 local_config = self.config.copy()
+                text = block[0].text
+                assert text is not None
```
waylan, line 276: Another assert....

markdown/extensions/def_list.py
```diff
@@ -33,13 +37,14 @@ class DefListProcessor(BlockProcessor):
...
-    def test(self, parent, block):
+    def test(self, parent: etree.Element, block: str) -> bool:
         return bool(self.RE.search(block))
 
-    def run(self, parent, blocks):
+    def run(self, parent: etree.Element, blocks: list[str]) -> bool | None:
 
         raw_block = blocks.pop(0)
         m = self.RE.search(raw_block)
+        assert m is not None
```
waylan, line 47: another assert...

markdown/extensions/def_list.py
```diff
@@ -53,20 +58,21 @@ def run(self, parent, blocks):
...
+                blocks.insert(0, raw_block)
+                return False
+            if sibling.tag == 'p':
+                # The previous paragraph contains the terms
+                state = 'looselist'
+                assert sibling.text is not None
+                terms = sibling.text.split('\n')
+                parent.remove(sibling)
+                # Acquire new sibling
+                sibling = self.lastChild(parent)
```
waylan, line 75: This appears to be out-of-scope as well.

waylan (changes requested):
I got much less that half way through this and keep seeing the same issues reoccurring and gave up. See my comments below and please make the adjustments everywhere, not just where I've noted them.

markdown/blockprocessors.py
oprypin: With lines  429-431 reverted:

```console
$ mypy markdown
markdown/blockprocessors.py:429: error: Item "None" of "Match[str] | None" has no attribute "group"  [union-attr]
Found 1 error in 1 file (checked 33 source files)
```

---

* The code before:
  ```python
  self.STARTSWITH = RE.match(string).group()
  ```

  where `RE.match(string)` can be `None`, and `None.group()` is an error

* The code after:
  ```python
  int_match = RE.match(string)
  assert int_match is not None
  self.STARTSWITH = int_match.group()
  ```

The old code doesn't care to check for the `None` case where it would violently error with `AttributeError`. mypy doesn't let it slide and exposes a potential bug.

The new code directly checks for it and errors with a clearer message.

There is no change regarding which situations an error does or does not happen.

*TBD: It is worth checking whether this bug case can actually happen (in current code already). In any case we should be happy that mypy flagged this.*

markdown/core.py
oprypin: This is the pragmatic approach. The only other option is to drop these and ignore type errors and all type checks whenever these are referenced - this is both in this codebase and in any dependant codebases.
(Which is particularly painful because the precise type checks that I'm adding for `toc_tokens` become unusable :smiling_face_with_tear: )

The type checking solutions have definitively decided to not support the use case of setting random unrecognized attributes on objects.

https://stackoverflow.com/a/61218152

We should just say that non-core extensions should not set new attributes on Markdown. They can set attributes on the extension itself or on a treeprocessor, and the users of the code should be able to access it through the extension object.

---

With this reverted:

```console
$ mypy markdown
markdown/extensions/toc.py:356: error: "Markdown" has no attribute "toc_tokens"  [attr-defined]
markdown/extensions/toc.py:357: error: "Markdown" has no attribute "toc"  [attr-defined]
markdown/extensions/toc.py:425: error: "Markdown" has no attribute "toc"  [attr-defined]
markdown/extensions/toc.py:426: error: "Markdown" has no attribute "toc_tokens"  [attr-defined]
markdown/extensions/meta.py:52: error: "Markdown" has no attribute "Meta"  [attr-defined]
markdown/extensions/meta.py:86: error: "Markdown" has no attribute "Meta"  [attr-defined]
Found 6 errors in 2 files (checked 33 source files)
```

markdown/core.py
oprypin: 100% of code body changes in this pull request are in response to error messages produced by mypy. Don't need to tell me this regarding unrelated changes because I'm always the first one to say this as well.

This particular change is because mypy doesn't like that these two differently-typed values share the same variable name.

---

With lines 425-429 reverted:
```console
$ mypy markdown                                             
markdown/core.py:427: error: Incompatible types in assignment (expression has type "StreamReader", variable has type "StreamReaderWriter")  [assignment]
Found 1 error in 1 file (checked 33 source files)
```

markdown/core.py
oprypin: Again I agree and we can make this change earlier in the chain. In fact yes, let's do that in https://github.com/Python-Markdown/markdown/pull/1400

But this change **was** made directly in response to mypy errors. You can run check out the code locally, revert this part, and observe the error message.

---

With lines 453-455 reverted:
```console
$ mypy markdown
markdown/core.py:453: error: Incompatible types in assignment (expression has type "bytes", variable has type "str")  [assignment]
markdown/core.py:456: error: No overload variant of "write" of "IO" matches argument type "str"  [call-overload]
markdown/core.py:456: note: Possible overload variants:
markdown/core.py:456: note:     def write(self, Buffer, /) -> int
markdown/core.py:456: note:     def write(self, bytes, /) -> int
Found 2 errors in 1 file (checked 33 source files)
```

markdown/core.py
oprypin: The function's signature doesn't change, I'm just formalizing the de-facto signature. The `*` is to start named-only parameters, not positional-only parameters. They were also named-only before.

markdown/extensions/admonition.py
oprypin: This change was made in response to mypy errors. Again this is because mypy doesn't like that a variable name is reused with different types in different places. Which is really a limitation of mypy, most other type checkers would deal with this perfectly fine.

---

With lines 78-82 reverted:

```console
$ mypy markdown
markdown/extensions/admonition.py:84: error: Incompatible types in assignment (expression has type "Element | None", variable has type "Element")  [assignment]
markdown/extensions/admonition.py:87: error: Incompatible types in assignment (expression has type "None", variable has type "Element")  [assignment]
markdown/extensions/admonition.py:101: error: Incompatible types in assignment (expression has type "Element | None", variable has type "Element")  [assignment]
markdown/extensions/admonition.py:113: error: Incompatible types in assignment (expression has type "None", variable has type "Element")  [assignment]
```

markdown/extensions/admonition.py
oprypin: The code already has a problem where `sibling` might be None and `sibling.tag` will be an error. mypy correctly detects and reports it, and requires us to make an explicit assertion, which indicates that we either know that this will never actually happen, or perhaps don't care if it does but then the error will be more explicit.

*TBD: It is worth checking whether this bug case can actually happen (in current code already). In any case we should be happy that mypy flagged this.*

---

With line 151 reverted:
```console
$ mypy markdown
markdown/extensions/admonition.py:152: error: Item "None" of "Element | None" has no attribute "tag"  [union-attr]
markdown/extensions/admonition.py:152: error: Item "None" of "Element | None" has no attribute "text"  [union-attr]
markdown/extensions/admonition.py:153: error: Item "None" of "Element | None" has no attribute "text"  [union-attr]
markdown/extensions/admonition.py:154: error: Item "None" of "Element | None" has no attribute "text"  [union-attr]
markdown/extensions/admonition.py:155: error: Argument 1 to "SubElement" has incompatible type "Element | None"; expected "Element"  [arg-type]
markdown/extensions/admonition.py:158: error: Incompatible types in assignment (expression has type "Element | None", variable has type "Element")  [assignment]
```

markdown/extensions/def_list.py
oprypin: Another place where the code currently assumes that the match will happen. mypy doesn't let it slide. I see here that it is 100% safe to assume that the match actually happens, based on the outcome of `test`. But it is OK that mypy forces us to write out this assumption explicitly, both so the writer double-checks this and that the reader *doesn't* need to think that maybe this is a bug.

---

With line 47 reverted:
```console
$ mypy markdown
markdown/extensions/def_list.py:48: error: Item "None" of "Match[str] | None" has no attribute "start"  [union-attr]
markdown/extensions/def_list.py:49: error: Item "None" of "Match[str] | None" has no attribute "end"  [union-attr]
markdown/extensions/def_list.py:56: error: Item "None" of "Match[str] | None" has no attribute "group"  [union-attr]
markdown/extensions/def_list.py:58: error: Item "None" of "Match[str] | None" has no attribute "group"  [union-attr]
Found 4 errors in 1 file (checked 33 source files)
```

markdown/extensions/def_list.py
oprypin: This is again in response to mypy errors, unfortunately these kind of gymnastics are necessary sometimes to make it happy.

---

With lines 61-75 reverted:

---

```console
$ mypy markdown
markdown/extensions/def_list.py:66: error: Item "None" of "Element | None" has no attribute "tag"  [union-attr]
markdown/extensions/def_list.py:69: error: Item "None" of "Element | None" has no attribute "text"  [union-attr]
markdown/extensions/def_list.py:69: error: Item "None" of "str | Any | None" has no attribute "split"  [union-attr]
markdown/extensions/def_list.py:70: error: Argument 1 to "remove" of "Element" has incompatible type "Element | None"; expected "Element"  [arg-type]
```

markdown/extensions/codehilite.py
oprypin: Another place where the code currently assumes that something is present there and violently errors if there's no match. mypy doesn't let it slide.

*TBD: It is worth checking whether this bug case can actually happen (in current code already). In any case we should be happy that mypy flagged this.*

---

With lines 275-278 reverted:

```console
$ mypy markdown
markdown/extensions/codehilite.py:276: error: Argument 1 to "code_unescape" of "HiliteTreeprocessor" has incompatible type "str | None"; expected "str"  [arg-type]
Found 1 error in 1 file (checked 33 source files)
```

oprypin:
> See my comments below and please make the adjustments everywhere, not just where I've noted them.

The adjustments would basically mean reverting the changes, but almost none of them can be reverted, because then mypy errors appear.

So this would just mean giving up trying to check the code with mypy.
Maybe there is some misconception, but the point of running the actual mypy checks is that it also checks function bodies and can complain about potential problems until you write code in a way that makes it clear to mypy that there is no problem.

markdown/core.py
waylan: > The type checking solutions have definitively decided to not support the use case of setting random unrecognized attributes on objects.
>
> ...
>
> We should just say that non-core extensions should not set new attributes on Markdown.

This is a problem because I expect to be able to do this. If there is no workaround, then we might as well forgot about ever completing this.

I found the previous discssion on this in #1383. Note in particular my comments about setters and getters. Perhaps this discussion should be continued there.

markdown/blockprocessors.py
waylan: So, now we know about a bug which we didn't know about before. That is good. But this is not the way to address it. Under no circumstances should source text cause Markdown to raise an error. In fact, that is one of the primary goals of the project as documented on the [home page](https://python-markdown.github.io/#goals). Therefore, this is not the proper way to fix the bug. The error needs to be silenced and some reasonable text needs to be included in the output (depending on the conditions under which the issue arises).

markdown/core.py
waylan: I embrace Python not being strongly typed. If that means making changes like this, then no thank you.

oprypin:
* Extracted simpler changes to https://github.com/Python-Markdown/markdown/pull/1401

waylan:
> Maybe there is some misconception

Python not being strongly typed if a feature which I embrace. I will continue to write my code that way. Anything which forces me to write strongly typed code in a language which is not strongly typed is unwelcome.

I see type annotations as documentation. The code dictates the annotations, not the other way around. Of course, over time as updates and changes are made to the code, the annotations could get out-of-sync with the code. Therefore, it is reasonable to include a type checker in the CI to ensure that doesn't happen. However, the type checker should only be checking the already defined annotations. There is no need to type check an unannotated private variable which is never used outside of a single method (for example).

If mypy cannot serve that purpose, then it is the wrong tool for the job.

waylan:
I am going to close this in favor of #1401. I suspect that mypy is not the right fit for this project. However, if it can be configured to fit, then this can be reopened. Or, if a different tool is proposed, that should be in a new PR.

oprypin:
I'll want to sync this branch to latest if #1401 is merged, and then leave this for posterity. Not sure if I'll be able to do that if it's closed.

oprypin:
Yeah mypy is pretty bad!
Sadly other checkers would ask for even more code changes :D

waylan:
Okay, I'll leave it open for now.

oprypin:
By disabling a large amount of mypy checks, I reduced this to a much smaller change #1402 that is based on #1401.
It will ensure the type annotations don't fall into disrepair.

facelessuser:
I think if you are using types, using mypy is very important to keep things from getting broken. I also agree that there is nothing wrong with limiting the checks and/or doing per line ignores for particularly tricky things that can be resolved later if there is interest.

waylan:
I have taken a step back and taken a fresh look at the overall matter of annotations and type checking. I think perhaps I dove in with a few misunderstandings and miscomprehensions. Sorry about that; especially if I have led anyone down a path of doing work I would never approve. That was not my intention.

So to be clear. upon reflection, the following are my general expectations.

1. **This library should always be a dynamically typed library.** I am opposed to this library being forcibly strongly typed. That probably means we will never have a `py.typed` file (or whatever it is called) and that should not be a goal to work towards.

2. The only purpose of annotations is to serve as documentation. The code dictates the annotations, not the other way around. In other words, we do not edit code to fit the annotations, we edit the annotations to fit the code. Of course, occasionally, the specific format of code may prevent defining annotations. In that case, an edit to the format (but not functionality) of the code is acceptable to allow  defining annotations (the recent edits to some `NamedTuple` objects is one such example of this).

    Of course, type checking may occasionally reveal an otherwise unknown bug in the code. And it would only be reasonable for those bugs to be fixed. However, the fix should always be framed as a fix to the now known bug, not a change to the code to match the type annotations. In other words, I would expect a test case which fails before the fix is applied and succeeds after regardless of any type annotations. 

3. Of course, the type annotations should remain in sync with the code. I would expect some tooling to check that on the CI server. Presumably, we would add another linter like tool.  **The tool should never error or warn on missing annotations.** The only task of the tool is to verify that the type annotations which exist agree with the code. If any fail, then a PR would not be accepted until an appropriate fix is made. Of course, an appropriate fix could mean the annotation needs to be updated, or it could mean the code needs to be updated. Is this an intentional change to the type of an object or did the type get changed inadvertently?

There are a few things we will need to do to meet the above goals. 

Item 1 above requires no action expect to perhaps remove some proposed changes.

Item 2 is mostly complete. Thanks for the hard work of everyone involved.

Item 3 seems to be the problem. Originally I thought that mypy could meet our needs. After all, the page entitled [*Using mypy with an existing codebase*](https://mypy.readthedocs.io/en/latest/existing_code.html#existing-code) in their documentation talks about how the tool can be used to progressively work through your code. However, on closer inspection, I see that the idea is to work through the code to gradually make it strongly typed. In fact, the suggestion is to do so before any annotations are added. That seems to suggest that perhaps mypy can only check annotations if the code is strongly typed. That is not a useful tool for our purpose. Unfortunately, I do not have any answers for how we can address this item. It occurs to me that it might not actually be possible to accomplish what I am proposing here.

oprypin:
> I'll want to sync this branch to latest if https://github.com/Python-Markdown/markdown/pull/1401 is merged, and then leave this for posterity. Not sure if I'll be able to do that if it's closed.

I did the merge, can be closed now :D

PR #1400, opened 2023-11-01
Correct the declarations of I/O buffers as bytes-based

oprypin:
* Correctly declare the parameter annotations as `BinaryIO`, because in fact it works only with byte-based buffers.
   
   Attempt at running code both **before** and after this change:

  ```pycon
  >>> from markdown import Markdown

  >>> from io import StringIO, BytesIO
  
  >>> Markdown().convertFile(BytesIO(b'input'), output=BytesIO())
  <markdown.core.Markdown object at 0x7fe3bba7c710>

  >>> Markdown().convertFile(StringIO('input'), output=BytesIO())
    File "markdown/core.py", line 423, in convertFile
      text = input_file.read()
            ^^^^^^^^^^^^^^^^^
    File "<frozen codecs>", line 500, in read
  TypeError: can't concat str to bytes
  
  >>> Markdown().convertFile(BytesIO(b'input'), output=StringIO())
    File "markdown/core.py", line 446, in convertFile
      output_file.write(html)
    File "<frozen codecs>", line 378, in write
  TypeError: string argument expected, got 'bytes'
  ```

* `sys.stdin.read()` is always a string, so remove the Python2-specific check.
    * Correct the patch in the test which incorrectly sets `sys.stdin` to a bytes buffer which it never is.

* `sys.stdout.buffer` always exists, so remove the Python2-specific fallback.
    * Correct the patch in the test which incorrectly sets `sys.stdout` to a bytes buffer which it never is. And fix a matching mistake because this necessitated `stdout.read().decode()`.

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -23,7 +23,7 @@ import codecs
 import sys
 import logging
 import importlib
-from typing import TYPE_CHECKING, Any, Callable, ClassVar, Mapping, Sequence, TextIO
+from typing import TYPE_CHECKING, Any, BinaryIO, Callable, ClassVar, Mapping, Sequence
 from . import util
 from .preprocessors import build_preprocessors
 from .blockprocessors import build_block_parser
@@ -387,8 +387,8 @@ class Markdown:
 
     def convertFile(
         self,
-        input: str | TextIO | None = None,
-        output: str | TextIO | None = None,
+        input: str | BinaryIO | None = None,
+        output: str | BinaryIO | None = None,
         encoding: str | None = None,
     ) -> Markdown:
         """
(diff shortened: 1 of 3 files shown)


markdown/core.py
```diff
@@ -448,12 +446,8 @@ def convertFile(
         else:
             # Encode manually and write bytes to stdout.
             html = html.encode(encoding, "xmlcharrefreplace")
-            try:
-                # Write bytes directly to buffer (Python 3).
-                sys.stdout.buffer.write(html)
-            except AttributeError:  # pragma: no cover
-                # Probably Python 2, which works with bytes by default.
-                sys.stdout.write(html)
+            # Write bytes directly to buffer (Python 3).
```
waylan:
```suggestion
            
```

waylan (commented):
Thanks for this (along with the explanation. I've been meaning to update this to remove the old legacy code and never remembered when I had the time/motivation.

markdown/core.py
oprypin: Done

waylan (approved):

PR #1402, opened 2023-11-02
Enable checking against mypy with minimal set of fixes

oprypin:
4 significant checks are disabled:
* `strict_optional` - check that an object might be `None` on access
* `warn_no_return` - requiring a `return`
* `assignment` - any assignment of a variable of a wrong type
* `var-annotated` - a type annotation is missing

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -175,7 +175,7 @@ class ListIndentProcessor(BlockProcessor):
         return block.startswith(' '*self.tab_length) and \
             not self.parser.state.isstate('detabbed') and \
             (parent.tag in self.ITEM_TYPES or
-                (len(parent) and parent[-1] is not None and
+                (len(parent) > 0 and parent[-1] is not None and
                     (parent[-1].tag in self.LIST_TYPES)))
 
     def run(self, parent: etree.Element, blocks: list[str]) -> None:
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -453,8 +453,8 @@ class Markdown:
                 # Don't close here. User may want to write more.
         else:
             # Encode manually and write bytes to stdout.
-            html_bytes = html.encode(encoding, "xmlcharrefreplace")
-            sys.stdout.buffer.write(html_bytes)
+            html = html.encode(encoding, "xmlcharrefreplace")
+            sys.stdout.buffer.write(html)  # type: ignore
 
         return self
 
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -85,7 +85,7 @@ class Markdown:
     callable which accepts an [`Element`][xml.etree.ElementTree.Element] and returns a `str`.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         """
         Creates a new Markdown instance.
 
@@ -445,8 +445,8 @@ class Markdown:
                 # Don't close here. User may want to write more.
         else:
             # Encode manually and write bytes to stdout.
-            html = html.encode(encoding, "xmlcharrefreplace")
-            sys.stdout.buffer.write(html)
+            html_bytes = html.encode(encoding, "xmlcharrefreplace")
+            sys.stdout.buffer.write(html_bytes)
 
         return self
 
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -85,7 +85,15 @@ class Markdown:
     callable which accepts an [`Element`][xml.etree.ElementTree.Element] and returns a `str`.
     """
 
-    def __init__(self, **kwargs) -> None:
+    tab_length: int
+    ESCAPED_CHARS: list[str]
+    block_level_elements: list[str]
+    registeredExtensions: list[Extension]
+    stripTopLevelTags: bool
+    references: dict[str, tuple[str, str]]
+    htmlStash: util.HtmlStash
+
+    def __init__(self, **kwargs):
         """
         Creates a new Markdown instance.
 
@@ -106,23 +114,23 @@ class Markdown:
 
         """
 
-        self.tab_length: int = kwargs.get('tab_length', 4)
+        self.tab_length = kwargs.get('tab_length', 4)
 
-        self.ESCAPED_CHARS: list[str] = [
+        self.ESCAPED_CHARS = [
             '\\', '`', '*', '_', '{', '}', '[', ']', '(', ')', '>', '#', '+', '-', '.', '!'
         ]
         """ List of characters which get the backslash escape treatment. """
 
-        self.block_level_elements: list[str] = BLOCK_LEVEL_ELEMENTS.copy()
+        self.block_level_elements = BLOCK_LEVEL_ELEMENTS.copy()
 
-        self.registeredExtensions: list[Extension] = []
+        self.registeredExtensions = []
         self.docType = ""  # TODO: Maybe delete this. It does not appear to be used anymore.
-        self.stripTopLevelTags: bool = True
+        self.stripTopLevelTags = True
 
         self.build_parser()
 
-        self.references: dict[str, tuple[str, str]] = {}
-        self.htmlStash: util.HtmlStash = util.HtmlStash()
+        self.references = {}
+        self.htmlStash = util.HtmlStash()
         self.registerExtensions(extensions=kwargs.get('extensions', []),
                                 configs=kwargs.get('extension_configs', {}))
         self.set_output_format(kwargs.get('output_format', 'xhtml'))
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -56,7 +56,7 @@ class HTMLExtractorExtra(HTMLExtractor):
         self.block_tags = set(self.block_level_tags) - (self.span_tags | self.raw_tags | self.empty_tags)
         self.span_and_blocks_tags = self.block_tags | self.span_tags
 
-    def reset(self):
+    def reset(self) -> None:
         """Reset this instance.  Loses all unprocessed data."""
         self.mdstack: list[str] = []  # When markdown=1, stack contains a list of tags
         self.treebuilder = etree.TreeBuilder()
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -40,6 +40,10 @@ class HTMLExtractorExtra(HTMLExtractor):
     Markdown.
     """
 
+    mdstack: list[str] = []  # When markdown=1, stack contains a list of tags
+    treebuilder: etree.TreeBuilder
+    mdstate: list[Literal['block', 'span', 'off', None]]
+
     def __init__(self, md: Markdown, *args, **kwargs):
         # All block-level tags.
         self.block_level_tags = set(md.block_level_elements.copy())
@@ -56,11 +60,11 @@ class HTMLExtractorExtra(HTMLExtractor):
         self.block_tags = set(self.block_level_tags) - (self.span_tags | self.raw_tags | self.empty_tags)
         self.span_and_blocks_tags = self.block_tags | self.span_tags
 
-    def reset(self) -> None:
+    def reset(self):
         """Reset this instance.  Loses all unprocessed data."""
-        self.mdstack: list[str] = []  # When markdown=1, stack contains a list of tags
+        self.mdstack = []  # When markdown=1, stack contains a list of tags
         self.treebuilder = etree.TreeBuilder()
-        self.mdstate: list[Literal['block', 'span', 'off', None]] = []
+        self.mdstate = []
         super().reset()
 
     def close(self):
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -179,7 +179,10 @@ class SubstituteTextPattern(HtmlInlineProcessor):
 
 class SmartyExtension(Extension):
     """ Add Smarty to Markdown. """
-    def __init__(self, **kwargs) -> None:
+
+    substitutions: dict[str, str]
+
+    def __init__(self, **kwargs):
         self.config = {
             'smart_quotes': [True, 'Educate quotes'],
             'smart_angled_quotes': [False, 'Educate angled quotes'],
@@ -189,7 +192,7 @@ class SmartyExtension(Extension):
         }
         """ Default configuration options. """
         super().__init__(**kwargs)
-        self.substitutions: dict[str, str] = dict(substitutions)
+        self.substitutions = dict(substitutions)
         self.substitutions.update(self.getConfig('substitutions', default={}))
 
     def _addPatterns(
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -179,7 +179,7 @@ class SubstituteTextPattern(HtmlInlineProcessor):
 
 class SmartyExtension(Extension):
     """ Add Smarty to Markdown. """
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
         self.config = {
             'smart_quotes': [True, 'Educate quotes'],
             'smart_angled_quotes': [False, 'Educate angled quotes'],
@@ -199,9 +199,8 @@ class SmartyExtension(Extension):
         serie: str,
         priority: int,
     ):
-        for ind, pattern in enumerate(patterns):
-            pattern += (md,)
-            pattern = SubstituteTextPattern(*pattern)
+        for ind, pattern_args in enumerate(patterns):
+            pattern = SubstituteTextPattern(*pattern_args, md)
             name = 'smarty-%s-%d' % (serie, ind)
             self.inlinePatterns.register(pattern, name, priority-ind)
 
@@ -253,7 +252,7 @@ class SmartyExtension(Extension):
         )
         self._addPatterns(md, patterns, 'quotes', 30)
 
-    def extendMarkdown(self, md):
+    def extendMarkdown(self, md: Markdown):
         configs = self.getConfigs()
         self.inlinePatterns: Registry[inlinepatterns.InlineProcessor] = Registry()
         if configs['smart_ellipses']:
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -47,10 +47,11 @@ import xml.etree.ElementTree as etree
 try:  # pragma: no cover
     from html import entities
 except ImportError:  # pragma: no cover
-    import htmlentitydefs as entities
+    import htmlentitydefs as entities  # type: ignore
 
 if TYPE_CHECKING:  # pragma: no cover
     from markdown import Markdown
+    from . import treeprocessors
 
 
 def build_inlinepatterns(md: Markdown, **kwargs: Any) -> util.Registry[InlineProcessor]:
@@ -191,7 +192,7 @@ class EmStrongItem(NamedTuple):
 # -----------------------------------------------------------------------------
 
 
-class Pattern:  # pragma: no cover
+class _BasePattern:
     """
     Base class that inline patterns subclass.
 
@@ -241,6 +242,32 @@ class Pattern:  # pragma: no cover
         """ Return a compiled regular expression. """
         return self.compiled_re
 
+    def type(self) -> str:
+        """ Return class name, to define pattern type """
+        return self.__class__.__name__
+
+    def unescape(self, text: str) -> str:
+        """ Return unescaped text given text with an inline placeholder. """
+        assert self.md is not None
+        try:
+            inlineprocessor: treeprocessors.InlineProcessor = self.md.treeprocessors['inline']
+            stash = inlineprocessor.stashed_nodes
+        except KeyError:  # pragma: no cover
+            return text
+
+        def get_stash(m: re.Match[str]) -> str:
+            id = m.group(1)
+            if id in stash:
+                value = stash.get(id)
+                if isinstance(value, str):
+                    return value
+                else:
+                    # An `etree` Element - return text content only
+                    return ''.join(value.itertext())
+        return util.INLINE_PLACEHOLDER_RE.sub(get_stash, text)
+
+
+class LegacyPattern(_BasePattern):
     def handleMatch(self, m: re.Match[str]) -> etree.Element | str:
         """Return a ElementTree element from the given match.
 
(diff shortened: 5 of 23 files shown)


waylan:
I'm sorry, this is is still way to much rewriting of code. I will not under any circumstances change the way I write code in a dynamically typed language to suite a type checker. The checker should only be verifying that existing annotations are correct.  

The only edits which are acceptable are edits to the annotations. If you can't make that work, then mypy is not the right tool.

I see three ways forward:

1. Configure mypy to do what I want (seems unlikely).
2. Find a different tool which does what I want.
3. Do nothing.

markdown/treeprocessors.py
```diff
@@ -218,7 +218,7 @@ def linkText(text: str | None) -> None:
                         text = data[strartIndex:index]
                         linkText(text)
 
-                    if not isString(node):  # it's Element
+                    if not isinstance(node, str):  # it's Element
```
waylan: This changes the functionality. Notice that the [isString](https://github.com/Python-Markdown/markdown/blob/3c6cebff6202f0f4ab088a3efa71b0b70eca2d0b/markdown/treeprocessors.py#L48) function only returns `True` if the object is an instance of `str` but not `AtomicString`. Because `AtomicString` is a subclass of `str`, your change has changed the intended behavior. Apparently we don't have a proper test case for this as no tests failed.

I by chance happened to see this and immediately recognized the issue. How many other changes have I missed? Please stop trying to rewrite the code to satisfy a type checker.

oprypin: I am aware of that, but if that situation actually matters then the current code is bugged. If it's `AtomicString` then `child.tail` will crash because `AtomicString` has no such attribute. Looking for such bugs at the moment.

waylan: Yes it actually matters. The entire purpose of `AtomicString` is to prevent further processing of a string of text. If you ignore that, then it no longer serves its purpose.

> child.tail will crash because AtomicString has no such attribute.

Well, neither does `str` so I don't see how your proposed change addresses that.

Curiously, the coverage report indicates that these lines of code have full test coverage. Maybe what we are missing is a case which should result in the if statement being `False`.

oprypin: Could you check the logic again, I think your last comment is incorrect.

`if not isString(node):  # it's Element`

`isString` checks that it's `str`.
`not isString` checks that it's `AtomicString` or `Element`.
But the comment proclaims `# it's Element`, and then the code proceeds to actually assume that wrongly.

oprypin: * Opened #1406 to fix this.

In order to make this small and simple improvement to the code I had to come up with totally contrived test cases  just to follow this-

> However, the fix should always be framed as a fix to the now known bug, not a change to the code to match the type annotations. In other words, I would expect a test case which fails before the fix is applied and succeeds after regardless of any type annotations.

waylan: Thank you for doing that. It was exactly what I was looking for. Much appreciated.

markdown/inlinepatterns.py
```diff
@@ -274,6 +261,27 @@ def get_stash(m):
         return util.INLINE_PLACEHOLDER_RE.sub(get_stash, text)
 
 
+class LegacyPattern(_BasePattern):
```
waylan, line 264: I have very serious objections to the changes to this file. We have API changes. Under no circumstances am I interested in making API changes for type checking. If that means it's not possible to move forward with mypy, then we don't move forward with mypy.

To be clear, this specific line of code is not any more or less relevant than any other. However, the changes are rather sweeping and it was not possible to easily highlight and comment on all relevant lines. I'm speaking about most of the changes to this file in general.

markdown/blockprocessors.py
```diff
@@ -175,7 +175,7 @@ def test(self, parent: etree.Element, block: str) -> bool:
         return block.startswith(' '*self.tab_length) and \
             not self.parser.state.isstate('detabbed') and \
             (parent.tag in self.ITEM_TYPES or
-                (len(parent) and parent[-1] is not None and
+                (len(parent) > 0 and parent[-1] is not None and
```
waylan, line 178: This change should be completely unnecessary. The code works just fine and any tool which prevents me from writing code that way is unwanted.

markdown/core.py
```diff
@@ -85,7 +85,7 @@ class Markdown:
     callable which accepts an [`Element`][xml.etree.ElementTree.Element] and returns a `str`.
     """
 
-    def __init__(self, **kwargs):
+    def __init__(self, **kwargs) -> None:
```
waylan: This addition adds no value and I will not tolerate being forced to add it.

markdown/core.py
```diff
@@ -445,8 +445,8 @@ def convertFile(
                 # Don't close here. User may want to write more.
         else:
             # Encode manually and write bytes to stdout.
-            html = html.encode(encoding, "xmlcharrefreplace")
-            sys.stdout.buffer.write(html)
+            html_bytes = html.encode(encoding, "xmlcharrefreplace")
+            sys.stdout.buffer.write(html_bytes)
```
waylan: As I have stated repeatedly, I embrace the fact that Python is dynamically typed and allows reuse of a variable. Any tool which disallows this is unacceptable.

markdown/extensions/md_in_html.py
```diff
@@ -56,7 +56,7 @@ def __init__(self, md: Markdown, *args, **kwargs):
         self.block_tags = set(self.block_level_tags) - (self.span_tags | self.raw_tags | self.empty_tags)
         self.span_and_blocks_tags = self.block_tags | self.span_tags
 
-    def reset(self):
+    def reset(self) -> None:
```
waylan: and again. And that goes for all remaining instances which I am not going to comment on individually.

markdown/extensions/smarty.py
```diff
@@ -199,9 +199,8 @@ def _addPatterns(
         serie: str,
         priority: int,
     ):
-        for ind, pattern in enumerate(patterns):
-            pattern += (md,)
-            pattern = SubstituteTextPattern(*pattern)
+        for ind, pattern_args in enumerate(patterns):
+            pattern = SubstituteTextPattern(*pattern_args, md)
```
waylan, line 206: Again the dynamic nature of Python is being taken away here.

markdown/postprocessors.py
```diff
@@ -75,7 +75,8 @@ def run(self, text: str) -> str:
         """ Iterate over html stash and restore html. """
         replacements = OrderedDict()
         for i in range(self.md.htmlStash.html_counter):
-            html = self.stash_to_string(self.md.htmlStash.rawHtmlBlocks[i])
+            raw: str = self.md.htmlStash.rawHtmlBlocks[i]
+            html = self.stash_to_string(raw)
```
waylan: Wait, so mypy doesn't allow his pattern (`funct1(funct2(value))`)? That's ridiculous.

markdown/treeprocessors.py
```diff
@@ -271,7 +271,12 @@ def __applyPattern(
             String with placeholders instead of `ElementTree` elements.
 
         """
-        new_style = isinstance(pattern, inlinepatterns.InlineProcessor)
+        if isinstance(pattern, inlinepatterns.InlineProcessor):
+            new_style = True
+            new_pattern = pattern
+        else:  # pragma: no cover
+            new_style = False
+            legacy_pattern = pattern
```
waylan: Again this (and all the associated changes in this file that follow) is removing the dynamic nature of python.

waylan (commented):
I just went through and noted (nearly) every objection I have to these changes.

waylan:
So, it seems that if you can configure mypy to not require any of the code changes included here, then it would be something we can use.

However, I will note that simply commenting lines of code may not be the best solution in some cases. For example. there is little difference in `def __init__(self) -> None:` and `def __init__(self): # ignore`. I would expect to be able to do `def __init__(self):`. And the same goes for every time we use Python's dynamic nature. There shouldn't need to be a comment on every such line. A global config option would be preferable. If mypy can't do that, then its not the right tool for the job.

markdown/blockprocessors.py
oprypin: The code can **actually** sometimes return an instance of `int` rather than `bool`. If we have to, I'll follow the usual over-the-top process of proving to you that this is a bug and writing a test.

markdown/core.py
oprypin: Adding any annotation indicates "please type-check the body of this function". The cool thing about it is that if you add a new function and not write any type annotations on it, mypy will not complain whatsoever. It is a **good** thing to have this tool. So this un-nuanced take on this is unproductive.

Here I'm adding this `-> None` because this function contains some type annotations that are being wasted otherwise. To not have problems with this, I'll have to move the annotations to class level. Let me know if that works better for your preference.

markdown/core.py
oprypin: OK let's use an ignore comment instead

Yeah it's totally a limitation of mypy

markdown/extensions/md_in_html.py
oprypin: Ditto 

> Adding any annotation indicates "please type-check the body of this function".

Changed for now

facelessuser:
Yes, it seems that the projects goals and this PR's goals are a bit at odds. It sounds like Python Markdown only cares about checking that types are defined, but type checking itself is not desired.

I will add one point though. There seems to be some disdain for explicitly defining a function has no return, but if that is disabled, it seems you have no actual way of checking if a return is required and missed? Maybe I'm wrong about this?

markdown/extensions/smarty.py
oprypin: This is indeed a limitation of mypy.
That being said, this change is just a mordernization that could've been done anyway.

markdown/inlinepatterns.py
oprypin: There is no actual change to the API, all usages can remain the same.

But this is still indeed the worst change here. I spent a lot of time thinking of this but it's the best I came up with. We could also solicit ideas regarding this.

I'll comment further on this soon.

markdown/postprocessors.py
oprypin: It's not that it doesn't allow nested calls, it's that there is another problem:

> Argument 1 to `stash_to_string` of `RawHtmlPostprocessor` has incompatible type `str | Element`; expected `str`

`self.md.htmlStash.rawHtmlBlocks[i]` is type-annotated as `str | Element`
-which is correct because `md_in_html` can indeed put an `Element` there:
https://github.com/Python-Markdown/markdown/blob/a63e6f3416f2e4fc7c2d5edb2768ceed9218df20/markdown/extensions/md_in_html.py#L166

However, `stash_to_string` is guarded by a `str` type annotation because indeed for an `Element` it will just produce garbage such as `"<Element 'a' at 0x7f55e83a8ea0>"`, and mypy rightly warns about it.

So once again there is a potential bug that mypy warns us about.
Upon looking into this though, I concluded that *maybe* there is no actual bug because `RawHtmlPostprocessor` consistently only puts strings into `htmlStash`.

For now turned it into an ignore comment.

markdown/treeprocessors.py
oprypin: OK let's just ignore-comment everything instead, I don't know

oprypin: A cool thing to do here would be to refactor this method into 2 methods. Despite some code duplication, it's actually less error-prone going forward, and even provides a clearer deprecation for old-style patterns. (Also only one `# pragma: no cover` comment now, rather than per-case ones 😊

This is demonstrated here: https://github.com/oprypin/python-markdown/commit/1c047f02bc86edbdd5a0eee69aaff37730a080ed?diff=split

markdown/inlinepatterns.py
oprypin: Let me explain what the problem here is.

This is the general explanation:
https://stackoverflow.com/questions/6034662/python-method-overriding-does-signature-matter

For the particular situation at hand:
`class InlineProcessor(Pattern):` violates the Liskov Substitution Principle.

And let's just quote directly from mypy's output:

```
Signature of "handleMatch" incompatible with supertype "Pattern"  [override]
     Superclass:
         def handleMatch(self, m: Match[str]) -> Element | str
     Subclass:
         def handleMatch(self, m: Match[str], data: str) -> tuple[Element | str | None, int | None, int | None]
```

The particular violation happens regardless of mypy, it's an actual run-time problem if users expect it to work like is widely accepted:

If you write:
```python
if isinstance(foo, Pattern):
    foo.handleMatch(single_argument)
```
it must work, because every instance of `Pattern` **must** have a single-argument `handleMatch` method, because it's written so right there!

An instance of `InlineProcessor` is an instance of `Pattern`, and every instance of it violates this principle.

More specific code example:

```pycon
>>> import re
>>> from markdown.inlinepatterns import EscapeInlineProcessor, Pattern

>>> foo = Pattern('a')
>>> foo.handleMatch(re.search('a', 'a'))  # OK

>>> foo = EscapeInlineProcessor('a')
>>> isinstance(foo, Pattern)
True
>>> foo.handleMatch(re.search('a', 'a'))
TypeError: EscapeInlineProcessor.handleMatch() missing 1 required positional argument: 'data'
```



The correct hierarchy would be this:

```mermaid
  graph TD;
      BasePattern["BasePattern<br>defines helper methods only"]-->Pattern["Pattern<br>abstract def handleMatch(m):"];
      BasePattern-->InlineProcessor["InlineProcessor<br>abstract def handleMatch(m, data):"];
      Pattern-->SimpleTextPattern["SimpleTextPattern<br>def handleMatch(m):"]
      Pattern-->SimpleTagPattern["SimpleTagPattern<br>def handleMatch(m):"]
      InlineProcessor-->EscapeInlineProcessor["EscapeInlineProcessor<br>def handleMatch(m, data):"]
      InlineProcessor-->AsteriskProcessor["AsteriskProcessor<br>def handleMatch(m, data):"]
```

But the current hierarchy is:

```mermaid
  graph TD;
      Pattern["Pattern<br>defines helper methods<br>abstract def handleMatch(m):"];
      Pattern-->InlineProcessor["InlineProcessor<br>abstract def handleMatch(m, data):"];
      Pattern-->SimpleTextPattern["SimpleTextPattern<br>def handleMatch(m):"]
      Pattern-->SimpleTagPattern["SimpleTagPattern<br>def handleMatch(m):"]
      InlineProcessor-->EscapeInlineProcessor["EscapeInlineProcessor<br>def handleMatch(m, data):"]
      InlineProcessor-->AsteriskProcessor["AsteriskProcessor<br>def handleMatch(m, data):"]
```

As such, the entire hierarchy of `InlineProcessor` is said to have improper overrides.

What my change accomplishes is to construct the correct type hierarchy to mypy and to all typechecking users going forward.
**However at run time I am keeping everything just like it was. `LegacyPattern` is just a new alias to `Pattern`**

oprypin: See a code search of occurrences in the wild: currently every subclass of `InlineProcessor` needs to add an ignore comment for this override.

https://github.com/search?q=%2FInlineProcessor%5C%29%2F+%2FhandleMatch.%2Bignore%2F+NOT+path%3Azerver&type=code

Of course at the moment that's not this repository's fault (as it doesn't have `py.typed`) but just because the type stubs are written in a way that directly mirrors it:
https://github.com/python/typeshed/blob/9ff393558a6f4fedc086bc9dc4386dbe529a3331/stubs/Markdown/markdown/inlinepatterns.pyi#L48

oprypin: There's also this issue report (which I found through the above code search because it's directly inspired by this case). Which if accepted would resolve this

* https://github.com/python/mypy/issues/10197

oprypin: An equivalent change to the external stubs would be this:
https://github.com/oprypin/typeshed/commit/43593a451a181726482dcfcce681e3f1d8b1bf08
Maybe this more condensed view is clearer to understand.

markdown/core.py
waylan: But when does an `__init__`  method return anything other than `None`? That's my point. Adding this annotation adds no value in that it tells me nothing I didn't already know. I should not be  required to add this annotation every time I define and `__init__` function.

You seem to object based on the fact that eliminating the annotation causes the tool to not run other checks on the line. That is a problem with the tool. If the tool can't do that, then fix the tool or use a different tool. Don't force me to waste time and effort adding otherwise pointless annotations.

markdown/core.py
waylan: Then lets find a tool which doesn't have this limitation. Quit wasting my time with this nonsense.

markdown/extensions/smarty.py
waylan: To be clear. if the code had been written that way originally when it was submitted in a PR, I would have accepted It. But I would be willing to accept the existing code also (as I did). I don't have an objection to either way, but I do have an objection to a tool forcing one way on me and not allowing the other. That is not a tool I am interested in using.

markdown/core.py
oprypin: Adding `-> None` annotations to `__init__` **is** useless, I totally agree on that. However you ignore the point that I am actually arguing for. **You** want the tool to not do anything unless you call for it. As such, not adding any annotations accomplishes that. But when you **do** want to go ahead and tell the tool "please type-check this function body", this is the designated way to do that. You are not forced, and I demonstrated that with the current state of this pull request.

Separately on your last point: the tool **does not** require this annotation, you can simply configure it to check all function bodies. If you want it to check only **some** function bodies, this is how to indicate it.

markdown/postprocessors.py
waylan: > Upon looking into this though, I concluded that _maybe_ there is no actual bug because `RawHtmlPostprocessor` consistently only puts strings into `htmlStash`.

Exactly. Again, this is a waste of time and not helpful at all. Unless maybe this is telling us that we need to better annotate the stash to indicate that it only accepts strings. Would doing that resolve the issue? If so, then that would be how to address this, not this.

oprypin: I think you did not read my comment very attentively because I explicitly say that there is in fact code that puts Elements into the stash.

markdown/treeprocessors.py
waylan: > OK let's just ignore-comment everything instead, I don't know

Actually, as per my [other comment](https://github.com/Python-Markdown/markdown/pull/1402#issuecomment-1792588190), not don't just ignore-comment everything.

> A cool thing to do here would be to refactor this method into 2 methods.

Perhaps, but that should be addresses separately.

waylan:
> I will add one point though. There seems to be some disdain for explicitly defining a function has no return, but if that is disabled, it seems you have no actual way of checking if a return is required and missed? Maybe I'm wrong about this?

I'm okay with that. The only thing that should be checked is that existing annotations are correct. Anything that is not annotated should be ignored. Seems like a simple concept. I don't understand why that is so hard... except that perhaps it is because the goal of the type checker is to enforce static typing, which is something I have repeatedly said I am not interested in.

markdown/core.py
```diff
@@ -85,6 +85,14 @@ class Markdown:
...
     """
 
+    tab_length: int
+    ESCAPED_CHARS: list[str]
+    block_level_elements: list[str]
+    registeredExtensions: list[Extension]
+    stripTopLevelTags: bool
+    references: dict[str, tuple[str, str]]
+    htmlStash: util.HtmlStash
+
```
waylan, line 95: These are instance attributes, not class attributes. Will you please stop changing the code to fit a tool which I have repeatedly stated does not fit our needs.

I am closing this issue as you are clearly either not understanding my direction or intentionally ignoring it.

oprypin:
> The only thing that should be checked is that existing annotations are correct.

In which way, just purely conceptually, do you think this is supposed to be accomplished? What is "correct" in isolation?

The only known way to do this is to check code that uses annotated members. If something doesn't match up then it shows an error, and you need to check if it's the code that is wrong or the annotation that is wrong.

markdown/core.py
oprypin: This is the correct way to annotate instance attributes according to all tools and documentation

facelessuser:
> I'm okay with that. The only thing that should be checked is that existing annotations are correct. Anything that is not annotated should be ignored. Seems like a simple concept. I don't understand why that is so hard... except that perhaps it is because the goal of the type checker is to enforce static typing, which is something I have repeatedly said I am not interested in.

That's fair, and it is possible I've misinterpreted some interaction. So yeah, I'm good with that.

waylan:
> The only known way to do this is to check code that uses annotated members. If something doesn't match up then it shows an error, and you need to check if it's the code that is wrong or the annotation that is wrong.

Yes, when its annotated that is reasonable. And when it is not annotated, it should be ignored.

In other words, suppose we have to functions, one of which is annotated and one which is not. The tool would see a line of code which calls the annotated function, It would see the annotation and confirm that the return value matches up. However, when it sees the unannotated function is would not confirm anything about the return value as there is nothing to confirm. That is what I expect to happen.

In the first case, the assumption would then be that perhaps the annotation is wrong. However, we could check for a potential bug in the code and fix that instead if need be. In the second case, we are alerted to nothing as it has been for many years.

And no, I don't want to have to add an ignore comment to every one of those lines. The tool should just ignore them either by default or via a global setting.

oprypin:
I think mypy in the current config matches what you describe - it validates only things that are annotated.

Which of the comment threads don't match this? There was an example where something clearly annotated as `str | Element` is being passed to something clearly annotated as `str`. How is that not a correct finding that either the annotation is wrong or the code can have a problem?

Although a big snag might be in the fact that the whole standard library is annotated, so usages (inside annotated functions) are checked against it as well.

And there *was* one consistently mentioned limitation of mypy - that it can't cope with reassigning a different-typed value to the same variable name, and it should be able to. I have replied with "limitation" to those threads. But it's a conceptually different problem. And if we do look at it separately, perhaps this pull request can be said to have only this as a blocker.

waylan:
> Which of the comment threads don't match this? There was an example where something clearly annotated as `str | Element` is being passed to something clearly annotated as `str`. How is that not a correct finding that either the annotation is wrong or the code can have a problem?

I've never suggested that that was the case. In fact, if the issue is that the annotation was wrong, I have made no objection. However, I have consistently stated that if the code is wrong, then each individual instance should be addressed as a separate bug, not included wholesale in this large PR. This PR (or one like it) should not be changing code. only annotations. And yet, you have repeatedly ignored that request.

> And there _was_ one consistently mentioned limitation of mypy - that it can't cope with reassigning a different-typed value to the same variable name, and it should be able to. I have replied with "limitation" to those threads. But it's a conceptually different problem. And if we do look at it separately, perhaps this pull request can be said to have only this as a blocker.

And a very big blocker is it, I have stated repeatedly that I will not change this lib to be statically typed, which mypy requires. In other words, I will not change the code to appease a tool which is designed to enforce something we are not enforcing. And it is unreasonable to expect us to ignore-comment every single dynamically typed object in our code line-by-line. If mypy had a global option to not enforce that, then it wouldn't be a tool which enforced static types. Therefore, it seems that if does not fit our needs.

PR #1406, opened 2023-11-02
Fix edge-case crash in InlineProcessor

oprypin:
If an inlineprocessor returns an AtomicString (even though that is pointless, a plain string is atomic in that context), there can be an exception in 2 separate places. The added test case was crashing before this change.

```
  File "tests/test_apis.py", line 722, in testInlineProcessorDoesntCrashWithAtomicString
    new = self.inlineprocessor.run(tree)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "markdown/treeprocessors.py", line 384, in run
    lst = self.__processPlaceholders(
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "markdown/treeprocessors.py", line 223, in __processPlaceholders
    if child.tail:
       ^^^^^^^^^^
AttributeError: 'AtomicString' object has no attribute 'tail'
```

```
  File "tests/test_apis.py", line 722, in testInlineProcessorDoesntCrashWithAtomicString
    new = self.inlineprocessor.run(tree)
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "markdown/treeprocessors.py", line 387, in run
    self.__handleInline(text), child
    ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "markdown/treeprocessors.py", line 136, in __handleInline
    data, matched, startIndex = self.__applyPattern(
                                ^^^^^^^^^^^^^^^^^^^^
  File "markdown/treeprocessors.py", line 310, in __applyPattern
    if not isinstance(node.text, util.AtomicString):
                      ^^^^^^^^^
AttributeError: 'AtomicString' object has no attribute 'text'
```

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Remove legacy import needed only in Python 2 (#1403)
 * Fix typo that left the attribute `AdmonitionProcessor.content_indent` unset
   (#1404)
+* Fix edge-case crash in `InlineProcessor` with `AtomicString` (#1406).
 * Improve and expand type annotations in the code base (#1401).
 
 ## [3.5.1] -- 2023-10-31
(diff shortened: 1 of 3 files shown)


waylan (approved):
Well, it looks like I was wrong about this. Turns out any string should not follow that code path.

Thanks for the which make sit clear.

PR #1410, opened 2023-11-06
Split changelog workflows to separate logic on when each gets run

waylan:


waylan:
Hmm, neither of the changelog workflows should be running in this PR as the requirements are not met for either (no changes were made in `markdown/` and no changes to the changelog). This used to work before the change in #1388. I thought that if we separated out the validation, then the previous behavior would return. Not sure what I'm missing here.

oprypin:
https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#onpushpull_requestpull_request_targetpathspaths-ignore

Probably there's a typo
`paths` `path`

waylan:
> Probably there's a typo
> `paths` `path`

That was it. Thank you. 😊 

Strangely, that was not changed in #1388. Now I'm not sure when that stopped working.

PR #1435, opened 2024-01-22
change: set LICENSE to from BSD to BSD-3-Clause

aricma:
Hi, 
I am working on a list of dependencies and their attached licenses for an other project. I realised that Python-Markdown/markdown did not have a specific license attached to it.
PyPi would only give me the "BSD License" and the GitHub API did not have an entry. GitHub does not even render the license information itself when you open the `Python-Markdown/markdown/LICENSE` file.

Screenshot 2024-01-22 at 21 41 59

It looks to me like the ./LICENSE file does not follow any of the GitHub templates for open source licenses.

There are a lot of packages on PyPi with only "BSD License"s and many use the "BDS-3-Clause" on GitHub.

I do not know which "BSD License" this project needs or which license Python-Markdown would prefer. I did go ahead and prepared a PR with the "BSD-3-Clause".

I hope this helps. I would love to resolve this and have a license information fetch-able from GitHub for Python-Markdown/markdown.

Changes:
- change: set LICENSE to from BSD to BSD-3-Clause via a GitHub template


waylan:
We already have a license file at [LICENSE.md](https://github.com/Python-Markdown/markdown/blob/master/LICENSE.md), not `LICENSE`. We properly include it in the PyPI distribution. Now you have added a second license file.

I don't recall why it was given the `.md` file extension. Does GitHub and/or PyPI require a specific filename to recognize it? If so, can you point to that documentation? I suppose we can change the filename if need be. But let's not have two different files.

waylan:
Looks like the `.md` file extension was added way back in a6fd7b5, which was part of a reorganization of the documentation. However, the documentation now uses a completely different engine, so the needs then no longer apply. Therefore, there is no need to retain the file extension.

waylan:
> It looks to me like the ./LICENSE file does not follow any of the GitHub templates for open source licenses.

What is the issue? I see 3 possible concerns:

1. We have 3 lines of `Copyright <YEAR> <COPYRIGHT HOLDER>`? If this is the issue, then that is GitHub's problem. We need to list all 3. However, I should note that the first one (`2007, 2008`) should be updated to present day (`2007 - 2024` or `2007 - present`??).

2. There is the paragraph `All rights reserved.` which appears to not be in any templates. That has been there since a license was first added to the project back on 2008 (see 3a78823). I'm reluctant to remove this as it could change the license which all of the present code has been provided under. Not sure if that clause is legally significant or not.

3. The 3 clauses use bullet points (`*`) rather than numbered list items (`1.`, `2.`, `3.`). Again this has been like that from the beginning. Although, I wouldn't expect the changing the formatting of the list to have any legally significant change.

Maybe the template for the BSD-3-Clause was different back in 2008?

aricma:
Hey @waylan,
thank you for responding to my PR 🙏.

Let me implement the changes you recommended and then we can have another look 👍.

1. I will remove the old LICENSE.md in favour of the new LICENSE. I would omit the `.md` extension since the document does not contain markdown syntax. In addition to your comment: The extension is no longer needed for your documentation.
2. I will change the date range back to `2007 - present`.
3. I will add back the line `All rights reserved.`

I agree that these changes are rather small and that GitHub might need to up their game of recognising licenses from plain text.
However the old template does not say in any way which license the document is talking about. To help people understand the license I would recommend going for the new template which explicitly states the "BSD-3-clause".

The added benefit is that GitHub does recognise your license and developer fetching the repository information via http will get "spdx_id: BSD-3-Clause".

Here the api call: `curl https://api.github.com/repos/Python-Markdown/markdown`
And the current response:
```
{
...
"license": {
    "key": "other",
    "name": "Other",
    "spdx_id": "NOASSERTION",
    "url": null,
    "node_id": "MDc6TGljZW5zZTA="
  },
...
}
```

aricma:
@waylan I added the changes as proposed.

The files look almost the same. The wording in point 3 and the disclaimer at the end is slightly different, due to the new template. But there is no semantic difference.

I need to point out that Github might still not recognise the license. At least in my fork of your repo I had to fiddle a bit with the LICENSE file to make it work.

Since I am working on a branch the license information of the repo is not changed until the merge into master and even then  you might have to follow this stack overflow(https://stackoverflow.com/questions/66258037/github-license-not-recognised) and rename the file or create a new LICENSE, select the BSD-3-Clause template, and change the file content.

Changing the license is an owner job anyway, right?

LICENSE
```diff
@@ -0,0 +1,30 @@
+BSD 3-Clause License
+
+Copyright (c) 2007 - present, Python-Markdown
```
waylan: I am not comfortable with removing the other copyright notices here. Either they all need to be left in or we are not making any change.

```
Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)
Copyright 2004 Manfred Stienstra (the original version)
```

I suppose with express permission from the named individuals, we could remove them. But not without. Unfortunately, I have no way of contacting Manfred and Yuri hasn't responded to my attempts to contact him for some years now. We may be stuck with this license as-is.

aricma: I added the copyright information from the LICENSE.md file on master ✅.

Issue #1433, opened 2024-01-19
Strange and inconsistent parsing of lists with headers and multiple lines


Andre601:
A strange parsing behaviour can be observed with the markdown library when used on Lists containing headers and multiple lines.

## Problem

When a list entry contains a header and subsequent text using indents of 4 spaces, will the first entry render fine.
However, any subsequent entry will have their text after the header be rendered as code blocks, due to the behaviour of turning lines with 4 spaces indent into code blocks.

This behaviour strangely enough can only be observed if there is a gap in-between two list entries. If they are right after each other will the result be as expected (That being text rendering as normal paragraph).
Even stranger is the behaviour with any additional paragraphs. Should a list entry have more than one paragraph, meaning there is a gap between the first and second text after the header, will it render the second paragraph fine, will still having this odd behaviour on the first.

See my tests below for possible results.

There are fixes/workarounds for this.

The first one being to have no gaps in-between the list entries. This is the easiest in terms of keeping consistency, but may worsen readability of the raw content.

The second option is to start the first paragraph with 2 indents instead of 4. This may cause visual inconsistencies in the raw text if an entry has more than one paragraph, by having different indents, since the second paragraph needs to have 4 indents or else it won't be included as part of the list entry.

----

## Tests

These tests were all made using `python -m markdown test.txt` with the test.txt file containing the below displayed markdown content.

### Test 1

Base-line test showing the issue.

Input:
```markdown
- ### List 1
    Entry 1.1

- ### List 2
    Entry 2.1

- ### List 3
    Entry 3.1
```
Output:
```html
<ul>
<li>
<h3>List 1</h3>
<p>Entry 1.1</p>
</li>
<li>
<h3>List 2</h3>
<pre><code>Entry 2.1
</code></pre>
</li>
<li>
<h3>List 3</h3>
<pre><code>Entry 3.1
</code></pre>
</li>
</ul>
```

### Test 2

Test with multiple paragraphs

Input:
```markdown
- ### List 1
    Entry 1.1
    
    Entry 1.2

- ### List 2
    Entry 2.1
    
    Entry 2.2

- ### List 3
    Entry 3.1
    
    Entry 3.2
```
Output:
```html
<ul>
<li>
<h3>List 1</h3>
<p>Entry 1.1</p>
<p>Entry 1.2</p>
</li>
<li>
<h3>List 2</h3>
<pre><code>Entry 2.1
</code></pre>
<p>Entry 2.2</p>
</li>
<li>
[...]

PR #1436, opened 2024-01-22
Heading in list

waylan:
Here is a start at a fix for #1433. So far I have tests, one of which is failing.

### DIFF
--- a/.spell-dict
+++ b/.spell-dict
@@ -32,6 +32,7 @@ customizable
 dedent
 deliminators
 deregister
+detabbed
 Dmitry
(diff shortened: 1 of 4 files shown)


waylan:
This is ready to go. Turns out the first item in a list was being parsed differently because the list parser does not look ahead and it doesn't know it is in a loose list until it gets to the second list item. The fix here just detects that its in the second (or later) item and adjusts the content `after` the heading (by detabbing it) so that it gets parsed correctly. See the commit message for 3fca127 for more details.

PR #1441, opened 2024-02-09
Refactor TOC sanitation

### LINKED ISSUE #660: Footnote ref number in TOC

waylan:
- All postprocessors are run on heading content (not just `RawHtmlPostprocessor`).
- Footnote references are stripped from heading content. Fixes #660.
- A more robust `striptags` is provided to convert headings to plain text. Unlike, markupsafe's implementation, HTML entities are not unescaped.
- Both the plain text `name` and rich `html` <ins>and unescaped raw `data-toc-label`</ins> are saved to `toc_tokens`, which means users can now access the full rich text content of the headings directly from the `toc_tokens`.
- `data-toc-label` is sanitized separate from heading content.
- ~A `html.unescape` call added to `slugify` and `slugify_unicode`, which ensures `slugify` operates on Unicode characters, rather than HTML entities. By including in the functions, users can override with their own slugify functions if they desire.~ <ins>A `html.unescape` call is made just prior to calling `slugify` so that `slugify` only operates on Unicode characters. Note that `html.unescape` is not run on the `name` or `html`.</ins>

Note that this first commit includes minimal changes to the tests to show very little change in behavior (mostly the new `html` attribute of the `toc_tokens` was added). A refactoring of the tests will be in a separate commit.


markdown/extensions/toc.py
```diff
@@ -63,41 +68,81 @@ def unique(id: str, ids: MutableSet[str]) -> str:
     return id
 
 
-def get_name(el: etree.Element) -> str:
```
oprypin, line 66: Removal breaks pymdownx.tabbed

* https://github.com/facelessuser/pymdown-extensions/blob/8f5283f71e8833f1ba0fa0bbe2680e1ad2c6cb19/pymdownx/tabbed.py#L377
* https://github.com/mkdocs/regressions/actions/runs/7848285213/job/21419074861

markdown/extensions/toc.py
oprypin, line 1: There are some regressions when the raw symbols `&` or `<` are used

https://github.com/mkdocs/regressions/actions/runs/7848285213/job/21419076709#step:7:832

https://github.com/mkdocs/regressions/actions/runs/7848285213/job/21419082259#step:7:123

oprypin, line 1: Seems it now sanitizes over mkdocstrings' use case of inserting  `<code class="doc-symbol doc-symbol-toc doc-symbol-function">`

oprypin (commented):
Seems like a good direction to move in :)

I did a cursory check whether the change affects performance on a large MkDocs site, and seems it doesn't!

waylan:
In my initial commit, I deleted a couple public functions which are no longer needed. I probably should have left them in and marked them as deprecated instead. Do we need to do that? Are others calling them?

Also, it occurs to me that these changes could result in various slugs being different, which would break existing links out in the wild. True, none of the slugs in our own tests are changed, but by fully rendering the HTML before generating a slug could result in additional/different content from any third-party extensions that make use of postprocessors.

markdown/extensions/toc.py
oprypin: Forgot link: https://github.com/mkdocs/regressions/actions/runs/7848285213/job/21419076447

markdown/extensions/toc.py
waylan: Good to know. We will need to restore and mark deprecated then. With a quick glance at that code, presumably the new functions should be usable there. But we need to provide a graceful migration path.

markdown/extensions/toc.py
waylan: Those appear to be related to slugifying HTML entities. Will need to take a closer look at that. It may be that the old code was handling this poorly and so this is an improvement.

waylan: It looks like `show_symbol_type_toc: true` is no longer working. Will need to investigate.

waylan: @oprypin where are the source for those? I'm trying to determine what the Markdown input was so I can replicate.

oprypin: https://github.com/search?q=repo%3Aikrima%2Fgamedevguide+%22Commands+for+toggling+debug%22&type=code

https://github.com/search?q=repo%3Aspaceship-prompt%2Fspaceship-prompt+%2FDo.%2Bspaceship.%2Bremove%2F&type=code

waylan: Okay, I found the first one

https://github.com/ikrima/gamedevguide/blob/464aaca2973931f50503de85db650561179b54ed/docs/ue4guide/performance-optimization/tolva-tool/tolva-perf-tool-helpers.md?plain=1#L247C1-L247C47

It turns out the behavior of Markdown is exactly the same before and after the change. 

### Before

```python
>>> md = markdown.Markdown(extensions=['toc'])
>>> md.convert('## Commands for toggling debug & perf markers:')
'<h2 id="commands-for-toggling-debug-perf-markers">Commands for toggling debug &amp; perf markers:</h2>'
>>> md.toc_tokens
[{'level': 2, 'id': 'commands-for-toggling-debug-perf-markers', 'name': 'Commands for toggling debug &amp; perf markers:', 'children': []}]
```

### After

```python
>>> md = markdown.Markdown(extensions=['toc'])
>>> md.convert('## Commands for toggling debug & perf markers:')
'<h2 id="commands-for-toggling-debug-perf-markers">Commands for toggling debug &amp; perf markers:</h2>'
>>> md.toc_tokens
[{'level': 2, 'id': 'commands-for-toggling-debug-perf-markers', 'name': 'Commands for toggling debug &amp; perf markers:', 'html': 'Commands for toggling debug &amp; perf markers:', 'children': []}]
```

The difference in the regression test is the addition of `amp` in the slug/id of the heading. I'm assuming a custom/third party slugify function is being used. The difference is that I moved the `html.unescape` call to be within the `slugify` method. And that means third party replacement methods would need to do the same. My thinking was that this could be customizable by the user via the method, but there is no graceful way to make that change without changing existing slugs (and breaking links). So I may need to revert that by moving the call out of slugify (to be called right before slugify).

waylan: > I'm assuming a custom/third party slugify function is being used.

Yep, I assumed correctly. See config here: https://github.com/ikrima/gamedevguide/blob/464aaca2973931f50503de85db650561179b54ed/mkdocs.yml#L102

@facelessuser he is using your slugify function. Do you have any thoughts about this?

oprypin: There's not much to think about. Slugify implementations always operated on text, it doesn't make any sense to suddenly start passing text with HTML entities to them

tests/test_extensions.py
oprypin, line 1: A user showed another edge case here https://github.com/mkdocs/mkdocs/issues/3357#issuecomment-1937863852

Would be good to add a unittest for this.

Behavior at master (bad):

```pycon
>>> import markdown
>>> md = markdown.Markdown(extensions=['toc'])
>>> md.convert('## <foo@example.org>')
'<h2 id="amp102amp111amp111amp64amp101amp120amp97amp109amp112amp108amp101amp46amp111amp114amp103"><a href="&#109;&#97;&#105;&#108;&#116;&#111;&#58;&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;">&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;</a></h2>'
>>> md.toc_tokens
[{'level': 2,
'id': 'amp102amp111amp111amp64amp101amp120amp97amp109amp112amp108amp101amp46amp111amp114amp103',
'name': '\x02amp\x03#102;\x02amp\x03#111;\x02amp\x03#111;\x02amp\x03#64;\x02amp\x03#101;\x02amp\x03#120;\x02amp\x03#97;\x02amp\x03#109;\x02amp\x03#112;\x02amp\x03#108;\x02amp\x03#101;\x02amp\x03#46;\x02amp\x03#111;\x02amp\x03#114;\x02amp\x03#103;',
'children': []}]
```

Behavior with this PR (good):

```pycon
>>> md = markdown.Markdown(extensions=['toc'])
>>> md.convert('## <foo@example.org>')
'<h2 id="fooexampleorg"><a href="&#109;&#97;&#105;&#108;&#116;&#111;&#58;&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;">&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;</a></h2>'
>>> md.toc_tokens
[{'level': 2,
'id': 'fooexampleorg',
'name': '&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;',
'html': '<a href="&#109;&#97;&#105;&#108;&#116;&#111;&#58;&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;">&#102;&#111;&#111;&#64;&#101;&#120;&#97;&#109;&#112;&#108;&#101;&#46;&#111;&#114;&#103;</a>',
'children': []}]
```

vedranmiletic:
FWIW, I installed this using

```
pip install git+https://github.com/Python-Markdown/markdown.git@refs/pull/1441/head
```

and tested it on [my group's website page that had issues](https://group.miletic.net/hr/nastava/materijali/web-arhitekture/), angle brackets are removed correctly and the e-mails show up nicely. Not sure I can help with actual code review, but from a user's perspective, this fixes the issue.

markdown/extensions/toc.py
waylan: This was backed out.

tests/test_extensions.py
waylan: A test was added for this. Thank you.

markdown/extensions/toc.py
waylan: @pawamoy if you have any input on this that would be appreciated.

I can never remember how mkdocstrings implements `show_symbol_type_toc: true` so I'm making notes this time.

- For each object/class/function/method, mkdocstrings-python uses a template to render Markdown as found under [python/src/mkdocstrings_handlers/python/templates/material/_base](https://github.com/mkdocstrings/python/tree/eaf9b8240069f7369f401fe048892043c8b173d3/src/mkdocstrings_handlers/python/templates/material/_base).
- Each template creates a `heading` object and assigns some text to a  `toc_label` attribute (see [module](https://github.com/mkdocstrings/python/blob/eaf9b8240069f7369f401fe048892043c8b173d3/src/mkdocstrings_handlers/python/templates/material/_base/module.html#L19-L25) as an example.
- mkdocstrings then uses that to build the Markdown heading and [assigns](https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/base.py#L310) the value of `toc_label` to the `data-toc-label` attribute of the heading.
- The TOC extension then parses that Markdown and uses the `data-toc-label` when building the toc_token, which MkDocs then uses to build the TOC of the page.

The problem is that the `toc_label`/`data-toc-label` contains raw HTML when `show_symbol_type_toc: true`. For example, for the `markdown` module, the value would be `<code class="doc-symbol doc-symbol-toc doc-symbol-module"></code>&nbsp;markdown`. Note that the `code` tag contains no content, but it is assigned a set of classes which will result in the provided CSS inserting the correct symbol in the TOC for a module.

As first glance, it would seem that this never would have worked. In fact, before the changes in this PR, we called `code_escape` on the content of `data-toc-label`. However, that was called before any placeholders were swapped for raw HTML. Presumably, by using raw HTML mkdocstrings was bypassing sanitation. I consider this to be a bug in the old implementation. In fact, the old `stashedHTML2text` function (called after `code_escape`) is also supposed to strip tags from the raw HTML. So, if things were working as intended, mkdocstrings would never have been able to implement their existing solution.

The new implementation in this PR actually implements the intended behavior and so mkdocstrings' hack is no longer possible. This puts me in the same conundrum as MkDocs with page titles. If MkDocs strips tags from all page titles, then mkdocstrings companion feature `show_symbol_type_heading` would stop working because the similar raw HTML would be stripped from headings. But I digress.

I see 3 possible ways forward:

1. Leave the PR as-is and mkdocsstrings will be broken until they find a different solution.
2. Change the intended behavior to allow raw HTML in `data-toc-label` .
3. Provide a mechanism to pass the raw value of `data-toc-label`  along with the sanitized version in the `toc_tokens`.

I find option 1 attractive because I am fixing a bug which was never supposed to exist. However, I understand why this would be undesirable to the mkdocstrigns devs. I wouldn't like it if I was in their shoes. Option 3 only adds more complexity to an already more complex token API and would require changes by both mkdocs and mkdocstrings, so not exactly a graceful change.

And that leaves option 2. I should note that I am assuming what the intended behavior was when this was originally developed. Unfortunately, sanitizing the main content of the heading and the content of `data-toc-label`  was done with the same code, so it is not clear why all of the various function calls were there. However, in some cases, raw HTML would not make it through, but in others it would. mkdocstrings just happened to find an edge case that worked for their needs. However, now that sanitizing the main heading content and sanitizing `data-toc-label` are separated, we can do whatever we want with the value of `data-toc-label`.

One complication I should note is that if we remove the calls to `strip_tags` and `escape_cdata` for the content of `data-toc-label`, then, while the `toc_token` will contain usable raw HTML, the rendered `toc` (using the `[TOC]` marker) will get escaped raw HTML when passed through Markdown's seralizer. This potential inconsistency is why the intended behavior is to assume that no raw HTML is permitted in `data-toc-label`. The general thinking of this code change is to fully render the content of a heading (all Markdown and raw HTML), then to extract the plain text and use that as the TOC entry. However, if we want to preserve the raw HTML for the `[TOC]` when passing through the serializer, then we need to use the unrendered content of `data-toc-label`. And that means that the code which builds the HTML toc (within the extension) needs different data that what `toc_tokens` contain. Uhg. To implement option 2 we also need to implement option 3. Every way I look at this, option 1 is the most attractive to me. Maybe there is something obvious I'm missing.

pawamoy: Thanks for tagging me @waylan. I was following this discussion but didn't have much to bring to the table. Basically: I don't mind breaking changes, as long as there actually*is* a solution for mkdocstrings to keep its features. Is there, with this PR merged? I understand the need to sanitize things, but does it leave us an escape hatch to customize headings, navigation and ToC with HTML (at least spans with classes)? Where is this escape hatch? If not in Markdown extensions, MkDocs plugins? MkDocs themes? I would appreciate some guidance on where to look for a solution if this PR is going to be merged as is 🙂

waylan: Apparently, I'm still not clear on how mkdocstrings works here. I just tried this on the pre-this-fix version of Markdown and this is what I get:

```python
>>> md = markdown.Markdown(extensions=['toc', 'attr_list'])
>>> md.convert('# Header {data-toc-label="<code class=\"doc-symbol doc-symbol-toc doc-symbol-module\"></code>&nbsp;markdown"}')
'<h1 id="header">Header</h1>'
>>> md.toc_tokens
[{'level': 1, 'id': 'header', 'name': '&nbsp;markdown', 'children': []}]
```

Specifically, the `code` tag is stripped out of the `toc_tokens`. Or more specifically, the behavior is exactly the same (with the exception of the new `html` attribute) before and after the change. So I don't understand how mkdocstrings is preserving the `code` tag in `data-toc-label`.

pawamoy: Once Jinja templates have been rendered to HTML, this HTML is stashed, and only headings are reinserted:

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/extension.py#L124-L130

The headings coming from docstrings are recorded in a nested Markdown converter:

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/rendering.py#L213-L222

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/base.py#L362-L371 

While the HTML headings coming from Jinja templates are recorded by our `do_heading` Jinja filter:

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/base.py#L306-L313

Later, right after `toc` has finished processing these headings, we remove these headings since theyll appear again after the HTML is unstashed:

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/extension.py#L235-L249

Maybe the important bit here is that headings are written directly as HTML (never Markdown), then stashed and unstashed at the end. Or maybe it's how the heading filter works:

https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/base.py#L318-L337

waylan: > While the HTML headings coming from Jinja templates are recorded by our `do_heading` Jinja filter:

I got everything up to this point. It's what happens after this, that I can't seem to follow.

> Later, right after `toc` has finished processing these headings, we remove these headings since theyll appear again after the HTML is unstashed:

All this does is extract the text of an element (and its children) and insert that as text of the parent. But if that text is raw HTML, it will get escaped when the parent is serialized. So I'm not sure how that works. And its not clear to me what this is even supposed to accomplish.

> Maybe the important bit here is that headings are written directly as HTML (never Markdown), then stashed and unstashed at the end. Or maybe it's how the heading filter works:
> 
> https://github.com/mkdocstrings/mkdocstrings/blob/628f3af226bbd76b58d51baf0628cbbee4260ba1/src/mkdocstrings/handlers/base.py#L318-L337

This is the bit that really confuses me. It seems to me that the `do_heading` method is called by mkdocstrings' jinja templates. So this is called before Markdown is even run right? And this returns raw HTML, inserting that raw HTML into a Markdown document, which Markdown would escape and/or strip. Seems to me like it should be stashing the raw HTML and returning some placeholder text and then the _PostProcessor should be stepping through `md.toc_tokens` and swaping out placeholders for the stashed raw HTML.  As a reminder, once the `toc` extension is done running, the `toc_tokens` are in their final state. So any processor which runs after them could modify them in any way without concern that Markdown will alter them further.

pawamoy: > So this is called before Markdown is even run right?

No, Jinja templates are rendered *during* Markdown conversion. The autodoc instruction like `::: path.to.object` are actually processed by a Markdown extension. This extension then does all the Jinja rendering, and stashes the resulting HTML.

oprypin: I saw you guys struggling with a repro but had no time to reply :crying_cat_face: 

Here's a repro of what's happening:

```python
import xml.etree.ElementTree as etree
import markdown
import markdown.treeprocessors
from markupsafe import Markup

class MyTreeprocessor(markdown.treeprocessors.Treeprocessor):
    def run(self, root):
        el = etree.Element("h2", {
            "id": "test",
            "data-toc-label": Markup('<code class="foo"></code>&nbsp;Test')
        })
        root.append(el)

md = markdown.Markdown(extensions=["toc"])
md.treeprocessors.register(
    MyTreeprocessor(), "test", priority=20
)
md.convert("a")
print(md.toc_tokens)
```

Output before:
```
[{'level': 2, 'id': 'test', 'name': '<code class="foo"></code>&nbsp;Test', 'children': []}]
```
Output now:
```
[{'level': 2, 'id': 'test', 'name': '&nbsp;Test', 'html': '', 'children': []}]
```

waylan: > No, Jinja templates are rendered _during_ Markdown conversion. The autodoc instruction like `::: path.to.object` are actually processed by a Markdown extension. This extension then does all the Jinja rendering, and stashes the resulting HTML.

Okay, I was missing this detail. You are in a partially parsed document generating more Markdown which also needs to be parsed. That is why the added complexity exists.

> Here's a repro of what's happening:

Thank you @oprypin. The simple example is a big help. The new behavior is what should happen. The fact that it was not behaving that way before was a bug. Using the simplified example, here is how I would work around the fix:

```python
class MyTreeprocessor(markdown.treeprocessors.Treeprocessor):
    def run(self, root):
        id = "test"
        self.md.my_stash_object['id'] = '<code class="foo"></code>&nbsp;'
        el = etree.Element("h2", {
            "id": id,
            "data-toc-label": "Test"
        })
        root.append(el)

class MyPostProcessor(markdown.postprocessors.Postprocessor):
    def run(self, text):
        for token in self.md.toc_tokens:
            if token['id'] in self.md.my_stash_object:
                token['name'] = self.md.my_stash_object[token['id']] + token['name']
        return text
```

Obviously you will need to set up `md.my_stash_object` which I assume is just a dict. And I didn't address nested tokens. But that should get you started.

oprypin: By the way, based on the same example, don't you think that probably this `'html'` item should contain the unmodified content of `data-toc-label`?

oprypin: @waylan Regarding your latest example, it is worth noting that in mkdocstrings we don't even want to insert this `h2` element into the document in the first place, it was done *solely* so it's picked up into `toc_tokens`, and then there's another treeprocessor to clean up the heading afterwards. Which would now bring us to a total of 3 separate components just to be able to insert something into the TOC. It will work, yes. But, can't we come up with some clean API to do that? Of course one immediately thinks: fine, forget about inserting the headings, why don't you just modify `toc_tokens` directly in the first place? Well, the problem is how to know the position where to insert the headings relative to the surrounding parts of the document.

markdown/extensions/toc.py
```diff
@@ -93,11 +97,80 @@ def _html_sub(m: re.Match[str]) -> str:
...
+                if sup.tail:
+                    # Preserve the tail text
+                    siblings = list(parent)
+                    pos = siblings.index(sup)
+                    if pos == 0:
+                        parent.text = f'{parent.text or ""}{sup.tail}'
+                    else:
+                        sibling = siblings[pos - 1]
+                        sibling.tail = f'{sibling.tail or ""}{sup.tail}'
+                parent.remove(sup)
```
oprypin: There is a bug with removing multiple fnrefs within one parent. This is due to modifying an iterable while iterating over it.

Example input that produces bad output: `# Header 1[^1] plus 1[^1]`

markdown/extensions/toc.py
waylan: So here is my dilemma. If we had not had the bug which mkdocstrings relies on, then how would mkdocstrings have originally implemented this? I'm simply fixing a bug and implementing the behavior we always intended to have. If that breaks mkdocstrings, then that is intended. To be clear, I don't want to intentionally break mkdocstrings, but when it relies on a bug to function, I don't see any alternative here.

> By the way, based on the same example, don't you think that probably this `'html'` item should contain the unmodified content of `data-toc-label`?

No the `html` item contains the **rendered** inner-html of the heading. The rendered inner-html would not contain the `data-toc-label` as that attribute is assigned to the parent element. If we were to change to have the `html` item contain the outer-html (the h1-6 tag), even that would not contain the `data-toc-label`, as that attribute is removed by the toc extension when the data is extracted from the attribute. In other words, the `data-toc-label` attribute never makes it to the final rendered document.

On further reflection, I suppose we could possibly include another item in the token which would contain the original, unsanitized value of `data-toc-label`. A token might then look like this:

```
{
    'level': 2, 
    'id': 'test',
    'name': '&nbsp;Test',
    'data-toc-label': '<code class="foo"></code>&nbsp;Test',
    'html': '',
    'children': []
}
```

We could also possibly include a config option on the extension which optionally does not remove the `data-toc-label` attribute from heading elements. Although, that would require parsing the rendered HTML to retrieve the value, so I don't see much value there.

oprypin: @waylan I think there's some misunderstanding regarding what I meant so I'll open another comment thread

oprypin, line 1: I think for an element such as this:
`<h2 data-toc-label="foo">inner</h2>`

The result should be
`'id': 'foo',
    'name': 'foo',
    'html': 'foo',`

As currently in this PR, it is this for some reason:
`'id': 'foo',
    'name': 'foo',
    'html': 'inner',`

oprypin: The first part was understood correctly though. Sure it makes sense what you're saying about breaking mkdocstrings, I'm just saying the situation for extensions which want to add custom toc items is changing from very complicated to very very complicated, and maybe some API could be thought up to allow doing this in a better way

waylan: Why? `html` is the raw unsanitized html content of the heading. That is `inner` in your example. To be clear, in my opinion, the `html` item is not affected at all by `data-toc-label`. This allows one to still have access to the actual content of the heading even when `data-toc-label` overrides it. That is why I also suggested in a previous comment that perhaps we also include the raw unsanitized `data-toc-label`. In this way we give third parties access to everything. They get the sanitized default toc label in `name` and the raw unsantized content of both `html` and `data-toc-label`. Then they can do whatever they want with whichever of those items that they want.

waylan: I'm trying to provide an API by including the raw data in the toc_token. I'm not sure what else to do here.

pawamoy: @waylan Again I'm not against breaking changes if this is for the best, and if we have a way to keep our features. If Python-Markdown can provide a simplified API (standardized for others too!) for our use-case, that'd be great. If not, then we would simply appreciate that you give us time to adjust (going the very very complicated way) before releasing anything :slightly_smiling_face:

oprypin: The questions that `toc_tokens` answers are these:

1. What's the id of the heading
2. What's the title of the heading, sanitized

The only thing that logically makes sense to add to this list (and what I've been assuming this is about):

3. What's the title of the heading, unsanitized

Questions that you want to answer instead:
* What's the inner HTML of the heading (added in this PR)
* What's the value of the heading's attribute that can override the title (proposed by you for this PR)

---

**Crucially**, in the current state of this PR, there is still **no way at all** to answer question 3 above. With **both** additions, yes it will be possible to answer it, by always using the two new attributes together.

I think the good way to solve that is by answering this question directly and not bloat this with 2 new attributes that don't even have any consistency with existing ones.

So to recap

<table><tr><th>I propose</th><th>You propose</th></tr>
<tr><td width="50%">

• What's the title of the heading, sanitized?
  `'name'`

• What's the title of the heading, unsanitized?
  `'html'`

• What's the 'data-toc-label' value, sanitized?
  *missing*

• What's the 'data-toc-label' value?
  *missing, but easy to look up in treeprocessors*

</td><td width="50%">

• What's the title of the heading, sanitized?
  `'name'`

• What's the title of the heading, unsanitized?
  `'data-toc-label' or 'html'`

• What's the 'data-toc-label' value, sanitized?
  *missing*

• What's the 'data-toc-label' value?
  `'data-toc-label'`, *also easy to look up in treeprocessors*

</td></tr></table>

waylan: So my thinking is to give the user the raw data and let them chose what to do with it. Anything else results in us making assumptions about what is wanted by the user. In my experience, those assumptions never work out long term. Yes, they may meet mkdostrings' needs now, but are likely to not meet other's needs in the future. However, if we provide the raw data, then we can meet everyone's needs as anyone can take the raw data and do whatever they want with it. Does that mean that perhaps some users will be reimplementing some of the logic to reprocess the raw data, Yes, but that is the only way to meet everyone's needs.

```
• What's the title of the heading, unsanitized?
'data-toc-label' or 'html'
```

That is exactly makes my point. My proposal is better for any future unanticipated use cases. How do I know that some future user wants to make that decision for themselves and not be stuck with my logical for that choice? If I provide both, then they can do whatever they want without any restrictions imposed by me.

```
• What's the 'data-toc-label' value, sanitized?
missing
```

Not true. If `data-toc-label` was set, then the santized value of `data-toc-label` is the value assigned to `name`.

```
• What's the title of the heading, sanitized?
'name'
```

Only if  `data-toc-label` was not set (see above). But this was already the intended behavior prior to this change anyway. So it is not up for a change here except to patch the bug which in certain edge cases let unsantizied content through,

```
• What's the 'data-toc-label' value?
'data-toc-label', also easy to look up in treeprocessors
```

It is not easy to look up as it is removed by the toc extension... that is unless you run a treeprocessor beforehand, but in that case you are outside the toc extension anyway so I don't need to accommodate you. In fact, in that case you should probably be writing your own replacement for the toc extension.

As far as I can tell, the objection here is that my proposal requires you to write more code, which replicates some of the existing code in the toc extension. That is why I provided that code in easily callable functions. A user can take the raw data, call the same functions we already have, or do their own different thing, or do some hybrid combination. If that is the only objection, then I'm inclined to go forward with my proposal.

waylan: Here is a corrected comparison

<table><tr><th>@oprypin's proposal</th><th>@waylan's proposal</th></tr>
<tr><td width="50%">

• What's the title of the heading, sanitized?
  `name` if `data-toc-label` is not set, otherwise *missing*

• What's the title of the heading, unsanitized?
  `html` if `data-toc-label` is not set, otherwise *missing*

• What's the `data-toc-label` value, sanitized?
  it may be the the value of `name`, but the user doesn't have enough info to know for sure

• What's the `data-toc-label` value?
  *missing*

</td><td width="50%">

• What's the title of the heading, sanitized?
  `name` if `data-toc-label` is not set, otherwise obtainable from sanitizing `html` using the provided functions

• What's the title of the heading, unsanitized?
  `html`

• What's the `data-toc-label` value, sanitized?
  `name` if `data-toc-label` is set, otherwise not relevant

• What's the `data-toc-label` value?
  `data-toc-label`

</td></tr></table>

To be clear, I am considering "the title of the heading" to be the content of the heading element (h1-6), not the calculated value the TOC extension uses for the table of contents. The calculated value always has been, and always will be `name`.

oprypin: > To be clear, I am considering "the title of the heading" to be the content of the heading element (h1-6)

Yes that's the nonsensical part, no idea why you're considering it like that. Let me compare your table to the current state, then.

<table><tr><th>Current state</th><th>@waylan's proposal</th></tr>
<tr><td width="50%">

• What's the inner HTML of the heading, sanitized?
  *missing*

• What's the inner HTML of the heading, unsanitized?
  *missing*

• What's the 'data-toc-label' value, sanitized?
  *missing*

• What's the 'data-toc-label' value?
  *missing*

</td><td width="50%">

• What's the inner HTML of the heading, sanitized?
*missing*

• What's the inner HTML of the heading, unsanitized?
`html`

• What's the data-toc-label value, sanitized?
`name if data-toc-label else None`

• What's the data-toc-label value?
`data-toc-label`

</td></tr></table>

So currently it's not possible to do anything, huh. And yet somehow people do it.

---

I do get your point that more values are possible to calculate. And 1 big mistake on my side, in your proposal it is in fact possible to obtain possible to obtain "`data-toc-label` value, sanitized" in a roundabout way, I see.

oprypin: Anyway yes I see. The proposal can work this way, sure. :+1:

markdown/extensions/toc.py
waylan: Did you actually test this, because I did. Look again. I am not iterating over the tree. Rather the `for` loop iterates over a sequence of references to `Element` instances returned by `findall`. Then, within the loop, we obtain the position of the `Element` instance within the tree (which we would have to do anyway because `findall` doesn't provide that info). In other words, `pos` is recalculated on each loop based on the condition of the tree at that time. In any event, I added another test case (in 1355826) to ensure this continues to work.

oprypin: Oh I'm very sorry about the false report. I was editing the code while testing it, which was my mistake. I must have first replaced `findall` with `iter` and then observed this problem.

markdown/extensions/toc.py
```diff
@@ -93,11 +97,80 @@ def _html_sub(m: re.Match[str]) -> str:
...
+    return el
```
oprypin: If I may offer a different implementation...

(You can also see it here https://github.com/mkdocs/mkdocs/pull/3578/files#diff-cda68073304c4a9b7bcfad3624acad56a02065877a9067cf5c1d7f79d6427954)

The idea of this is:
1. Do a deepcopy only if deemed necessary, don't guarantee returning a copy. Rename the method accordingly.

3. Apply the XPath lookup to find *parents* of 'sup' elements rather than 'sup' elements themselves, to not necessitate another global lookup through the whole element.

```suggestion
def remove_fnrefs(root: etree.Element) -> etree.Element:
    """ Remove footnote references from a copy of the element, if any are present. """
    # Remove footnote references, which look like this: `<sup id="fnref:1">...</sup>`.
    # If there are no `sup` elements, then nothing to do.
    if next(root.iter('sup'), None) is None:
        return root
    root = deepcopy(root)
    # Find parent elements that contain `sup` elements.
    for parent in root.findall('.//sup/..'):
        carry_text = ""
        for child in reversed(parent):  # Reversed for the ability to mutate during iteration.
            # Remove matching footnote references but carry any `tail` text to preceding elements.
            if child.tag == 'sup' and child.get('id', '').startswith('fnref'):
                carry_text = f'{child.tail or ""}{carry_text}'
                parent.remove(child)
            elif carry_text:
                child.tail = f'{child.tail or ""}{carry_text}'
                carry_text = ""
        if carry_text:
            parent.text = f'{parent.text or ""}{carry_text}'
    return root
```

markdown/extensions/toc.py
```diff
@@ -300,27 +373,30 @@ def run(self, doc: etree.Element) -> None:
         for el in doc.iter():
             if isinstance(el.tag, str) and self.header_rgx.match(el.tag):
                 self.set_level(el)
-                text = get_name(el)
+                innerhtml = render_inner_html(copy_element(el), self.md)
```
oprypin:
```suggestion
                innerhtml = render_inner_html(remove_fnrefs(el), self.md)
```

oprypin (commented):
The change looks good. This will be nice :)

waylan:
I finished updating the tests and documentation. I think this is ready to go.

waylan:
@oprypin @pawamoy I wanted to check with you before I did a release of this. As this is a change in behavior that affects your projects (in that a bug was fixed which allowed markup to be passed through to `toc_token['name']` in certain circumstance), I wanted to make sure you where ready for the release before I publish it.

Presumably at a minimum, MkDocs will need to do a release which includes mkdocs/mkdocs#3578 before I do my release. And I am assuming mkdostrings will want to do a release which takes advantage of mkdocs/mkdocs#3578 as well.

For the record, I intend to release this as version 3.6, so any earlier releases of your projects will likely want to be constrained to `markdown<3.6` for any users who want markup in their toc token names.

pawamoy:
Thanks for the heads up. I'm not ready for this. I'll start by adding an upper bound in mkdocstrings and publish a new release. I should be able to do that today, but if not, before the end of the week for sure :slightly_smiling_face: I'll report back once it's done!

oprypin:
I think actually there's no direct interaction with MkDocs, the release can be made here first. Extracting the primary heading has a separate implementation in MkDocs and extracting secondary headings will just change for the better with this release but still nothing to be done in MkDocs

waylan:
Well, there definitely is interaction with mkdocstrings. According to my own testing, the [show_symbol_type_toc](https://mkdocstrings.github.io/python/usage/configuration/headings/#show_symbol_type_toc) feature breaks with this change. Presumably, mkdocstings will need to retrieve the value from `token['data-toc-label']` rather than `token['name']` as it does now. Therefore, at a minimum, I'll wait for an indication that mkdocstrings is ready before doing a release.

pawamoy:
I've published a new version of mkdocstrings-python with an upper bound on Python-Markdown 3.6, you can release it :)

pawamoy:
Hello again! Today someone reported that we should support version 3.6, so I made time to work on this. And it turns out I can do this very simple thing:

```python
class _TocLabelsTreeProcessor(Treeprocessor):
    def run(self, root: Element) -> None:  # noqa: ARG002
        self._override_toc_labels(self.md.toc_tokens)  # type: ignore[attr-defined]

    def _override_toc_labels(self, tokens: list) -> None:
        for token in tokens:
            if token["name"] != token["data-toc-label"]:
                token["name"] = token["data-toc-label"]
                self._override_toc_labels(token["children"])
```

...and register this tree processor in our outer MkdocstringsExtension:

```python
        md.treeprocessors.register(
            _TocLabelsTreeProcessor(md),
            "mkdocstrings_post_toc_labels",
            priority=4.5,  # Right after 'toc'.
        )
```

I suppose it's a bit brutal and instead of blindly overwriting names with toc-labels, I should add a way to detect when it comes from mkdocstrings and overwrite only those? For example, `if token["name"] != token["data-toc-label"] and token["data-toc-label"].startswith("<code")`?

Otherwise it means I'm just removing the additional safety that was brought with this PR, right? Is this really bad or is there no practical impact :thinking:?

waylan:
The primary concern for MkDocs is that the page title should not ever contain any markup (to avoid markup in `<title>` etc). As the page title comes from `token['name']`, then by extension, `token['name']` should never contain any markup... but that only really matters for the top `<h1>` in a page. That said, in Markdown's own documentation, mkdocstrings does set a value to `data-toc-label` for the top `<h1>` on each page it generates.

So, yes, to avoid breaking themes (by causing them to generate invalid HTML), `token['name']` should never contain any markup. The correct fix is to pull the value from `token['data-toc-label']` (and why I added it). However, that would require an edit to theme templates when building the TOC. And that is something theme devs need to do, unless you can provide a way to override that block of a theme. IIRC, there is not a straightforward way for an extension to do that (it interferes with a user's overrides).

So, in the end, we have a chicken and egg problem. Theme's need to support `token['data-toc-label']` directly when building a page's toc, but they are not likely to do that unless users and/or extensions are assigning values to `token['data-toc-label']` which contain markup.

waylan:
Two other factors come into play,

1. For a theme to access `token['data-toc-label']`, then MkDocs needs to pass that on. I don't think it does (needs confirmation).
2. In mkdocs/mkdocs#3578, the same sanitation code is used by MkDocs to sanitize page titles. Therefore, assigning markup to `token['name'] may not be a problem after all as MkDocs will sanitize it before setting the value to the page title.

In other words, at this time, your suggested solution may actually be the only workable one.

pawamoy:
Thank you!

I'm not sure to understand why there is a data-toc-label attribute if it is only use to override the item name which is otherwise derived from the heading's text/id :thinking: I thought data-toc-label had *exactly* this use of providing an alternate label for the table of contents (label as, displayable text that doesn't affect anything else). But from your answer I gather that themes actually never used this attribute and always relied on the name itself, and it only *seemed* to work as I expected it to work because the toc-label was not sanitized properly. Now that it is, it unveiled the true inner working of themes, which is to use the name directly. Makes sense!

Anyway, I'll go with the solution above then, thanks! :slightly_smiling_face:

waylan:
Prior to this present change, the assumption was that both `name` and `data-toc-label` never contained markup. It was a weird anomaly that this ever worked for you. Therefore, with this change only sanitized content of `data-toc-label` gets copied to `name` (no markup is preserved as `name` still never has markup). The thinking is that if you want the raw content of `data-toc-label` then you need to access it directly from `data-toc-label`. In other words, with this change, we are officially supporting using markup in `data-toc-label` but to access that markup you can't get if from `name`; you have to get it from `data-toc-label` directly.

PR #1461, opened 2024-04-24
Refactor `abbr` Extension

### LINKED ISSUE #1460: `abbr` extension breaks title attribute added using `attr_list`
### Bug description

``` markdown
*[abbr]: Abbreviation Definition

![Image with abbr in title](abbr.png){title="Image with abbr in title"}
```

This renders the attribute code as plain text because the Abbreviation extension modifies the text in the attribute causing it to no longer be a valid attribute list.

### Reproduction

[9.5.18-abbr-attr_list-conflict.zip](https://github.com/Python-Markdown/markdown/files/15077227/9.5.18-abbr-attr_list-conflict.zip)

### Steps to reproduce

1. Enable the `abbr` and `attr_list` extensions
2. Include an abbreviation definition on a page
3. Use the abbreviation in a title attribute for an object using the attr_list syntax

waylan:
A new `AbbrTreeprocessor` has been introduced, which replaces the now deprecated `AbbrInlineProcessor`. Abbreviation processing now happens after Attribute Lists, avoiding a conflict between the two extensions. Fixes #1460.

As an aside, with this change we now have a collection of abbrs. It should probably be emptied on `reset()`. Also, it is now feasible to add a config option for predefined abbrs -- a feature request we have received multiple times in the past.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -18,7 +18,8 @@ A new `AbbrTreeprocessor` has been introduced, which replaces the now deprecated
 `AbbrInlineProcessor`. Abbreviation processing now happens after Attribute Lists,
 avoiding a conflict between the two extensions (#1460).
 
-The `AbbrPreprocessor` class has been renamed to `AbbrBlockprocessor`, which better reflects what it is. `AbbrPreprocessor` has been deprecated.
+The `AbbrPreprocessor` class has been renamed to `AbbrBlockprocessor`, which
+better reflects what it is. `AbbrPreprocessor` has been deprecated.
 
 A call to `Markdown.reset()` now clears all previously defined abbreviations.
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -18,6 +18,8 @@ A new `AbbrTreeprocessor` has been introduced, which replaces the now deprecated
 `AbbrInlineProcessor`. Abbreviation processing now happens after Attribute Lists,
 avoiding a conflict between the two extensions (#1460).
 
+The `AbbrPreprocessor` class has been renamed to `AbbrBlockprocessor`, which better reflects what it is. `AbbrPreprocessor` has been deprecated.
+
 A call to `Markdown.reset()` now clears all previously defined abbreviations.
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


waylan:
I have added support for `reset` but not a config option to predefine abbreviations. I'm not sure I want to add that and a quick search didn't turn up any requests. Maybe I was thinking of something else.

docs/changelog.md
```diff
@@ -10,6 +10,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
...
 
+### Changed
+
+#### Refactor `abbr` Extension
+
+A new `AbbrTreeprocessor` has been introduced, which replaces the now deprecated
+`AbbrInlineProcessor`. Abbreviation processing now happens after Attribute Lists,
+avoiding a conflict between the two extensions (#1460).
+
+The `AbbrPreprocessor` class has been renamed to `AbbrBlockprocessor`, which better reflects what it is. `AbbrPreprocessor` has been deprecated.
```
mitya57: This line should be probably wrapped too?

markdown/extensions/abbr.py
```diff
@@ -25,41 +25,103 @@
...
+                tail = tail[:m.start()]
+            el.tail = tail
+
+    def run(self, root: etree.Element) -> etree.Element | None:
+        ''' Step through tree to find known abbreviations. '''
+        if not self.abbrs:
+            # No abbreviations defined. Skip running processor.
+            return
+        # Build and compile regex
+        self.RE = re.compile(f"\\b(?:{ '|'.join(re.escape(key) for key in self.abbrs.keys()) })\\b")
```
mitya57: This can be slightly shortened by removing `.keys()`:

```suggestion
        self.RE = re.compile(f"\\b(?:{ '|'.join(re.escape(key) for key in self.abbrs) })\\b")
```

markdown/extensions/abbr.py
```diff
@@ -25,41 +25,103 @@
...
+        # Step through tree and modify on matches
+        self.iter_element(root)
+        return
```
mitya57: This line is also redundant and can be removed.

Issue #1465, opened 2024-05-20
Abbreviation processing order alternative


nbanyan:
Currently the `abbr` extension implements the last definition of an abbreviation.
In MkDocs with the `pymdownx.snippet` extension, a glossary can be auto-appended to Markdown pages.

The problem is that this prevents pages from overriding abbreviations from the glossary.  To circumvent this I've been manually adding the glossary snippet to the top of all pages. But this requires some page specific abbreviations to be declared before the snippet and some after.  As the lists continue to grow this can quickly cause some abbreviations to be incorrect.

**Example of the currently required order:**

In `glossary.md`:
``` markdown title="glossary.md"
*[process]: process definition one
*[ABC]: Abbreviation definition one
*[XYZ]: Abbreviation definition two (newly added)
```

In `page.md`:
``` markdown title="page.md"
*[specialized process]: process definition specialization
--8<-- "glossary.md"
*[XYZ-A]: Unrealized Abbreviation due to change in glossary.md
*[ABC]: Abbreviation definition one redefinition
```
* 'specialized process' must be defined before the snippet so the entire phrase is processed before 'process' is implemented
* 'ABC' must be defined after the snippet otherwise it will be overridden by the glossary definition
* 'XYZ-A' is no longer implemented since the snippet now defines and implements 'XYZ' first

**Proposals:**

* Option 1: An option to have the `abbr` extension only keep/process the first instance of an abbreviation instead of the last.
* Option 2: An option to sort the abbreviation list to process 'superset' abbreviations first to transparently illuminate the issues with terms like the 'specialized process' and 'XYZ-A' in the example.

Both options would allow effective use of auto-append and illuminate the issue of needing to analyze how abbreviations are processed every time one is written and reassessing/adjusting the ordering in pages when a new term is added to the glossary.

The second option would require more processing time, but prevent issues (from Option1) with a page declaring an abbreviation that breaks an abbreviation in an auto-appended glossary.

waylan:
> this requires some page specific abbreviations to be declared before the snippet and some after.

Personally, this is exactly how I would expect to solve this issue, which means we don't need to make any changes. It is not clear to me how this is resulting in incorrect abbreviations. Perhaps an example would help. 

Or are you saying that sometimes you get the order wrong and that results in the errors. If that is the case, then I don't see how your suggested changes would help. You would still need to ensure correct order, even if that order would be different.

[...]

PR #1467, opened 2024-06-03
Abbr Extension: Definition Sorting and Glossary storage

nbanyan:
Several changes/features here.  All input welcome on each.  

* **Sorting**: A simple sort by length to process longer abbreviations first solves all instances of smaller abbreviations preventing longer ones from being used.
* `glossary`: An option to include a Markdown file with abbreviations to apply to every page.
* `use_last_abbr`: An option to only keep the first instance of an abbreviation (useful when using other methods to append glossaries to pages).

The glossary ended up being the more complex of the three and involved replicating some of the code/process from **core.py**.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -33,12 +33,8 @@ abbreviation, which may be useful for documents that uses two terms with
 identical abbreviations.
 
 Added an optional `glossary` configuration option to the abbreviations extension.
-This provides a simple and efficient way to apply abbreviations to every page.
-
-Added an optional `use_last_abbr` configuration option to the abbreviations
-extension. Default (`True`) maintains the existing behavior. `False` causes
-the extension to only use the first instance of an abbreviation, rather than
-the last. 
+This provides a simple and efficient way to apply a dictionary of abbreviations
+to every page.
 
 
 ### Fixed
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -23,6 +23,10 @@ better reflects what it is. `AbbrPreprocessor` has been deprecated.
 
 A call to `Markdown.reset()` now clears all previously defined abbreviations.
 
+Abbreviations are now sorted by length before executing `AbbrTreeprocessor`
+to ensure that multi-word abbreviations are implemented even if an abbreviation
+exists for one of those component words.
+
 ### Fixed
 
 * Fixed links to source code on GitHub from the documentation (#1453).
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,16 +27,18 @@ Abbreviations are now sorted by length before executing `AbbrTreeprocessor`
 to ensure that multi-word abbreviations are implemented even if an abbreviation
 exists for one of those component words. (#1465)
 
-Added an optional `use_last_abbr` configuration option to the abbreviations
-extension. Default (`True`) maintains the existing behavior. `False` causes
-the extension to only use the first instance of an abbreviation, rather than
-the last.
-
 Empty abbreviations are now skipped by `AbbrTreeprocessor`. This avoids applying
 abbr tags to text without a title value. This also allows disabling an
 abbreviation, which may be useful for documents that uses two terms with
 identical abbreviations.
 
+Added an optional `glossary` configuration option to the abbreviations extension.
+This provides a simple and efficient way to apply abbreviations to every page.
+
+Added an optional `use_last_abbr` configuration option to the abbreviations
+extension. Default (`True`) maintains the existing behavior. `False` causes
+the extension to only use the first instance of an abbreviation, rather than
+the last. 
 
 
 ### Fixed
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -25,7 +25,19 @@ A call to `Markdown.reset()` now clears all previously defined abbreviations.
 
 Abbreviations are now sorted by length before executing `AbbrTreeprocessor`
 to ensure that multi-word abbreviations are implemented even if an abbreviation
-exists for one of those component words.
+exists for one of those component words. (#1465)
+
+Added an optional `use_last_abbr` configuration option to the abbreviations
+extension. Default (`True`) maintains the existing behavior. `False` causes
+the extension to only use the first instance of an abbreviation, rather than
+the last.
+
+Empty abbreviations are now skipped by `AbbrTreeprocessor`. This avoids applying
+abbr tags to text without a title value. This also allows disabling an
+abbreviation, which may be useful for documents that uses two terms with
+identical abbreviations.
+
+
 
 ### Fixed
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,14 +27,15 @@ Abbreviations are now sorted by length before executing `AbbrTreeprocessor`
 to ensure that multi-word abbreviations are implemented even if an abbreviation
 exists for one of those component words. (#1465)
 
-Empty abbreviations are now skipped by `AbbrTreeprocessor`. This avoids applying
-abbr tags to text without a title value. This also allows disabling an
-abbreviation, which may be useful for documents that uses two terms with
-identical abbreviations.
+Abbreviations without a definition are now ignored. This avoids applying
+abbr tags to text without a title value.
 
 Added an optional `glossary` configuration option to the abbreviations extension.
 This provides a simple and efficient way to apply a dictionary of abbreviations
 to every page.
+
+Abbreviations can now be disabled by setting their definition to `""` or `''`.
+This can be useful when using the `glossary` option.
 
 
 ### Fixed
--- a/docs/extensions/abbreviations.md
+++ b/docs/extensions/abbreviations.md
@@ -51,7 +51,20 @@ Usage
 See [Extensions](index.md) for general extension usage. Use `abbr` as the name
 of the extension.
 
-This extension does not accept any special configuration options.
+The following options are provided to configure the output:
+
+* **`use_last_abbr`**:
+    `True` to use the last instance of an abbreviation, rather than the first instance.
+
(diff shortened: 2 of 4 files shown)


waylan:
So far I've only taken a cursory look at your changes. Here are my initial thoughts.

> * **Sorting**: A simple sort by length to process longer abbreviations first solves all instances of smaller abbreviations preventing longer ones from being used.

This is good. I should have caught the issue this fixes with the recent refactor. Thank you.

> * `glossary`: An option to include a Markdown file with abbreviations to apply to every page.

I'm not sure I am comfortable with this. In our discussion in #1465 I suggested allowing a user to pass in a Python object (such as a dict), not importing a file of Markdown content. I'm inclined to reject this on the basis that extensions already exist which allow Markdown content to be imported.

> * `use_last_abbr`: An option to only keep the first instance of an abbreviation (useful when using other methods to append glossaries to pages).

As previously explained, I don't see any value in this. You still need to ensure your definitions are in a certain order. That order is now reversed which is counterintuitive (IMO) and works the opposite of every other reference-type syntax (reference links, footnotes, etc) in Markdown. I realize this is an option which if off by default, but with there being no value to changing the setting, then why even offer it? That said, if you can provide a concrete example which demonstrates its value and changes my mind, then I would suggest renaming it `use_first_abbr` with the default being `False`. That name more clearly aligns with the description of the option.

nbanyan:
Discarded `use_last_abbr` and changed `glossary` to accept and manage dictionary objects directly.

docs/changelog.md
```diff
@@ -23,6 +23,24 @@ better reflects what it is. `AbbrPreprocessor` has been deprecated.
 
 A call to `Markdown.reset()` now clears all previously defined abbreviations.
 
+Abbreviations are now sorted by length before executing `AbbrTreeprocessor`
+to ensure that multi-word abbreviations are implemented even if an abbreviation
+exists for one of those component words. (#1465)
+
+Empty abbreviations are now skipped by `AbbrTreeprocessor`. This avoids applying
```
waylan: Changelog needs to be updated to reflect latest changes.

docs/extensions/abbreviations.md
```diff
@@ -46,7 +51,20 @@ Usage
 See [Extensions](index.md) for general extension usage. Use `abbr` as the name
 of the extension.
 
-This extension does not accept any special configuration options.
+The following options are provided to configure the output:
+
+* **`use_last_abbr`**:
```
waylan: Docs need updated to reflect latest changes.

markdown/extensions/abbr.py
```diff
@@ -23,6 +23,7 @@
 from __future__ import annotations
 
 from . import Extension
+from ..util import parseBoolValue
```
waylan: It looks like this is not called and can be removed.

markdown/extensions/abbr.py
```diff
@@ -41,18 +42,41 @@ class AbbrExtension(Extension):
...
 
     def extendMarkdown(self, md):
         """ Insert `AbbrTreeprocessor` and `AbbrBlockprocessor`. """
+        if (self.config['glossary'][0]):
+            self.load_glossary(self.config['glossary'][0])
+        self.abbrs.update(self.glossary)
         md.registerExtension(self)
         md.treeprocessors.register(AbbrTreeprocessor(md, self.abbrs), 'abbr', 7)
-        md.parser.blockprocessors.register(AbbrBlockprocessor(md.parser, self.abbrs), 'abbr', 16)
+        md.parser.blockprocessors.register(AbbrBlockprocessor(md.parser, self.abbrs, self.getConfigs()), 'abbr', 16)
```
waylan: Not sure why configs are being passed here as they are not used by the blockprocessor.

markdown/extensions/abbr.py
```diff
@@ -69,13 +93,14 @@ def iter_element(self, el: etree.Element, parent: etree.Element | None = None) -
             self.iter_element(child, el)
         if text := el.text:
             for m in reversed(list(self.RE.finditer(text))):
-                abbr = etree.Element('abbr', {'title': self.abbrs[m.group(0)]})
-                abbr.text = AtomicString(m.group(0))
-                abbr.tail = text[m.end():]
-                el.insert(0, abbr)
-                text = text[:m.start()]
+                if self.abbrs[m.group(0)]:
```
waylan, line 95: Curious why you felt the need to wrap this in an if statement. Seem reasonable to expect that if we have a match this should always be True. If there is a specific case where it is not, then we need a test case for it.

markdown/extensions/abbr.py
```diff
@@ -102,7 +129,7 @@ class AbbrBlockprocessor(BlockProcessor):
 
     RE = re.compile(r'^[*]\[(?P<abbr>[^\\]*?)\][ ]?:[ ]*\n?[ ]*(?P<title>.*)$', re.MULTILINE)
 
-    def __init__(self, parser: BlockParser, abbrs: dict):
+    def __init__(self, parser: BlockParser, abbrs: dict, config: dict):
```
waylan: Again, this appears to be not needed.

tests/test_syntax/extensions/test_abbr.py
```diff
@@ -119,21 +120,67 @@ def test_abbr_multiple_nested(self):
             )
         )
 
-    def test_abbr_override(self):
+    def test_abbr_glossary(self):
```
waylan, line 139: I understand why the new test was added, but why was the existing test removed/replaced? Shouldn't we have both tests?

waylan (changes requested):

markdown/extensions/abbr.py
nbanyan: This handles the situation in `test_abbr_empty()` of declaring an abbreviation without a definition. Without that test it adds `abbr` tags with an empty title instead of leaving it unmodified.

tests/test_syntax/extensions/test_abbr.py
nbanyan: Restored. I forgot that I didn't add that test and accidentally removed it when removing artifacts from `use_last_abbr`.

markdown/extensions/abbr.py
nbanyan: They were used for `use_last_abbr` and are no longer needed.

docs/extensions/abbreviations.md
nbanyan: Updated. Let me know if more details are needed for the `glossary` option and functions.

docs/changelog.md
nbanyan: Updated.

markdown/extensions/abbr.py
waylan: Ah, right. I wasn't paying attention to what you were actually testing. You are checking the value, not the key. And I suppose a user could pass in a glossary with keys which contain empty values. Yes, this needs to stay here.

tests/test_syntax/extensions/test_abbr.py
```diff
@@ -383,6 +446,47 @@ def test_abbr_with_attr_list(self):
...
+
+                Testing document text.
+                """
+            ),
+            self.dedent(
+                """
+                <p>abbr</p>\n<p>Testing document text.</p>
+                """
+            )
+        )
```
waylan: I don't think this is the correct behavior for this test case. The syntax for definitions are modeled after reference style links. Notice how an empty link reference behaves:

```markdown
[empty link reference][empty]

[empty]: 
```

The above renders as

```html
<p>[empty link reference][empty]</p>
<p>[empty]: </p>
```

Specifically, an empty link reference is not recognized as one at all. I think this is probably how abbreviation definitions should behave as well. The fact that a test for an empty definition didn't exist suggests that this edge case wasn't given any consideration in the past. Thanks for catching another bug.

What this means is that, given the current input, the expected output for `test_abbr_empty()` would be:

```html
<p><abbr title="Abbreviation Definition">abbr</abbr></p>
<p>*[abbr]:</p>
<p>Testing document text.</p>
```

Of course, that would also mean that your attempt to create a way to override/remove a previously defined abbreviation would no longer work. Perhaps a reasonable workaround would be to add support for wrapping the definition in quotes. Then empty quotes would be parsed as an empty definition.

```markdown
*[not-a-def]:
*[empty-def]: ""
```

waylan: On the other hand, there is no mechanism to predefine link references. However, with this change we now have a way to predefine abbr definitions. Therefore, it is reasonable to expect to be able to define  an empty definition to override/remove it. Therefore, maybe we can just leave things as-is. I'll need to think some more about this.

nbanyan: I like the more declarative nature of the `*[empty-def]: ""` syntax.
Potential alternatives to `""`: `[]`, `{}`, `_`, `...`

There are likely others and I don't know which ones feel more Markdown-ish.

waylan: Quotes are used in both inline and reference style [links](https://daringfireball.net/projects/markdown/syntax#link) for titles.  In reference style links the title can also be wrapped in regular parentheses (`()`), but in all my years I have never seen anyone actually use that. I suppose maybe it could be useful if the content needed to include quote characters. However, here we only need the wrapping characters when the content is empty, so that would be less of a concern. I think using quotes (supporting both single and double) makes the most sense in this context.

nbanyan: Added handling for ignoring `*[not-a-def]:` and removing definitions with `*[empty-def]: ""` or `*[empty-def]: ''`
Also, while testing I found that `*[]: Bad Definition` resulted in abbr tags everywhere, so added testing to avoid that.

waylan (approved):

waylan:
@nbanyan thanks for the work on this. Much appreciated.

PR #1475, opened 2024-08-16
Bump version to 3.7

waylan:


waylan:
It appears that the docs build fails due to [griffe v. 1.0](https://github.com/mkdocstrings/griffe/releases/tag/1.0.0) (a dependency of `mkdocstrings[python]`).

PR #1483, opened 2024-09-27
Fix DRY by adding create_element method in abbr extension

yves-chevallier:
I've identified some duplicated code in `abbr.py`. In my specific use case, I'd prefer to use an `a` element instead of `abbr` for another extension. It would be cleaner and more flexible to simply override the `create_element` method, adhering more to the DRY principle.

Not entirely sure if this change aligns with the project's goals, but I thought it was worth proposing.


waylan:
I'm indifferent to this being merged. However, if it is to be merged, then it needs a changelog entry. The failing `checkspelling` test can be ignored as it is not related to this change.

waylan (approved):

PR #1494, opened 2024-12-12
Backslash Unescape IDs set via `attr_list` on `toc`

### LINKED ISSUE #1493: toc: Unusual characters in heading ids not well supported
I noticed that `toc` encodes characters like `*` as `\x0242\x03`, 42 being the index of `*` in the ASCII table. This causes a discrepancy between the permalink of a heading and the link in the table of contents.

```bash
mkdir /tmp/toc
cd /tmp/toc
python -m venv .venv
. .venv/bin/activate
python -m pip install mkdocs
python -m mkdocs new .
```

index file:

```md
# Welcome

Demonstrating an issue with HTML ids and `toc`.

## `*Foo*` { id="\*Foo\*" }

- Click on `*Foo*`'s permalink: `#*Foo*` in the URL.
- Click on `*Foo*` in the table of contents: `#%0242%03Foo%0242%03` in the URL.
```

mkdocs config:

```yaml
site_name: My Docs

markdown_extensions:
- attr_list
- toc:
    permalink: true
```

Serve and observe the behavior described in the index page.

I'm not saying this is a bug. I'm just curious if this is expected, and whether there would be a way improve support for headings with such "unusual" ids. This would help for the work I'm doing with mkdocstrings, where we try to expand our languages support, and some languages might use uncommon characters in object identifiers. Not only `toc` would have to work, but also mkdocs-autorefs, which picks up ids from the table of contents when registering URLs and anchors to objects.

I believe HTML5 supports any kind of characters in ids. Some of them just cause a bit of pain, like `.` or `#`, because they then need to be escaped in CSS selectors.

waylan:
Fixes #1493.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -14,6 +14,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 * DRY fix in `abbr` extension by introducing method `create_element` (#1483).
 
+### Fixed
+
+* Backslash Unescape IDs set via `attr_list` on `toc` (#1493).
+
 ## [3.7] -- 2024-08-16
 
 ### Changed
--- a/markdown/extensions/toc.py
+++ b/markdown/extensions/toc.py
@@ -391,7 +391,7 @@ class TocTreeprocessor(Treeprocessor):
                 if int(el.tag[-1]) >= self.toc_top and int(el.tag[-1]) <= self.toc_bottom:
                     toc_tokens.append({
                         'level': int(el.tag[-1]),
-                        'id': el.attrib["id"],
+                        'id': unescape(el.attrib["id"]),
                         'name': name,
                         'html': innerhtml,
                         'data-toc-label': data_toc_label
--- a/tests/test_syntax/extensions/test_attr_list.py
+++ b/tests/test_syntax/extensions/test_attr_list.py
@@ -72,6 +72,13 @@ class TestAttrList(TestCase):
             '<h1 foo="&quot;bar">Heading</h1>'
         )
 
+    def test_backslash_escape_value(self):
+        self.assertMarkdownRenders(
+            '# `*Foo*` { id="\\*Foo\\*" }',
+            '<h1 id="*Foo*"><code>*Foo*</code></h1>'
+        )
+
+
     def test_table_td(self):
         self.assertMarkdownRenders(
             self.dedent(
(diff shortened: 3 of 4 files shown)


facelessuser (approved):

pawamoy:
Thank you @waylan for the super fast fix, and @facelessuser for reviewing :heart:

pawamoy:
I'm looking into the autorefs warning.

pawamoy:
OK here's what going wrong:

- when rendering API docs, we enable automatic summaries (`summary: true` in mkdocs.yml)
- previously this was an Insiders feature, and it was recently made available to everyone, so it's now active in this project too
- when we render the `extensions` sub-package, we summarize each submodule
- to summarize items, we render the first line of their docstring
- in the case of the `admonition` module, its first line contains a reference: `[rST][]`
- since we only render this first line, the reference is unresolved by Python-Markdown itself, and autorefs kicks in (and fails too)

Possible solutions:

- don't use "defered inline references" (not sure what to call them) in docstrings summaries (the first line) and instead put their URL directly within them (so `[rST](https://...)` instead of `[rST][]`), or move them lower, into the docstring body
- use pymdownx.snippets to auto-append references at the end of each markdown document (done on the fly, no modification to any file, except maybe to remove all refs from everywhere and gather them in the single auto-appended file, see https://facelessuser.github.io/pymdown-extensions/extensions/snippets/#auto-append-snippets)
- disable auto-summaries

Let me know if I should open a new issue and PR :slightly_smiling_face:

waylan:
The failing test is not related to any changes here so should be handled separately. I'm ignoring it for this PR so a new PR would be appropriate. But thank you for explaining why it was failing.

As an aside, if you are rendering a partial document, then you need to account for reference links and properly render them. IMO there should be no requirement that the document author needs to account for this and alter the way links are formatted. In fact, personally, my preferred style is to always use reference links in documentation. We don't actually have a defined style guide for this project, but it is reasonable to expect that some of your users might.

Yes, I am suggesting that the issue is with mkdocstrings' behavior. I would expect one of two behaviors (possibly both options available via a config option):

1. Reference links are resolved.
2. Links are striped from summaries and only the label text is included.

As far as a temporary fix goes, we should probably either:

1. Disable summaries. But I'm not clear on what they look like so maybe we do want them. I need more info to know.
3. Simply reformat that one link as a quick fix.
4. Consider reworking the text so that the link doesn't appear in the summary. I haven't actually looked at it so this may or may not make sense to do.

But any further discussion should probably happen elsewhere.

PR #1498, opened 2024-12-23
Yaml can be indented only two spaces, so we need to account for that

jonathan-s:
When you're using yaml for the metadata part and you've got something like a list that is only indented two spaces (which also is valid yaml). In a case like this, that extra data is omitted in the final metadata. So this PR aims to fix that particular issue.

### DIFF
--- a/markdown/extensions/meta.py
+++ b/markdown/extensions/meta.py
@@ -21,17 +21,18 @@ for details.
 
 from __future__ import annotations
 
-from . import Extension
-from ..preprocessors import Preprocessor
-import re
 import logging
+import re
 from typing import Any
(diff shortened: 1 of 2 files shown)


waylan:
This extension does **NOT** parse YAML. Yes, it uses YAML style delimiters to mark the start and/or end of your meta-data, but the content of the meta-data is not YAML, nor is it trying to be (despite appearing on the surface to be similar to YAML). It is its own separate format. If you want support for YAML meta-data, then I suggest something like https://github.com/waylan/docdata.

jonathan-s:
Thanks for the heads up :).

PR #1503, opened 2025-01-24
Adjust `md_in_html` "markdown" blocks to process content consistently

### LINKED ISSUE #1074: Nested one-liners broken with md_in_html
Version 3.3.3. I've found several anomalies in the behavior of nested markdown=1 blocks. Here's one that's probably diagnostic:

```
markdown.markdown('<div class="outer" markdown="block"><div class="inner" markdown="block">*foo*</div></div>', extensions=["extra"])
```

returns:

```
'<p><div class="inner" markdown="block"><div class="outer" markdown="block">*foo*</div></p>\n</div>'
```

The inner `<div>` is now outside.

Other weird things in this example are the addition of the `<p>` tags, which are unmatched, and a stray `\n` .

### LINKED ISSUE #1075: <input> is not escaped inside inline code inside <div>
I wrote something like this:

```
<div class="outer" markdown="1">

Code: `<label><input/></label>`

</div>
```

With stable Markdown 3.3.3 I get the expected result:

```
>>> print(markdown.markdown('<div class="outer" markdown="1">\n\nCode: `<label><input/></label>`\n\n</div>', extensions=["extra"]))

<div class="outer">
<p>Code: <code>&lt;label&gt;&lt;input/&gt;&lt;/label&gt;</code></p>
</div>
```

But I installed latest git version to solve #1066 and I got mysterious result:

```
>>> print(markdown.markdown('<div class="outer" markdown="1">\n\nCode: `<label><input/></label>`\n\n</div>', extensions=["extra"]))

<div class="outer">
<p>Code: <code>&lt;label&gt;<input/>&lt;/label&gt;</code></p>
</div>
```

label is escaped properly, but input is not escaped, it's strange

### LINKED ISSUE #1502: md_in_html can process content out of order if block HTML is nested under it
I have Block plugins that expect a start `/// start` and end `///`. In almost all cases, it seems it is a reasonable assumption that Python Markdown will process the start before the end allowing the plugin to work.

In a case like below, this holds true. I add some logging to show when the processed lines are encountered.

```html
/// 1

<p>
Hello i'm a 'p' in a tab which in in a div!
</p>

/// 2
```

Order logged:

```
/// 1
/// 2
```

A problem is introduced though when `md_in_html` is used.

```html
<div class="my-div" markdown>

/// 1

<p>
Hello i'm a 'p' in a tab which in in a div!
</p>

/// 2

</div>
```

Order logged:

```
/// 2
/// 1
```

As we can see the order of processing is now out of order, and since our extension relies on this order, the extension also breaks.

If the wrapping tag is inline, there is no problem. If the nested tag is inline, there is also no problem. So this is only when block tags are used within a block tag being processed by `md_in_html`.

facelessuser:
Ensure `md_in_html` processes content inside a "markdown" block the same way content is processed outside of a "markdown" block.

- Flatten the HTML content into placeholders so that the parser will treat the "markdown" block content in the same way it does when `md_in_html` is not enabled. The placeholders are expanded once the parser reaches them in a linear fashion. This allows extensions to deal with HTML content and consume it the same way it deals with them with them when the content is not nested under a "markdown" block.

- Instead of content being processed in dummy tags, content is now processed under the real parent allowing extensions to have better context to make better decisions.

Additionally, fix some issues with tags and inline code.

Also, fix some issues with one-liner block tags, e.g. `<tag><tag>...`

Resolves #1502 
Resolves #1075
Resolves #1074

### DIFF
--- a/.spell-dict
+++ b/.spell-dict
@@ -178,7 +178,7 @@ plugins
 configs
 pre
 formatters
-
+unflattened
 dedented
 Setext
 unindented
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,10 +17,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Backslash Unescape IDs set via `attr_list` on `toc` (#1493).
-* `md_in_html` should now process content inside "markdown" blocks a similar way
+* `md_in_html` will process content inside "markdown" blocks a similar way
   as they are parsed outside of "markdown" blocks giving a more consistent
   expectation to external extensions (#1503).
-* `md_in_html` should not handle tags within inline code blocks better (#1075).
+* `md_in_html` handle tags within inline code blocks better (#1075).
+* `md_in_html` fix handling of one-liner block HTML handling (#1074)
 
 ## [3.7] -- 2024-08-16
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,6 +17,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 * Backslash Unescape IDs set via `attr_list` on `toc` (#1493).
+* `md_in_html` should now process content inside "markdown" blocks a similar way
+  as they are parsed outside of "markdown" blocks giving a more consistent
+  expectation to external extensions (#1503).
 
 ## [3.7] -- 2024-08-16
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -20,6 +20,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * `md_in_html` should now process content inside "markdown" blocks a similar way
   as they are parsed outside of "markdown" blocks giving a more consistent
   expectation to external extensions (#1503).
+* `md_in_html` should not handle tags within inline code blocks better (#1075).
 
 ## [3.7] -- 2024-08-16
 
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -179,8 +179,9 @@ class HTMLExtractorExtra(HTMLExtractor):
                             child.tail = None
                             state = child.attrib.get('markdown', 'off')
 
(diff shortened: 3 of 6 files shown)


facelessuser:
I'll update changelog once I know if this change would like to be considered as a a bug or a feature. More importantly, I need to make sure this change is generally desired.

facelessuser:
I think this is ready for review now.

facelessuser:
I decided to just add the change as a "fix" for now in the changelog. It fixes surprising behaviors and doesn't change any existing, tested functionality. If we wish to reclassify it, I can on request.

facelessuser:
I'm actually going to leave this in draft a little longer to test some more corner cases.

facelessuser:
I think I'll be able to fix #1075 while I'm in here, or at the very least improve results. I'm going to add more tests as I validate cases that are not covered in tests.

facelessuser:
I've also fixed some issues with #1074. Working in the HTML parser isn't my idea of fun, so I just wanted to resolve the issues while I'm deep in the code as I'm unlikely to jump into it for a while.

Again, existing tests are still passing and in addition, new tests check for the new inline block tag behavior.

These changes really make `md_in_html` feel more natural and they bring back probably some lost functionality from before the HTML parser rewrite.

I think I'm almost done validating edge cases and such.

facelessuser:
@waylan I think I've done all I can do on this. Whenever you get a chance, please take a look, give it a try, and let me know what you think. If there are some cases you think I'm missing, or you see some things you don't like, let me know.

waylan:
I do not have the bandwidth to review this at this time and don't know when I will. @facelessuser I trust you to proceed as you see fit.

facelessuser:
@waylan Thanks for letting me know. It seems you are generally not against the idea of what I've proposed, so with that in mind, I will spend some time going over it once again with a critical eye and merge if nothing is found. I'm more than happy to support any surprises post-release.

facelessuser:
@waylan I'll need either a review approval or permissions to override review when merging if I am to proceed.

waylan (approved):
I haven't actually reviewed this, but am approving so that @facelessuser can move forward as he sees fit.

Issue #875, opened 2019-11-05
Wikilinks: link-transform


klml:
I would like to have an option to change the link-transform in [WikiLinks](https://python-markdown.github.io/extensions/wikilinks/).

Currently the link is created according to the text notation: `[[Wiki Link]]` produces `href="/Wiki_Link/"`

I would like to have the possibility to lowercase the link: `[[Wiki Link]]` should produce `href="/wiki_link/"` (and for the sake of completeness: uppercase;)

Is this in the scope of Python-Markdown? So I would create an PR.

waylan:
If I make any changes to this extension is would be to remove it from being included with Markdown. It only remains as it is very old. I suspect a better approach for you would be to either fork it or create your own new extension which offers the behavior you desire.

klml:
In any case somebody else needs this feature, I added it in [com/neurobin/mdx_wikilink_plus](https://github.com/neurobin/mdx_wikilink_plus#configuration) as `url_case`.

PR #1504, opened 2025-01-25
WikiLinks extension: Allow literal dot in wikilink pattern

rdmark:
Having a literal dot in a wikilink is arguably a common use case, and allowing this library to recognize such patters would broaden its utility.

For context: I'm trying to use Python-Markdown to generate online documentation for a software package, and as part of that I'm transcoding roff man pages to html for publishing on the website. A roff man page typically has names such as `ad.1` or `afp.conf.5` and I would like to be able to create wikilinks to them.

### DIFF
--- a/markdown/extensions/wikilinks.py
+++ b/markdown/extensions/wikilinks.py
@@ -51,7 +51,7 @@ class WikiLinkExtension(Extension):
         self.md = md
 
         # append to end of inline patterns
-        WIKILINK_RE = r'\[\[([\w0-9_ -]+)\]\]'
+        WIKILINK_RE = r'\[\[([\w0-9_ -\.]+)\]\]'
         wikilinkPattern = WikiLinksInlineProcessor(WIKILINK_RE, self.getConfigs())
         wikilinkPattern.md = md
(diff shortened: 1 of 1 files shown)


markdown/extensions/wikilinks.py
```diff
@@ -51,7 +51,7 @@ def extendMarkdown(self, md):
         self.md = md
 
         # append to end of inline patterns
-        WIKILINK_RE = r'\[\[([\w0-9_ -]+)\]\]'
+        WIKILINK_RE = r'\[\[([\w0-9_ -\.]+)\]\]'
```
facelessuser, line 54: Regardless of whether this is accepted or not, `-` should be at the end so it is not interpreted as specifying a range of characters between space and dot.

rdmark: That's a good suggestion, thanks.

rdmark:
As per discussion in https://github.com/Python-Markdown/markdown/issues/875 (which is for essentially the same kind of code change) a PR like this won't be accepted.

I have found a different way to solve my problem, so let me close this for now. Besides, a more holistic look at what makes a valid wikilink would be required.

PR #1505, opened 2025-01-28
Remove PHP tests and create tests for cases that are useful to keep

### LINKED ISSUE #1480: GPL-2.0 in test tests/php
Hi, 
thank you very much for providing this package.
I am currentliy doing a license analysis of markdown and came across the GPL-2.0 license text in [tests/php](https://github.com/Python-Markdown/markdown/blob/fc74d69758ff288e26ff616c703b22ce5a02a68a/tests/php/License).
How is the relation of the GPL-2.0 to the files in this repo, does it apply to any of the files, if yes it would be very useful that those files contain a reference to the GPL-2.0 otherwise no one can determine to which files it applies and so it will be very hard to comply to the license.

facelessuser:
Resolves #1480


waylan (approved):
Thanks for doing this.

PR #1506, opened 2025-01-31
Treat center like a block-level element

### LINKED ISSUE #1481: InlineHTML <center> not treated as HTML correctly
If `<center>` has tags nested underneath it, the `<center>` tag is surrounded by `<p>` tags, breaking the HTML.

```py
from markdown import markdown
text = """# My Header

<center>
Test
</center>

<center>
<div>Something</div>
</center>

<div>
Test div
</div>

<div>
<p>Some paragraph</p>
</div>

<div>
<article>Some paragraph</article>
</div>

Something else
"""
print(markdown(text))
```
Results in:

image

It seems like <div> works correctly, so not 100% why the difference.

facelessuser:
Fixes #1481

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,7 +24,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   expectation to external extensions (#1503).
 * `md_in_html` handle tags within inline code blocks better (#1075).
 * `md_in_html` fix handling of one-liner block HTML handling (#1074).
-* Ensure `<center>` is treated like a block-level element.
+* Ensure `<center>` is treated like a block-level element (#1481).
 
 ## [3.7] -- 2024-08-16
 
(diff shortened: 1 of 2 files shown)


waylan (approved):

rnestler:
I think this change broke some pages on my pelican based blog which had the following markdown
```md
<center>
![foo](image.png)
</center>
```

It changed rendering like this:
```diff
-<p><center>
-<img alt="foo" src="image.png">
-</center></p>
+<center>
+![foo](image.png)
+</center>
```

So the image inside the `<center>` tag doesn't get converted anymore.

facelessuser:
Yes, center is not an inline element, it is a block element. Python Markdown does not parse block element content by default. You would need to enable the `md_in_html` extension and then do the following

```
<center markdown>
![foo](image.png)
</center>
```

Remember, the `<center>` element is deprecated and considered obsolete in HTML5.

rnestler:
> Yes, center is not an inline element, it is a block element. Python Markdown does not parse block element content by default. You would need to enable the `md_in_html` extension and then do the following

Good to know.

> Remember, the `<center>` element is deprecated and considered obsolete in HTML5.

Yes I'm aware, and now I also fixed up these old blog posts to not use it, but a css class.

PR #1510, opened 2025-02-21
Optimize raw HTML post-processor

### LINKED ISSUE #1507: Optimizing the raw HTML post-processor
Locally, I was able to shave 20 seconds off a MkDocs build (from ~50s to ~30s) by optimizing a bit the raw HTML post processor. Would you be open to a PR that optimizes the post-processor code a bit? Any guidance on how to provide benchmarks and results to confirm to performance gain?

pawamoy:
Closes #1507 

TODO

- [x] write a second batch of tests for cases where the list contains duplicates
- [x] ~~maybe return new instances of `_BlockLevelElements` instead of `list`/`set` in relevant methods~~
- [x] ~~maybe use `stacklevel=2`~~
- [x] ~~maybe use `warnings.deprecated` (and `typing_extensions.deprecated` on Python 3.12-)~~ use existing `deprecated` function instead of `warnings.warn`


pawamoy:
Hmm, the list->set change could be seen as breaking. We can instead create a new `Markdown._block_level_elements` attribute, and use that in `isblocklevel()`. Let me know if you think that's best.

pawamoy:
Flake8 warnings fixed in https://github.com/Python-Markdown/markdown/pull/1511.

markdown/postprocessors.py
```diff
@@ -73,37 +72,26 @@ class RawHtmlPostprocessor(Postprocessor):
...
-
-            return replacements[key]
-
-        if replacements:
+            if key := m.group(1):
+                wrapped = True
+            else:
+                key = m.group(2)
+                wrapped = False
+            if (key := int(key)) >= len(self.md.htmlStash.rawHtmlBlocks):
```
pawamoy: Could use `html_counter` instead:

```suggestion
            if (key := int(key)) >= self.md.htmlStash.html_counter:
```

markdown/postprocessors.py
```diff
@@ -73,37 +72,26 @@ class RawHtmlPostprocessor(Postprocessor):
...
+                return m.group(0)
+            html = self.stash_to_string(self.md.htmlStash.rawHtmlBlocks[key])
+            if self.isblocklevel(html) or not wrapped:
+                return pattern.sub(substitute_match, html)
+            return pattern.sub(substitute_match, f"<p>{html}</p>")
+
+        if self.md.htmlStash.html_counter:
```
pawamoy, line 88: Or we could *not* use `html_counter` and only rely on the actual list, `rawHtmlBlocks`:

```suggestion
        if self.md.htmlStash.rawHtmlBlocks:
```

markdown/postprocessors.py
```diff
@@ -73,37 +72,26 @@ class RawHtmlPostprocessor(Postprocessor):
...
-        if replacements:
+            if key := m.group(1):
+                wrapped = True
+            else:
+                key = m.group(2)
+                wrapped = False
+            if (key := int(key)) >= len(self.md.htmlStash.rawHtmlBlocks):
+                return m.group(0)
+            html = self.stash_to_string(self.md.htmlStash.rawHtmlBlocks[key])
+            if self.isblocklevel(html) or not wrapped:
```
pawamoy: Micro-optimization (given the list->set change is applied): make `isblocklevel` check lazy.

```suggestion
            if not wrapped or self.isblocklevel(html):
```

waylan: Yes, I noticed this immediately. This is a good change. Any time we can avoid calling `isblocklevel` we should.

markdown/postprocessors.py
waylan: I don't see any reason to recompute the length when we already have it stored.

What's not clear to me is why we are reassigning `key` here. I understand the need to convert a string to an integer, but why do that here? Why not when it is initially assigned on lines 76 and/or 79?

markdown/postprocessors.py
waylan: I was going to say let's be consistent and use the `html_counter` everywhere. But then I noticed that we don't actually document that attribute of the [class instance](https://python-markdown.github.io/reference/markdown/util/#markdown.util.HtmlStash). Although, there is no good reason to not have it documented (to be clear, I am not suggesting this PR should add the documentation; just making a general observation about the state of the code). And in this case, I don't think there is any performance advantage to either check. I'm indifferent to this.

waylan (changes requested):
This looks good generally. However, I haven't tested it or done any performance comparisons of my own. From a quick reading of the code only the few things you pointed out stand out to me. I'll need to take a closer look when I have more time.

facelessuser:
I'll likely try this branch on some projects to inspect for regressions, just to be sure

waylan:
> Hmm, the list->set change could be seen as breaking. We can instead create a new `Markdown._block_level_elements` attribute, and use that in `isblocklevel()`. Let me know if you think that's best.

This is a valid concern. I'm not sure what to think about it.

facelessuser:
Yes, I saw the change of blocks using a list to set and meant to comment on it. I'm pretty sure it'll break some of my plugins, so keeping the original but deprecating it in favor of a new structure is probably the way to go.

markdown/postprocessors.py
pawamoy: Oh, just an oversight, we can definitely coerce to int when first assigning.

pawamoy: Ah I remember now, that's because I'm doing `if key := m.group(1)`, and casting to `int` here would fail when `m.group(1)` is `None`.

markdown/postprocessors.py
pawamoy: Let's be consistent and use `html_counter`.

pawamoy:
Thanks for confirming @facelessuser! I'll change it.

pawamoy:
I just added a new private `_block_level_elements` attribute for now, but we should probably deprecate the old one and make the new one public. I'll wait for your input on this @waylan.

markdown/core.py
```diff
@@ -114,6 +114,7 @@ def __init__(self, **kwargs):
         """ List of characters which get the backslash escape treatment. """
 
         self.block_level_elements: list[str] = BLOCK_LEVEL_ELEMENTS.copy()
+        self._block_level_elements: set[str] = set(BLOCK_LEVEL_ELEMENTS)
```
facelessuser: I would prefer a non-private version of the sets as I imagine we can and should deprecate the old one and eventually remove it in the future. I'm likely to begin referencing the new one as soon as I can, I just don't want it to break things when it releases.

waylan: I agree. Lets make it public.

waylan: There is one other concern. Some extensions are likely to alter the list (by adding or removing elements). Therefore, the new attribute could get out-of-sync with the pre-existing attribute. How should we address that?

One possible solution is to create a custom class which subclasses `set` and also provides the common `list` methods (which should issue a deprecation warning when called). Then there are not two sequences to keep in sync and both the old and new APIs continue to work. This also removes the need to come up with a new attribute name on the `Markdown` class. At some future point the `list` methods could instead raise an error when called. Then at a later point, the custom class could be removed.

pawamoy: Sounds good! `list` has the following methods in common with `set`:

- `copy`: ~~nothing to do~~ make sure to return a copy of the current instance (not a raw set)
- `pop`: the `list` method accepts an `index` argument, and raises `IndexError` instead of `KeyError`. We can catch KeyError and re-raise IndexError (and emit a warning?) during the deprecation period. As for `index`, I think we should call set's `pop` but emit a warning that it pops an arbitrary item.
- `remove`: both accept a single positional argument, but `list` raises `ValueError` instead of `KeyError`. We can catch and re-raise again (and emit a warning?) during the deprecation period.
- `clear`: nothing to do

...and the following additional methods:

- `append`: easy, call `self.add`
- `sort`: doesn't make sense for a set, so just `pass`
- `reverse`: doesn't make sense for a set, so just `pass`
- `insert`: easy, call `self.add`
- `index`: doesn't make sense for a set, return `0`?
- `extend`: easy, call `self.update`
- `count`: 1 if value in self else 0

I'll proceed with this, let me know if you'd like to change a few things :slightly_smiling_face:

waylan: That looks good with two exceptions. 

1. A common action would be to remove items. I imagine code exists which finds the index of an object and removes or pops by index. `set.remove` and `set.pop` accept the object as an argument rather than the index. We could return the object from `index` so that it works when passed into `remove` or `pop`, but that seems weird in all other respects. Not sure how best to approach this.
2. Presumably `count` could return `self.__len__()`.

waylan: Maybe `index` should manually iterate through the set while tracking position and return the position of the matching object. Then `pop` and `remove` can check if the argument they receive is an index or not and act accordingly.

pawamoy: Both `set.remove` and `list.remove` accept the object as argument, so nothing much to do here. About `pop`: interesting suggestion. So `index(value)` would actually return `value` (if present), and `pop` would receive it to discard the element. I'll do that.

Also, what about subscript syntax, like `block_level_elements[0]`?

pawamoy: If we want to preserve order (to make index based API work), we could use a dict instead of a set, internally.

waylan: > Also, what about subscript syntax, like `block_level_elements[0]`?

Ah right. Someone might do `del block_level_elements[block_level_elements.index(value)]` or similar. Yeah, we probably need to address that as well. 

It seems clear that we need to have `index` return a position and then any list method (including double underscore methods) which accepts an index to be able to act on one.

pawamoy: Yep. I believe we'll have to use some kind or ordered set, or to maintain an internal list counterpart.

facelessuser: It's possible we can just wrap the list and set type so that when a change is made to the list a hook is called to update the set and vise versa. I don't think we need to build up a whole new structure.

facelessuser: I don't think we should kill ourselves to have compatibility, but make a reasonable effort to not break things

pawamoy: So, something like a plain class with two hidden structs (list and set), with all accompanying methods?

```python
class _BlockLevelElements:
    def __init__(self, elements: list[str]) -> None:
        self._list = elements.copy()
        self._set = set(self.list)

    def __contains__(self, value):
        return value in self._set
    def append(...): ...
    def add(...): ...
    ...
```

pawamoy: I pushed something in https://github.com/Python-Markdown/markdown/pull/1510/commits/d4813dc6c1064e9c06ac2543fc452a57ae73773f, let me know what you think.

markdown/util.py
```diff
@@ -45,60 +44,269 @@
...
+class _BlockLevelElements:
+    def __init__(self, elements: list[str]) -> None:
+        self._list = elements.copy()
+        self._set = set(self._list)
+
+    def __add__(self, other: list[str]) -> list[str]:
+        warnings.warn(
+            "Using block level elements as a list is deprecated, use it as a set instead.",
+            DeprecationWarning,
+        )
```
pawamoy: I emit the same warning every time, so that users only get it once. The message goes straight to the point: don't use it as a list.

markdown/util.py
```diff
@@ -45,60 +44,269 @@
...
+        # Using `+` means user expects a list back.
+        return self._list + other
```
pawamoy: Methods with no side-effects return a `list` or `set` directly, not a copy of our hybrid instance. The reason is we assume that as soon as a user explicitly uses it as a list (or set), they will only use it as a list (or set), so no need to continue providing both interfaces.

markdown/util.py
```diff
@@ -45,60 +44,269 @@
...
+
+    def __and__(self, other: set[str]) -> set[str]:
+        # Using `&` means user expects a set back.
+        return self._set & other
+
+    def __contains__(self, item):
+        return item in self._set
```
pawamoy: Our initial performance optimization is right here :joy:

markdown/util.py
```diff
@@ -45,60 +44,269 @@
...
+        # In-place union should update both list and set.
+        self._set.update(*others)
+        for other in others:
+            for element in other:
+                if element not in self._set:
+                    self._list.append(element)
+
+
+# Type it as `set[str]` to express our intent for it to be used as such.
+BLOCK_LEVEL_ELEMENTS: set[str] = _BlockLevelElements([
```
pawamoy: I can't guarantee I didn't make subtle mistakes, so it should probably need proper testing.

If this all seems too complicated, happy to try something simpler (it was implemented quickly, no problem with discarding all this :smile:).

markdown/util.py
waylan: Not necessarily. The same object is likely to be  accessed by multiple extensions. It is reasonable to assume that some newly updated extensions will assume a set and some older (not yet updated) extensions will assume a list. Whatever solution we chose needs to continue to work with both throughout the life of the `Markdown` class instance.

markdown/core.py
facelessuser: @pawamoy I think something like that may work, it needs to be tested well though. I saw in your link that things like `__add__` only add to the list but don't add to the set.

facelessuser: Never mind, it is creating a new object and you are returning a pure list but providing a warning. Things may break if they set it back, but I understand what's going on.

markdown/util.py
pawamoy: Yeah I thought about that, though since these methods return a new object (that's what I meant with "no side-effect", which wasn't super clear), I figured there are very few chances multiple extensions would actually share references to such newly created lists/sets (each new list/set would probably be isolated to the extension that created it). Anyway, I can definitely change that to return a `_BlockLevelElements` instance, to be more robust :slightly_smiling_face:

markdown/core.py
pawamoy: Yeah I figured new lists/sets created this way would likely *not* be shared across multiple extensions, but assigning it back would create such a scenario. That's also something @waylan [remarked](https://github.com/Python-Markdown/markdown/pull/1510#discussion_r1973703365), so to be more robust we can just create a new `_BlockLevelElements` instance :+1:

markdown/util.py
facelessuser: It should definitely get tested. Eventually it'll go away as well. There's probably ways in which people can still mess things up, but I think this tries to provide a reasonable attempt to keep stuff from breaking. I'm open to other ideas if someone has them as well.

pawamoy: Alright, I'll start writing tests :slightly_smiling_face: Just wanted initial feedback before tackling them :blush:

markdown/util.py
```diff
@@ -44,7 +44,269 @@
 """
 
 
-BLOCK_LEVEL_ELEMENTS: list[str] = [
+# TODO: Raise errors from list methods in the future.
+# Later, remove this class entirely and use a regular set.
+class _BlockLevelElements:
```
waylan: Elsewhere the type is specified as `set`. However, as this is not a subclass of set, that would be incorrect. Also, later when we remove this, then we will have a `set`. Perhaps this class should properly reflect that now. Unfortunately, there are no specialized container datatypes for sets in [collections](https://docs.python.org/3/library/collections.html). And sets are not a [Sequence Type](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range) either. Not really sure how best to address this.

markdown/util.py
waylan: Sorry, I missed that this comment was in a specific method which returns a modified object. Yeah, in this case, returning the same object type makes sense.

markdown/util.py
pawamoy: My feeling was that it was OK to lie about the type, so that users actually get type warnings when they use it as a list instead of a set. That's one more (very effective) way of communicating the change. And since we implement all set functionality, the lie is not too big :slightly_smiling_face:

markdown/util.py
pawamoy: No worries, I wasn't very clear about it. @facelessuser [expressed](https://github.com/Python-Markdown/markdown/pull/1510#discussion_r1973856848) the same concern, so we might want to return a new instance of `_BlockLevelElements`, just to be safe. Let me know what you prefer.

pawamoy (commented):
First batch of tests added. It's pretty basic. A second batch should be added to test lists with duplicate elements within them. I updated the code to try and reduce the possibility of duplicates but it's not thorough.

markdown/util.py
```diff
@@ -44,7 +44,269 @@
...
+
+    def intersection(self, *others: set[str]) -> set[str]:
+        # User expects a set back.
+        return self._set.intersection(*others)
+
+    def intersection_update(self, *others: set[str]) -> None:
+        # In-place intersection should update both list and set.
+        self._set.intersection_update(*others)
+        self._list.clear()
+        self._list.extend(sorted(self._set))
```
pawamoy: I took a shortcut, but we should probably maintain the list order, while preventing duplicates.

markdown/util.py
```diff
@@ -44,7 +44,269 @@
...
+
+    def symmetric_difference(self, other: set[str], /) -> set[str]:
+        # User expects a set back.
+        return self._set.symmetric_difference(other)
+
+    def symmetric_difference_update(self, other: set[str], /) -> None:
+        # In-place symmetric difference should update both list and set.
+        self._set.symmetric_difference_update(other)
+        self._list.clear()
+        self._list.extend(sorted(self._set))
```
pawamoy: Same shortcut, though we should maintain list order by first removing then adding elements, preventing duplicates, probably.

markdown/util.py
```diff
@@ -44,7 +44,269 @@
...
+
+    def union(self, *others: set[str]) -> set[str]:
+        # User expects a set back.
+        return self._set.union(*others)
+
+    def update(self, *others: set[str]) -> None:
+        # In-place union should update both list and set.
+        self._set.update(*others)
+        self._list.clear()
+        self._list.extend(sorted(self._set))
```
pawamoy: Shortcut again, we should maintain order and only append elements, preventing duplicates.

markdown/extensions/md_in_html.py
```diff
@@ -42,7 +42,7 @@ class HTMLExtractorExtra(HTMLExtractor):
 
     def __init__(self, md: Markdown, *args, **kwargs):
         # All block-level tags.
-        self.block_level_tags = set(md.block_level_elements.copy())
+        self.block_level_tags = md.block_level_elements.copy()
```
waylan: I just noticed this, but this was already being converted to a set in this extension. Maybe we should keep that behavior rather than leave it as our custom type.

markdown/util.py
waylan: We could do that, but some users run type validation on their code and this could result in bug reports. Either we need to fully document our reasoning for the inconsistency in the code comments or we need to make a change to the code.

tests/test_block_level_elements.py
```diff
@@ -0,0 +1,328 @@
...
+Copyright 2007-2023 The Python Markdown Project (v. 1.7 and later)
+Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)
+Copyright 2004 Manfred Stienstra (the original version)
+
+License: BSD (see LICENSE.md for details).
+
+Tests for block level elements.
+===============================
+
+Tests specific to the hybrid list/set container for block level elements.
```
waylan: Let's add a comment here outlining the deprecation path for this file, specifically noting when it can be removed. We should add a similar comment on the `markdown.util._BlockLevelElements` class as well which reminds us to also remove these tests when the class is removed from the code.

markdown/extensions/md_in_html.py
pawamoy: Right, I removed `set()` around the copy because it was redundant when our block level elements were a plain set, but now it's a hybrid container with a custom `copy` method we should wrap again :+1:

markdown/util.py
pawamoy: Users running type checkers on their code is exactly what I expect, yes :smile: So, my opinion is, yes, lets keep lying about the type, and document that prominently (both with code comments and release/changelog notes).

---

Type correctness would otherwise mean that:

- we type the variables as `_BlockLevelElements`
- which in turn means we'd have to document this class properly
- which in turn (probably) means we'd have to make it public

(also we'd likely want to use `Self` from `typing_extensions` as return annotation of methods retuning `self`, which means an additional dependency depending on the Python version)

I know it's far from being exhaustive but a [search on GitHub](https://github.com/search?q=md.block_level_elements+language%3APython+-user%3APython-Markdown&type=code&p=1) returns only **one** result for use of `md.block_level_elements` (apart from @facelessuser's pymdownx which [creates a new `set` from it](https://github.com/facelessuser/pymdown-extensions/blob/4d6b5e6249a8c9f8fe1444c453b29d313f65e3d5/pymdownx/blocks/__init__.py#L162), and discarding forks and venvs), [and it's using `remove`](https://github.com/bcgov/developer-experience-team/blob/ee78edc73b4d384ea700ad1134e475cf13d8068f/markdown-extensions/wiki-janitor/wikijanitor.py#L7), which is both a list and set method. That's IMO a strong indicator that `md.block_level_elements` has very few uses in the wild, and so immediately advertising its type as `set` should have a very small impact. I could be wrong though :shrug:

markdown/extensions/md_in_html.py
pawamoy: Done. I removed the call to `.copy()` since calling `set` on the object creates a new object anyway, and the container is "single-level" (only strings, no references to other mutable structs).

tests/test_block_level_elements.py
pawamoy: Done, feel free to rephrase :slightly_smiling_face:

markdown/core.py
pawamoy: Initial conversation is resolved I believe, marking as such.

markdown/postprocessors.py
pawamoy: Let me know if you have another idea to make this more elegant, I'll mark as resolved for now :slightly_smiling_face:

pawamoy:
About duplicates: I don't prevent their insertion, otherwise we lose backwards compatibility again. We'll just have to add test cases with duplicates :slightly_smiling_face: 

A few other remarks/suggestions:

- we could use [`warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) to decorate deprecated methods (instead of calling `warnings.warn` ourselves), so that type checkers will warn about their use, and IDEs will strike them through (visual indication of deprecation). For Python less than 3.13 we'd have to depend on typing-extensions.
- in any case, we could pass `stacklevel=2` to warnings, so that they show the user line calling these deprecated methods.
- I've developed a small tool called [Yore](https://github.com/pawamoy/yore) that aims at helping manage legacy code. Instead of adding various `# TODO` comments throughout the code, you instead add `# YORE` structured comments and can run `yore check` to check if legacy code should be removed/transformed (Python version just reached its end of life / you're ready to bump the major version) and `yore fix` to automatically apply prepared transforms. Just sharing, can help setting it up if you're interested :smile: (it's just a matter of adding the dev-dep and maybe calling it in tox/CI).

pawamoy:
I added tests with duplicates.

markdown/util.py
```diff
@@ -44,7 +44,298 @@
...
+        while True:
+            try:
+                self._list.remove(element)
+            except ValueError:
+                break
+        # We raise `ValueError` for backwards compatibility.
+        try:
+            self._set.remove(element)
+        except KeyError:
+            raise ValueError(f"{element!r} not in list") from None
```
pawamoy: Not sure about this. It's difficult to deprecate a raised exception in favor of another one. Maybe we should immediately start raising `KeyError`.

pawamoy: About deprecating raised exceptions: we could create new exception that inherits from both `ValueError` and `KeyError`, so that both can be caught in user code. I was hoping we could hook into the `except` mechanism to only emit a warning when `ValueError` or `KeyError` is caught, but it's not possible unfortunately, see https://discuss.python.org/t/instance-subclass-check-for-except-clauses/71583. So we're stuck with either:

- not emitting any warning
- always emitting a warning, that users will have to ignore even if they catch the right exception

waylan:
Sorry, but I was out for a couple of weeks. Back now and trying to catch up.

> * we could use [`warnings.deprecated`](https://docs.python.org/3/library/warnings.html#warnings.deprecated) to decorate deprecated methods (instead of calling `warnings.warn` ourselves

Yes, let's do that. I would have mentioned this earlier but missed that we weren't using `warn.deprecated` already.

>  we could pass `stacklevel=2` to warnings

We have not traditionally done this in the past. I have always taken the position that we are not responsible for Python's default behavior. As Python by default hides deprecation warnings, then users should expect that and act accordingly. But I'm not opposed to this either.

> * I've developed a small tool called [Yore](https://github.com/pawamoy/yore) that aims at helping manage legacy code. Instead of adding various `# TODO` comments throughout the code, you instead add `# YORE` structured comments and can run `yore check` 

It is easy enough to search for #TODO. Adding dependencies just creates more complexity.

markdown/util.py
waylan: I'm inclined to suggest that we not bother ourselves with the errors here. We can just call `_list.remove` and `_set.remove` and allow any errors to bubble up to the calling code. Is there any reason why we shouldn't do that?

markdown/util.py
waylan: It's not just type validators that I'm thinking about here. What about someone who is using `isinstance` (or similar) on the object in their code? They could reasonably expect `isinstance(set)` to return True, which is does not. For that matter existing code could reasonable expect `isisntance(list)` to return True. My objection is that neither return True. However, if `_BlockLevelElements` was a subclass of something, then it could return True for one of either `set` or `list`. That's why I pointed out above that set is not a [Sequence Type](https://docs.python.org/3/library/stdtypes.html#sequence-types-list-tuple-range). I was hoping it was and then `_BlockLevelElements` could just subclass the relevant baseclass and return True for both. I think we need to return True for at least one. The question is: which one? Using `set` is a breaking change but what it will end up being when the transition is complete. Using `list` avoids the breaking change, but it not forward looking. And I don't think there is any way to raise a deprecation warning if a user calls `isinstance` on the objection.

My point is that I think `_BlockLevelElements` needs to be either a set or a list at a minimum. Ideally it would be both, but that is probably not sensible. Any solution is going to be a compromise. I'm saying that having `_BlockLevelElements` not be either a set or a list is not the compromise I want to make.

markdown/util.py
waylan: I think this is fine as-is. The warning will alert the user that they need to change their code, but in the meantime, we are returning an object of the type that they expect.

markdown/util.py
pawamoy: Good point. I think we could inherit from both `list` and `set` since we implement all their methods anyway? This way `isinstance` would return `True` for both. Also, we can customize the behavior of `isinstance` thanks to [`__instancecheck__`](https://docs.python.org/3.11/reference/datamodel.html#class.__instancecheck__) to emit a deprecation warning when checking against `list` :slightly_smiling_face:

markdown/util.py
pawamoy: Alright! Thanks.

markdown/util.py
pawamoy: Agree on keeping things simple. I'll try a different approach (we still want to remove *all* matching items from the list, not just one) :thinking:

pawamoy:
> Sorry, but I was out for a couple of weeks. Back now and trying to catch up.

Absolutely no worries, and no need to apologize, thank you very much for your answers :slightly_smiling_face: I'll address the various points you've answered.

pawamoy:
Oh, I noticed again that you have your own `deprecated` implementation. I'll use it instead of `warnings.deprecated` or `typing_extensions.deprecated` under 3.13, and we can later remove this custom implementation in favor of the official one, to reduce the amount of changes in this PR. I'll just have to move the function definition up, since we'll have to use it at the module-level.

markdown/util.py
pawamoy: OK `__instancecheck__` works the other way unfortunately, it's not meant for this use-case. Inheriting from both `set` and `list` is not possible: CPython doesn't allow it. Inheriting from `set` breaks a lot of tests (121 of them), not sure exactly why. Inheriting from `list` works, but as you pointed, `isinstance(ref, set)` will return `False`.

markdown/util.py
```diff
@@ -38,13 +38,283 @@
...
+
+    def pop(self, index: int | None = None, /) -> str:
+        # In-place pop should update both list and set.
+        if index is None:
+            index = -1
+        else:
+            warnings.warn(
+                "Using block level elements as a list is deprecated, use it as a set instead.",
+                DeprecationWarning,
+            )
```
pawamoy: We still use `warnings.warn` here because we don't want to deprecate the whole method, only the `index` parameter.

markdown/util.py
waylan: > Inheriting from both `set` and `list` is not possible: CPython doesn't allow it.

That is what I thought. Not sure why, but I had a vague recollection that that was the case.

> Inheriting from `set` breaks a lot of tests (121 of them), not sure exactly why.

Do these tests fail if we use `set` directly rather than out custom class? If so, we really need to track this down. We don't want to encounter this issue for the first time in a future step of the deprecation...

That is if we ever make the deprecation. I'm thinking that this may be a blocker for this change moving forward. Can you provide some clear date on the improvements this one change makes?

pawamoy: > Do these tests fail if we use set directly rather than out custom class? If so, we really need to track this down. We don't want to encounter this issue for the first time in a future step of the deprecation...

Tests pass when block level elements variable are sets. I ran a debugging session and it turns out `set(md.block_level_elements)` returns an empty set for some reason (while `list(md.block_level_elements)` returns a valid list with all elements). I *think* it's because by inheriting from `set`, `_BlockLevelElements` indeed becomes a `set`, **but an empty one**, since we store elements in `self._set` (and `self._list`). CPython probably then takes a shortcut and doesn't use `__iter__`. I believe we can solve this by getting rid of `self._set` and calling parent methods instead of `self._set.METHOD`.

> That is if we ever make the deprecation. I'm thinking that this may be a blocker for this change moving forward. Can you provide some clear date on the improvements this one change makes?

Yep, totally understandable, 700 lines for changing a variable from `list` to `set` is not the most elegant backward compatible change. I don't have clear data, but I can tell you this:

- without the [post-processor change](https://github.com/Python-Markdown/markdown/pull/1510/commits/d1eadc3a943ac8abbfb6217d11857fcc8d9a47c6), the list->set would have been very important
- but now that the post-processor is much more efficient, the list->set change brings a very, very tiny perf gain (at least in the context of the raw HTML post-processor)

To be honest I wouldn't mind if we dropped the list->set change entirely, to make this PR easier to merge. We can always revisit later.

waylan: > * but now that the post-processor is much more efficient, the list->set change brings a very, very tiny perf gain (at least in the context of the raw HTML post-processor)

That is pretty significant.

> To be honest I wouldn't mind if we dropped the list->set change entirely, to make this PR easier to merge. We can always revisit later.

I agree. Let's do that.

pawamoy: I squashed all fixups, and added a revert commit, so that we keep a trace of it in this PR (while you just have to squash when merging to get rid of both).

waylan (approved):
That greatly simplifies this PR. Looks good to me. @facelessuser could you take another look at this. I think it is probably ready to go but would like another set of eyes on it.

facelessuser:
@waylan I'll take a look this evening and update with my review.

facelessuser (approved):
LGTM

pawamoy:
Thank you both for your patience and reviews :pray:

ofek:
When might this be released?

facelessuser:
@waylan do you have a timeline for the next release?

waylan:
There's nothing holding back a release other than my time. I expect a release will be made before the end of the week.

PR #1511, opened 2025-02-21
Fix flake8 warnings about whitespace before/after f-string backets

pawamoy:


waylan:
Hmm, this was an intentional style choice. I find it easier to read with the whitespace. I'm assuming this is a new addition to flake8. Can you point me to the warning and/or the documentation? We may want to disable this new rule and/or enforce the opposite behavior.

pawamoy:
They show up as E201 and E202:

```console
% uvx tox -e flake8
flake8: install_deps> python -I -m pip install flake8
flake8: commands[0]> flake8 /media/data/dev/markdown/markdown /media/data/dev/markdown/tests
/media/data/dev/markdown/markdown/extensions/abbr.py:124:39: E201 whitespace after '{'
/media/data/dev/markdown/markdown/extensions/abbr.py:124:85: E202 whitespace before '}'
/media/data/dev/markdown/markdown/postprocessors.py:89:34: E201 whitespace after '{'
/media/data/dev/markdown/markdown/postprocessors.py:89:58: E202 whitespace before '}'
/media/data/dev/markdown/markdown/postprocessors.py:97:40: E201 whitespace after '{'
/media/data/dev/markdown/markdown/postprocessors.py:97:57: E202 whitespace before '}'
/media/data/dev/markdown/markdown/postprocessors.py:97:65: E201 whitespace after '{'
/media/data/dev/markdown/markdown/postprocessors.py:97:82: E202 whitespace before '}'
flake8: exit 1 (0.50 seconds) /media/data/dev/markdown> flake8 /media/data/dev/markdown/markdown /media/data/dev/markdown/tests pid=326045
  flake8: FAIL code 1 (2.80=setup[2.30]+cmd[0.50] seconds)
  evaluation failed :( (2.86 seconds)
```

They come from pycodestyle. Strangely, only `(` and `)` are mentioned in the docs, not `{` and `}`: https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes. Can't find anything relevant in their [changelog](https://pycodestyle.pycqa.org/en/latest/developer.html#changes).

Happy to revert and ignore these rules instead :slightly_smiling_face:

pawamoy:
pycodestyle's docs are not up to date... Latest version is 2.12.1, changelog only shows 2.11.1: https://pycodestyle.pycqa.org/en/latest/developer.html#id2. This was reported here: https://github.com/PyCQA/pycodestyle/issues/1271.

pawamoy:
OK so the errors don't show up in CI because it uses Python 3.11, and locally I used 3.13. They only show up on 3.12+ apparently.

waylan:
Looks like this is the relevant changelog entry:

https://github.com/PyCQA/pycodestyle/blob/b852994a89639577d86070f513a9c186a74479b6/CHANGES.txt#L4-L10

That points to PyCQA/pycodestyle#1252 which in turn points to PyCQA/flake8#1949. However, I don't see how those directly connect to this. Unless the issue addressed there was causing f-strings to not be checked previously. Perhaps this change in behavior is an unexpected side affect.

In any event, the relevant error codes (`E201` and `E202`) are just the basic codes for any Python code in brackets or braces. If we turn those off, that turns them off for all Python code. I only want the check disabled for the delimitators of f-string replacement fields (but not the contents of the replacement field or anywhere else).

I think I'm going to sit on this for now and see how things play out.

pawamoy:
Fine by me :blush: Just note that contributors running flake8 on 3.12+ will get the warnings and might try fixing them again, so maybe you could try to enforce the Python version used for this tox env, if that's possible :slightly_smiling_face:

PR #1515, opened 2025-03-19
Add special case for closing nested quotes; fixes #1514

### LINKED ISSUE #1514: smarty: double sets of quotes not working at end
The following example renders incorrectly:
``` py
import markdown

text = '''
He replied, "She said 'Hello.'"
'''

html = markdown.markdown(text, extensions=['smarty'])
print(html)
```

The output is:
```
<p>He replied, &ldquo;She said &lsquo;Hello.&rsquo;&ldquo;</p>
```

**The `&ldquo;` at the end of this example should be `&rdquo;`.**

The quote marks were replaced in this order:
1. openingSingleQuotesRegex
2. closingSingleQuotesRegex
3. openingDoubleQuotesRegex
4. remainingDoubleQuotesRegex

Unfortunately:
* closingDoubleQuotesRegex didn't match, because the `"` is not followed by a space (it's the last character of the string, which is stripped prior to replacement).
* closingDoubleQuotesRegex2 didn't match, because of the closing single quote replacement (i.e., the intermediate data in `SubstituteTextPattern.handleMatch()` contains `\u0003` before the `"`).

One possible way to fix this issue might be to change closingDoubleQuotesRegex to `r'"(?=\s|$)'` (i.e., double quote followed by whitespace *or end of string*). However that doesn't work in general, so a special case is likely needed.

ChrisMayfield:
This PR adds two more regular expressions to the smarty extension to handle nested quote marks at the end of a sentence, such as "She said 'Hello.'" Additional unit tests are included.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -27,6 +27,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * Ensure `<center>` is treated like a block-level element (#1481).
 * Ensure that `abbr` extension respects `AtomicString` and does not process
   perceived abbreviations in these strings (#1512).
+* The `smarty` extension correctly renders nested closing quotes (#1514).
 
 ## [3.7] -- 2024-08-16
 
--- a/markdown/extensions/smarty.py
+++ b/markdown/extensions/smarty.py
@@ -134,6 +134,8 @@ doubleQuoteStartRe = r'^"(?=%s\B)' % punctClass
 #   <p>He said, "'Quoted' words in a larger quote."</p>
 doubleQuoteSetsRe = r""""'(?=\w)"""
 singleQuoteSetsRe = r"""'"(?=\w)"""
+doubleQuoteSetsRe2 = r'(?<=%s)\'"' % closeClass
+singleQuoteSetsRe2 = r"(?<=%s)\"'" % closeClass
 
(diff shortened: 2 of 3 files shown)


mitya57:
Looks good, thank you!
Can you please also add a test for https://github.com/Python-Markdown/markdown/issues/1514#issuecomment-2733881169? It seems to pass with your change:
```python
        self.assertMarkdownRenders(
            '<span>He replied, "She said \'Hello.\'"</span>',
            '<p><span>He replied, &ldquo;She said &lsquo;Hello.&rsquo;&rdquo;</span></p>'
        )
```

And please add a changelog entry to `docs/changelog.md`.

ChrisMayfield:
Yes, I thought about adding that test with `<span>` tags, but none of the existing tests have HTML in the input text, only Markdown. So I instead added a test with parentheses based on [facelessuser's comment](https://github.com/Python-Markdown/markdown/issues/1514#issuecomment-2733953325), which tests a similar situation (closing quotes before non-word character). But I would be happy to add the `<span>` test as well, if you still prefer. Just let me know.

facelessuser:
I think adding the span case is good as well. As noted before, it triggered different logic, and your fix applies to all the broken cases.

mitya57 (approved):

PR #1487, opened 2024-10-08
Add Python 3.13 and drop Python 3.8

waylan:
Yesterday (2024-10-07), Python 3.13 was released and 3.8 reached end-of-life. This adjusts Python-Markdown's meta-data and tests to match.

However, I don't expect to immediately merge this. We can wait until we know we are working on our next major release. That is, unless there are some failing tests in 3.13.

PR #1516, opened 2025-03-22
Use PEP 639 license expressions in project metadata

cdce8p:
The latest setuptools release added support for [PEP 639](https://peps.python.org/pep-0639/) license expressions.

Project metadata diff
```diff
 ...
-License: BSD 3-Clause License
-        
-        Copyright 2007, 2008 The Python Markdown Project (v. 1.7 and later)  
-        Copyright 2004, 2005, 2006 Yuri Takhteyev (v. 0.2-1.6b)  
-        Copyright 2004 Manfred Stienstra (the original version)
-        
-        Redistribution and use in source and binary forms, with or without
-        modification, are permitted provided that the following conditions are met:
-        
-        1. Redistributions of source code must retain the above copyright notice, this
-           list of conditions and the following disclaimer.
-        
-        2. Redistributions in binary form must reproduce the above copyright notice,
-           this list of conditions and the following disclaimer in the documentation
-           and/or other materials provided with the distribution.
-        
-        3. Neither the name of the copyright holder nor the names of its
-           contributors may be used to endorse or promote products derived from
-           this software without specific prior written permission.
-        
-        THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS"
-        AND ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE
-        IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
-        DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
-        FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
-        DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
-        SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
-        CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
-        OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
-        OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+License-Expression: BSD-3-Clause
 ...
-Classifier: License :: OSI Approved :: BSD License
 ...
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [build-system]
 # Minimum requirements for the build system to execute.
-requires = ["setuptools>=61.2"]
+requires = ["setuptools>=77.0"]
 build-backend = "setuptools.build_meta"
 
 [project]
@@ -17,7 +17,7 @@ maintainers = [
     {name = 'Waylan Limberg', email = 'python.markdown@gmail.com'},
     {name = 'Isaac Muse'}
 ]
-license = {file = 'LICENSE.md'}
+license = "BSD-3-Clause"
 requires-python = '>=3.9'
 dependencies = [
     "importlib-metadata>=4.4;python_version<'3.10'"
@@ -25,7 +25,6 @@ dependencies = [
 keywords = ['markdown', 'markdown-parser', 'python-markdown', 'markdown-to-html']
 classifiers = [
     'Development Status :: 5 - Production/Stable',
-    'License :: OSI Approved :: BSD License',
     'Operating System :: OS Independent',
     'Programming Language :: Python',
     'Programming Language :: Python :: 3',
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,7 @@ maintainers = [
     {name = 'Isaac Muse'}
 ]
 license = "BSD-3-Clause"
+license-files = ["LICENSE.md"]
 requires-python = '>=3.9'
 dependencies = [
     "importlib-metadata>=4.4;python_version<'3.10'"


cdce8p:
Missed that Python 3.8 was still supported. Setuptools `v77` is only available for Python `>= 3.9` unfortunately.

Not completely sure about the policy for EOL Python versions, though I did find #760. Pushed a new commit to drop 3.8. Please let me know if I should create a separate PR for that instead or if you don't want to do it at this time.

mitya57 (approved):

docs/changelog.md
```diff
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 * DRY fix in `abbr` extension by introducing method `create_element` (#1483).
 * Clean up test directory some removing some redundant tests and port
   non-redundant cases to the newer test framework.
+* Dropped support for Python 3.8 as it has reached its EOL.
```
mitya57: To make `checkspelling` happy, please replace `EOL` with `end of life`.

waylan:
As per #760, Python 3.8 should be dropped and support for 3.13 should be added at the same time. In fact, that could have happened last October. So, let's get that out in the next release. However, that should probably be a separate PR from this.

Regarding the use of license expressions, adding support make sense as it allows tools to more easily identify the license used. However, the current proposed change removes reference to the license file. That file includes more than the text of the license. Specifically, the authors and copyright history. Note that PEP 639 also includes a [License-File field](https://peps.python.org/pep-0639/#add-license-file-field). We should be including that along with the License-Expression field.

pyproject.toml
```diff
@@ -17,19 +17,17 @@ maintainers = [
     {name = 'Waylan Limberg', email = 'python.markdown@gmail.com'},
     {name = 'Isaac Muse'}
 ]
-license = {file = 'LICENSE.md'}
-requires-python = '>=3.8'
+license = "BSD-3-Clause"
```
waylan, line 20: Shouldn't this be `License-Expression = "BSD-3-Clause"`

mitya57: No. `License-Expression` is a field in the generated `PKG-INFO` file. In `pyproject.toml` it is just `license`: <https://packaging.python.org/en/latest/specifications/pyproject-toml/#license>.

waylan: That's not confusing at all.

cdce8p:
> As per #760, Python 3.8 should be dropped and support for 3.13 should be added at the same time. In fact, that could have happened last October. So, let's get that out in the next release. However, that should probably be a separate PR from this.

I believe you opened a PR for it already a while ago, #1487. Not sure what the current status of it is but AFAICT the only thing missing might be a rebase.

> However, the current proposed change removes reference to the license file.

Not quite. Setuptools has a set of default patterns it uses to detect license files. Those are
```
['LICEN[CS]E*', 'COPYING*', 'NOTICE*', 'AUTHORS*']
```
So `LICENSE.md` was included all along. Just not with the full text inside the metadata itself. I didn't include it in the diff but the following line is still present in the metadata `License-File: LICENSE.md`. The file itself is located inside the `.dist-info` folder: `markdown-3.7.dist-info/licenses/LICENSE.md`.

With my last commit I added a line to make it more explicit.

cdce8p:
Rebased the PR. All tests should pass now. This would be ready from my side.

PR #1521, opened 2025-04-02
Fix incorrect TOC list structure in docs

SoulPrudya:
I initially thought the list hierarchy was implemented incorrectly and was ready to skip using this module. After testing, I realized that the issue was in the documentation, which had been incorrect for 16 years.


waylan (approved):
You are correct. The closing `li` tags where in the wrong place. Good catch. Thank you.

PR #1522, opened 2025-04-03
markdown/__main__.py: (parse_options): order options alphabetically

jaalto:
With -h, --help, order the option listing alphabetically.

### DIFF
--- a/markdown/__main__.py
+++ b/markdown/__main__.py
@@ -54,19 +54,6 @@ def parse_options(args=None, values=None):
     ver = "%%prog %s" % markdown.__version__
 
     parser = optparse.OptionParser(usage=usage, description=desc, version=ver)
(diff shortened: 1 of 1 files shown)


waylan:
What is the benefit of this change?

Currently, the options are listed logically. For example, `--encoding` is listed right after `--file` as it applies to the input file. Similarly, `--extension_configs` is listed right after `--extensions`. And then we group together `--quiet`, `--verbose`, and `--noisy` in that logically progressive order. Seems to me that this is much more helpful to users than to list the options alphabetically.

facelessuser:
I think this is a highly subjective change. I'm personally a -1 on this. I prefer the logical organization.

waylan:
I agree that this is subjective. As the maintainers prefer the existing scheme and no persuasive arguments were provided to support the. change, I am closing this as won't fix.

PR #1527, opened 2025-04-15
Ensure `md_in_html` does not drop content

### LINKED ISSUE #1526: 3.8 - Layout issues
## Introduction
The recent release of 3.8 has resulted in some layout issues for me and I am attempting to determine if I need to change my generated pages or if this is a problem with this release. If this is a "me" problem, then I have 2 years worth of docs (1,000+) to review and change before I can switch to 3.8.

I have narrowed down one issue and attached is a simple docker-compose project with three services - 3.7, 3.8 and 3.8-fix - that highlights the issue and a potential change I need to make, albeit that the results are not identical to 3.7.

The issue identified concerns the layout/format of multiple `div` tags. One example being:

```
<div style="display: inline-flex" markdown>
  <div class="circle"></div>AAAAA<div class="circle"></div>BBBBB<div class="circle"></div><span>CCCCC</span>
</div>
```

## Docker 
The are three docker projects, each contain 4 files, a `Dockerfile`, a python `requirements.txt`, an mkdocs configuration file and a sample `index.md`. 

The 3.7 and 3.8 projects differ only in pinning the version of markdown and 3.8-fix is the same as 3.8 but with a few minor changes to `index.md` to mitigate and attempt to resolve the layout issues.

[markdown-3.8-problem.tar.gz](https://github.com/user-attachments/files/19754989/markdown-3.8-problem.tar.gz)

Should be a simple case of extracting and running:

```console
docker compose up --build
```

and then browse to http://localhost:8037/ http://localhost:8038/ and http://localhost:8039/

## Screenshots

Image

Image

Image

facelessuser:
Resolves #1526

### DIFF
--- a/tests/test_syntax/extensions/test_md_in_html.py
+++ b/tests/test_syntax/extensions/test_md_in_html.py
@@ -1517,7 +1517,7 @@ class TestMdInHTML(TestCase):
             extensions=['md_in_html']
         )
 
-    def test_trialing_content_after_tag_in_md_block(self):
+    def test_trailing_content_after_tag_in_md_block(self):
 
         self.assertMarkdownRenders(
             self.dedent(
--- a/tests/test_syntax/extensions/test_md_in_html.py
+++ b/tests/test_syntax/extensions/test_md_in_html.py
@@ -1519,6 +1519,9 @@ class TestMdInHTML(TestCase):
 
     def test_trailing_content_after_tag_in_md_block(self):
 
+        # It should be noted that this is not the way `md_in_html` is intended to be used.
+        # What we are specifically testing is an edge case where content was previously lost.
+        # Lost content should not happen.
         self.assertMarkdownRenders(
             self.dedent(
                 """
--- a/tests/test_syntax/extensions/test_md_in_html.py
+++ b/tests/test_syntax/extensions/test_md_in_html.py
@@ -1517,6 +1517,24 @@ class TestMdInHTML(TestCase):
             extensions=['md_in_html']
         )
 
+    def test_trialing_content_after_tag_in_md_block(self):
+
+        self.assertMarkdownRenders(
+            self.dedent(
+                """
+                <div markdown>
+                <div class="circle"></div>AAAAA<div class="circle"></div>
+                </div>
+                """
+            ),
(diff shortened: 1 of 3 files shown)


facelessuser:
@waylan 

- I don't mind fixing unrelated lint errors, I'm just not sure how you want to resolve the errors: ignore or actually adjust.
- I don't know the convention for the changelog, just tell me how to fix the check and I'll do it.

tests/test_syntax/extensions/test_md_in_html.py
```diff
@@ -1517,6 +1517,24 @@ def test_md1_code_cdata(self):
             extensions=['md_in_html']
         )
 
+    def test_trialing_content_after_tag_in_md_block(self):
```
mitya57:
```suggestion
    def test_trailing_content_after_tag_in_md_block(self):
```

tests/test_syntax/extensions/test_md_in_html.py
```diff
@@ -1517,6 +1517,24 @@ def test_md1_code_cdata(self):
...
+            self.dedent(
+                """
+                <div markdown>
+                <div class="circle"></div>AAAAA<div class="circle"></div>
+                </div>
+                """
+            ),
+            '<div>\n'
+            '<div class="circle"></div>\n'
+            '<p>AAAAA<div class="circle"></p>\n'
```
mitya57, line 1535: Specifying invalid HTML as expected result is a bit weird. Maybe at least add a comment like "Yes, it is invalid, but what we really test here is that trailing content is not lost"?

mitya57 (changes requested):

tests/test_syntax/extensions/test_md_in_html.py
facelessuser: Fair argument, I can do that.

mitya57 (approved):

facelessuser:
I think I got all the issues fixed. I took a closer look and I wasn't getting lint errors in CI, it was just spelling, which I fixed. Looking back through the project history, I also now realize I should specify the version as "Unreleased" in the changelog. Everything is now passing.

Issue #1479, opened 2024-09-08
Document lack of sanitization of HTML output


MaddyGuthridge:
Like many other Markdown processors, Python-Markdown does not sanitize its output, meaning that malicious code can be embedded within markdown documents.

```md
# Some markdown document

<script>alert("Evil laughter")</script>
```

If this isn't made clear to users, there is a risk that they will unintentionally create opportunities for XSS attacks. It would be worthwhile documenting the lack of sanitization, and perhaps recommend an HTML sanitization library, such as [bleach](https://github.com/mozilla/bleach).

waylan:
Hmm, this used to me mentioned in our documentation. Not sure when or why it was removed. But, yes, I agree, we should be documenting this. Although, an argument has been made by some in the past that as all markdown parsers do not sanitize, there is no need to document this as there should be no expectation from users anyway. Personally, I recognize that not all users know or understand that and so we should be expressly stating as much. 

By the way, we used to recommend bleach as a solution. We stopped making that recommendation as the bleach project has been deprecated. That still appears to be the case.

dbader:
`nh3` seems like a good and actively maintained alternative to Bleach: https://pypi.org/project/nh3/ ([messense/nh3 on GitHub](https://github.com/messense/nh3))

Sources for the recommendation:
* https://adamj.eu/tech/2023/12/13/django-sanitize-incoming-html-nh3/
* https://realpython.com/podcasts/rpp/187/#t=3086

waylan:
`nh3` is great, but as it consists of bindings to a Rust package, that means it is not a pure Python package and not universally installable---and that gives me pause. I should also mention that we actually used to recommend [yourcelf/bleach-allowlist](https://github.com/yourcelf/bleach-allowlist),  which is a set of config options for Bleach which makes sense for HTML generated from Markdown. A similar package for `nh3` might be more useful to users that the raw package. 🤷‍♂

PR #1543, opened 2025-07-08
Document HTML sanitation policy

waylan:
This is a first draft of documentation addressing #1479. I don't think I will use this as-is. The same text is reused multiple times, which gets redundant. And we still don't address the API documentation in all the same methods/functions as well as the CLI documentation. I think perhaps a separate page would be more appropriate and then a one sentence warning which points to that page would be appropriate in each relevant location. The separate page could also document the policy for never raising an error on Markdown input and any other relevant security-based issues that might exist.

### DIFF
--- a/.spell-dict
+++ b/.spell-dict
@@ -78,7 +78,6 @@ munge
 namespace
 NanoDOM
 Neale
-nh3
 nosetests
 OrderedDict
 OrderedDicts
--- a/.spell-dict
+++ b/.spell-dict
@@ -54,6 +54,7 @@ implementers
 InlineProcessor
 Jiryu
 JSON
+JustHTML
 keepachangelog
 Kjell
 Krech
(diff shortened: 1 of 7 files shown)


waylan:
I think this is ready now. It ended up being more work than I expected. Any feedback is welcome.

A few notes:

1. We are now recommending [JustHTML](https://emilstenstrom.github.io/justhtml/) for HTML sanitation. Unlike other tools, it sanitizes by default and is a pure Python library.
2. I have included some direction for bleach and its Rust clone (nh3).
3.  I have made significant updates to the CLI documentation to better organize the subsections by importance. Also added some more information about piping output and the fact that we output HTML fragments. While technically not directly related to this issue, JustHTML provides an easy solution, so it seemed natural to include here.

mitya57 (approved):

facelessuser (approved):

PR #1546, opened 2025-07-21
Fix footnote ordering

### LINKED ISSUE #1367: Footnotes numbering

aeskildsen:
Footnotes are now displayed in the order they are referenced in the document text, not in the order they are defined.

- Set up a separate list to keep track of footnote reference order
- Reorder footnote dict before rendering the list of footnotes
- Update docstrings to reflect chagnes

Tracking footnote references is something the logically fits within inline processing. But it must happen before tree processing, because that is where the list of footnotes is rendered to html. For that reason, it is done within the block processor.

Fixes #1367


waylan:
The test failures for the `py314` test can be ignored as they are unrelated. I have opened #1547 to address them.

waylan:
> Tracking footnote references is something the logically fits within inline processing. But it must happen before tree processing, because that is where the list of footnotes is rendered to html. For that reason, it is done within the block processor.

While I understand that it may be easier to implement this way, I am unconvinced that it is imposable to implement in the inline processor. My concern is that it may match things it shouldn't. For example, what if a footnote is included in a code span? Some tests might convince me.

In fact, this needs some new tests added to demonstrate that it is working as advertised.

aeskildsen:
Thank you for the quick feedback. The code span example is definitely a valid concern.

I can add some tests to demonstrate the functionality + a footnote reference inside a code span.

I'll think about how to implement this with the inline processor. The problem is not detecting footnote references in the inline processor, it's the timing, since the tree processor which renders the footnotes runs before that.

aeskildsen:
I've added a bunch of tests, including footnote references within code spans. See in particular `TestFootnotes.testFootnoteOrder`.

Concerning the block vs inline discussion: One consideration I would to add is that my implementation does not remove or modify any of the footnote references/labels during block processing. It just registers the ids of any footnote references it finds in the blocks. I have not touched the parts of the code that remove substrings or render html elements. I merely reorder the dict that holds the ids and footnote content just before the footnotes are rendered.

waylan:
Sorry that I didn't mention this previously, but all new tests should be under the appropriate subdir of `/tests/test_syntax` ([`/tests/test_syntax/extensions.test_footnotes.py`](https://github.com/Python-Markdown/markdown/blob/master/tests/test_syntax/extensions/test_footnotes.py) in this case) and make use of [markdown.test_tools.TestCase](https://python-markdown.github.io/test_tools/#markdowntest_toolstestcase) when it makes sense to do so.

aeskildsen:
Ah, I see. No worries, I've moved and adapted the new tests.

tests/test_syntax/extensions/test_footnotes.py
```diff
@@ -336,3 +336,192 @@ def test_superscript_text(self):
...
+            '<p>First footnote.&#160;<a class="footnote-backref" href="#fnref:first" '
+            'title="Jump back to footnote 1 in the text">&#8617;</a></p>\n'
+            '</li>\n'
+            '<li id="fn:last">\n'
+            '<p>Second footnote.&#160;<a class="footnote-backref" href="#fnref:last" '
+            'title="Jump back to footnote 2 in the text">&#8617;</a></p>\n'
+            '</li>\n'
+            '</ol>\n'
+            '</div>'
+        )
```
aeskildsen: This tests the new behavior (footnotes ordered by reference appearence).

tests/test_syntax/extensions/test_footnotes.py
```diff
@@ -336,3 +336,192 @@ def test_superscript_text(self):
...
+            '<div class="footnote">\n'
+            '<hr />\n'
+            '<ol>\n'
+            '<li id="fn:1">\n'
+            '<p>The footnote.&#160;<a class="footnote-backref" href="#fnref:1" '
+            'title="Jump back to footnote 1 in the text">&#8617;</a></p>\n'
+            '</li>\n'
+            '</ol>\n'
+            '</div>'
+        )
```
aeskildsen: These test parsing of footnote references within other structures.

aeskildsen:
Sorry, don't review or merge this just yet. I've come up with a test case that messes up the order (when a footnote reference id inside e.g. a code span is identical to a proper footnote reference id elsewhere). It is an edge case, but I'd rather get this corrected before review.

I'll provide a fix early next week.

aeskildsen:
I've gone down the rabbithole with this and can report a few things:

> While I understand that it may be easier to implement this way, I am unconvinced that it is imposable to implement in the inline processor.

It is indeed possible (trivial, even) to implement tracking the order of footnote references in an Inlineprocessor. What is not possible, however, is to then generate the list of footnotes in the order we have tracked, because the list of footnotes is produced by a Treeprocessor. To my knowledge, that is simply how the architecture is designed; Treeprocessors always run before Inlineprocessors.

So instead of trying to *generate* the footnotes div in the correct order, I have explored different ways of *reordering* it in a Postprocessor. They all lead to a fair amount of reparsing that seems counterintuitive and excessive at this stage in the pipeline. I've had some succes with a regex-based reordering method that works on the generated footnotes div. But it fails `tests.test_legacy.TestExtensionsExtra.test_raw_html` due to what I think is a state management issue between `footnotes` and `md_in_html`. I don't currently have the resources to dig deeper into the `extras` extension to figure out what that issue is.

So. Based on my work thus far, **I will suggest that for this PR we go back to [0c2e4bf](https://github.com/Python-Markdown/markdown/pull/1546/commits/0c2e4bf56734206c0bd592428df525a94da567df) and proceed with review**.

This entails ignoring the edge case that [ce45459](https://github.com/Python-Markdown/markdown/pull/1546/commits/ce45459432b06aa63c2df7eab350957f1910742d) tests. It is a pretty remote possibility, but even if it happens, it will not produce any errors. It will simply result in an illogical order of footnotes. This is, in my view, much preferable to the current behavior where the order of footnote definitions dictates the order of footnotes.

waylan:
> It is indeed possible (trivial, even) to implement tracking the order of footnote references in an Inlineprocessor. What is not possible, however, is to then generate the list of footnotes in the order we have tracked, because the list of footnotes is produced by a Treeprocessor. To my knowledge, that is simply how the architecture is designed; Treeprocessors always run before Inlineprocessors.

Actually, Inlineprocessors are wrapped in and run from [markdown.treeprocessors.InlineProcessor](https://python-markdown.github.io/reference/markdown/treeprocessors/#markdown.treeprocessors.InlineProcessor), which is the first treeprocessor by default (as listed [here](https://python-markdown.github.io/reference/markdown/treeprocessors/#markdown.treeprocessors.build_treeprocessors)). Therefore, it is possible to run a Treeprocessors either before or after inline processors by setting the priority either higher or lower than `markdown.treeprocessors.InlineProcessor`.

aeskildsen:
> Actually, Inlineprocessors are wrapped in and run from [markdown.treeprocessors.InlineProcessor](https://python-markdown.github.io/reference/markdown/treeprocessors/#markdown.treeprocessors.InlineProcessor), which is the first treeprocessor by default (as listed [here](https://python-markdown.github.io/reference/markdown/treeprocessors/#markdown.treeprocessors.build_treeprocessors)). Therefore, it is possible to run a Treeprocessors either before or after inline processors by setting the priority either higher or lower than `markdown.treeprocessors.InlineProcessor`.

Ah, that makes a lot of sense. Thank you for explaining that.

[5d7f8ca](https://github.com/Python-Markdown/markdown/pull/1546/commits/5d7f8ca75789dc16ef234f007b520b66af770d08) adds an implementation that uses a treeprocessor, which is the cleanest solution of the ones I've experimented with.

It does fail one test, however: `test_legacy.TestExtensionsExtra.test_raw_html`. I don't fully understand why that particular test gets the order wrong, but I *think* it has to do with how state gets jumbled when `md_in_html` calls markdown multiple times. It's the same problem I encountered with a regex-based approach. As far as I can tell, the footnotes extension works correctly with `.reset()`, so that shouldn't be the issue.

facelessuser:
> As far as I can tell, the footnotes extension works correctly with `.reset()`, so that shouldn't be the issue.

The tests should be cleaning up after themselves. I would probably need to understand more about what the differences are and why they are different.

aeskildsen:
The specific input file that fails to parse correctly is tests/extensions/extra/raw-html.txt

But let me see if I can show this in a simpler way. The problem occurs when we use the `extra` extension:

```python
import markdown

MD = """
<div markdown="1">
Reference to [^first]

[^first]: First footnote definition
</div>

<div markdown="1">
Reference to [^second]

[^second]: Second footnote definition
</div>
"""

result = markdown.markdown(MD, extensions=["extra"], extension_configs={})
print(result)
```

Prints this:

```html
<div>
<p>Reference to <sup id="fnref:first"><a class="footnote-ref" href="#fn:first">2</a></sup></p>
</div>
<div>
<p>Reference to <sup id="fnref:second"><a class="footnote-ref" href="#fn:second">1</a></sup></p>
</div>
<div class="footnote">
<hr />
<ol>
<li id="fn:second">
<p>Second footnote definition&#160;<a class="footnote-backref" href="#fnref:second" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
<li id="fn:first">
<p>First footnote definition&#160;<a class="footnote-backref" href="#fnref:first" title="Jump back to footnote 2 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

Which should be this:

```html
<div>
<p>Reference to <sup id="fnref:first"><a class="footnote-ref" href="#fn:first">1</a></sup></p>
</div>
<div>
<p>Reference to <sup id="fnref:second"><a class="footnote-ref" href="#fn:second">2</a></sup></p>
</div>
<div class="footnote">
<hr />
<ol>
<li id="fn:first">
<p>First footnote definition&#160;<a class="footnote-backref" href="#fnref:first" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
<li id="fn:second">
<p>Second footnote definition&#160;<a class="footnote-backref" href="#fnref:second" title="Jump back to footnote 2 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

*Why?*
I'm fairly certain that the footnotes extension is working as expected, as it passes all other tests. The `md_in_html` extension appears to be involved here, but I don't know why or how. Perhaps the FootnoteInlineProcessor gets called in an improper order? Its job is to register the order of the footnote references and construct the `sup` element that contains the footnote reference.

facelessuser:
`md_in_html` is a fairly complicated piece of code. Worst case, this illustrates that it may be impossible for this approach to guarantee order, but maybe there is something we can figure out.

I will state that plugins can do almost anything, and can cause some content to be processed out of the expected order, so there may be other 3rd party plugins that could potentially manifest something similar.

If it turns out we cannot guarantee order, I do wonder if all this effort is worth it. I do imagine it is possible to reasonably ensure order, but it may not be through the current approach. Just to be clear, I have not explored the current approach, so I don't know yet if what you are doing can simply be tweaked or not. I'll have to find some time to explore further.

waylan:
I will note that the failing test passes without any of the changes here. In other words, the original order is preserved. So, the changes here seem to be changing the order.

waylan:
With the changes here, this is what I am getting:

```python
>>> src = """
... <div markdown="1">
... Reference to [^first]
...
... [^first]: First footnote definition
... </div>
...
... <div markdown="1">
... Reference to [^second]
...
... [^second]: Second footnote definition
... </div>
... """
>>> md = markdown.Markdown(extensions=['footnotes', 'md_in_html'])
>>> print(md.convert(src))
<div>
<p>Reference to <sup id="fnref:first"><a class="footnote-ref" href="#fn:first">2</a></sup></p>
</div>
<div>
<p>Reference to <sup id="fnref:second"><a class="footnote-ref" href="#fn:second">1</a></sup></p>
</div>
<div class="footnote">
<hr />
<ol>
<li id="fn:second">
<p>Second footnote definition&#160;<a class="footnote-backref" href="#fnref:second" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
<li id="fn:first">
<p>First footnote definition&#160;<a class="footnote-backref" href="#fnref:first" title="Jump back to footnote 2 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

That matches your result. But let's dig a little deeper...

```python
>>> md.inlinePatterns[2]
<markdown.extensions.footnotes.FootnoteInlineProcessor object at 0x000001A3C5160B00>
>>> md.inlinePatterns[2].footnotes.footnote_order
['second', 'first']
```

Here we can see that the `footnote_order` is set to `['second', 'first']`, which explains the output. Now to work out how they order was set that way. Let's add a debugging print statement...

```diff
diff --git a/markdown/extensions/footnotes.py b/markdown/extensions/footnotes.py
index 0d08dc7..7f29d06 100644
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -163,6 +163,7 @@ class FootnoteExtension(Extension):
     def addFootnoteRef(self, id: str) -> None:
         """ Store a footnote reference id in order of appearance. """
         if id not in self.footnote_order:
+            print(f'Saving footnote ID: {id}')
             self.footnote_order.append(id)

     def get_separator(self) -> str:
```
and check the output...

``` python
>>> markdown.markdown(src, extensions=['footnotes', 'md_in_html'])
Saving footnote ID: second
Saving footnote ID: first
'<div>\n<p>Reference to <sup id="fnref:first"><a class="footnote-ref" href="#fn:first">2</a></sup></p>\n</div>\n<div>\n<p>Reference to <sup id="fnref:second"><a class="footnote-ref" href="#fn:second">1</a></sup></p>\n</div>\n<div class="footnote">\n<hr />\n<ol>\n<li id="fn:second">\n<p>Second footnote definition&#160;<a class="footnote-backref" href="#fnref:second" title="Jump back to footnote 1 in the text">&#8617;</a></p>\n</li>\n<li id="fn:first">\n<p>First footnote definition&#160;<a class="footnote-backref" href="#fnref:first" title="Jump back to footnote 2 in the text">&#8617;</a></p>\n</li>\n</ol>\n</div>'
```

Sure enough, the `second` ID is being processed/saved before the `first`. Presumably, the `md_in_html` extension is messing with the order that items are processed. I don't recall all the ins-and-outs of that extension, but even if we were able to change that extension so that it processed things in order, there is no guarantee that all third-party extensions would do so. And that brings us to the point that @facelssuser was making: It may not make sense to even attempt to support this.

aeskildsen:
EDIT: I wrote this before reading your latest reply, @waylan .

> I will note that the failing test passes without any of the changes here. In other words, the original order is preserved. So, the changes here seem to be changing the order.

Correct. I am fairly certain that is because the order is dictated by footnote _definitions_, which are detected during block processing. As opposed to my recent changes which detect the order of footnote _references_ during inline processing. My most recent implementation assumes that inline matches are processed in document order, depth first. I suspect that is not the case in this particular test, due to `md_in_html`. I completely understand if my assumption is wrong and that processing order may not be guaranteed, as @facelessuser mentioned.

Though I would also suspect that other plugins which rely upon inline matches to occur in document order would be susceptible to similar issues.

aeskildsen:
I will also say that the test we are talking about passes in my earlier implementation [0c2e4bf](https://github.com/Python-Markdown/markdown/pull/1546/commits/0c2e4bf56734206c0bd592428df525a94da567df). That implementation relies upon block processing to detect the order.

I think we can safely disregard the edge case I mentioned afterwards and tested in [5d7f8ca](https://github.com/Python-Markdown/markdown/pull/1546/commits/5d7f8ca75789dc16ef234f007b520b66af770d08). Worst case scenario is: A user uses the same footnote id in a code example as in a real footnote in the same document. That footnote may get a different place in the order of footnotes than expected, but it will do so silently and not produce an error.

For me, that would be preferable to the current behavior, where improper order can be achieved much easier than in the edge case; one just has to follow the instructions in the documentation and put definitions "anywhere in the document", as [stated in the PHP Markdown Extra docs](https://michelf.ca/projects/php-markdown/extra/#footnotes).

aeskildsen:
> Presumably, the `md_in_html` extension is messing with the order that items are processed. I don't recall all the ins-and-outs of that extension, but even if we were able to change that extension so that it processed things in order, there is no guarantee that all third-party extensions would do so. And that brings us to the point that @facelssuser was making: It may not make sense to even attempt to support this.

Agreed - it seems like a lot of work for what is, admittedly, a minor issue. I completely understand if that is not a relevant goal to pursue. Though I hope you will consider what I suggested above concerning [0c2e4bf](https://github.com/Python-Markdown/markdown/pull/1546/commits/0c2e4bf56734206c0bd592428df525a94da567df), which passes the test.

waylan:
I have done some spelunking with the Python debugger and md_in_html is not processing anything out of order. It behaves exactly as you would think. In fact, after the blockparser has run, but before any treeprocessors have run, the tree contains the following structure:

```html
<div>
<p>Reference to [^first]</p>
</div>
<div>
<p>Reference to [^second]</p>
</div>
```

The issue would seem to be with treeprocessors or inlineprocessors. What would be causing them to process things out of order I don't know.

facelessuser:
Oh, that's interesting. I may dig a little if/when I get some time.

aeskildsen:
I'll await for now then, as I don't know much about the internals of the processors. Do let me know if I can do anything to assist.

waylan:
I resolved the issue with one minor change in 1c160ef. `list.pop()` removes and returns the last item from a list. Therefore, Elements in the stack were being processed in reverse order. However, only Elements with child Elements ever appear in the stack. Therefore, simple Markdown documents did not usually encounter this issue. However, deeply nested documents could. Once I understood the issue, it was easy to create other examples which failed without raw HTML. For example, prior to my change...

```python
>>> src = '''
... * Reference to [^first]
...
... * Reference to [^second]
...
... [^first]: First footnote definition
... [^second]: Second footnote definition
... '''
>>> print(markdown.markdown(src, extensions=['footnotes']))
<ul>
<li>
<p>Reference to <sup id="fnref:first"><a class="footnote-ref" href="#fn:first">2</a></sup></p>
</li>
<li>
<p>Reference to <sup id="fnref:second"><a class="footnote-ref" href="#fn:second">1</a></sup></p>
</li>
</ul>
<div class="footnote">
<hr />
<ol>
<li id="fn:second">
<p>Second footnote definition&#160;<a class="footnote-backref" href="#fnref:second" title="Jump back to footnote 1 in the text">&#8617;</a></p>
</li>
<li id="fn:first">
<p>First footnote definition&#160;<a class="footnote-backref" href="#fnref:first" title="Jump back to footnote 2 in the text">&#8617;</a></p>
</li>
</ol>
</div>
```

Note that that example uses a "loose list" (there is a blank line between the list items), which results in each list item being wrapped in a `<p>` element. This creates the same scenario as the `md_in_html` example and fails in the same way. Simply by ensuring that we always pop the first element (with `stack.pop(0)`) the issue is resolved. I have added that as a test in 62b12b7.

Fortunately, none of the tests seem to rely on this weird behavior, so I think it should be safe to move forward. Hopefully no third-party extensions are relying on this and break from the change. ¯\\\_(ツ)\_/¯

waylan:
I'm wondering if perhaps 1c160ef should be in a separate PR by itself and this PR should be rebased on top of that. At a minimum, we need to note this change in the changelog. @facelessuser any thoughts?

facelessuser:
I'm not sure I have any strong opinions on the change being separate. If we want to specifically track when this change occured, then yeah, maybe submit it as a separate change. This footnote change exposed potential issues with this, and seems like a fine place to fix it as well.

I would definitely note the change in the changelog though, if there are concerns, it might cause issues.

markdown/extensions/footnotes.py
```diff
@@ -214,7 +226,7 @@ def makeFootnotesDiv(self, root: etree.Element) -> etree.Element | None:
 
 
 class FootnoteBlockProcessor(BlockProcessor):
-    """ Find all footnote references and store for later use. """
+    """ Find footnote definitions and references, storing both for later use. """
```
waylan: I'm assuming this change applies to a previous iteration and should be reversed now.

markdown/extensions/footnotes.py
```diff
@@ -226,8 +238,9 @@ def test(self, parent: etree.Element, block: str) -> bool:
         return True
 
     def run(self, parent: etree.Element, blocks: list[str]) -> bool:
-        """ Find, set, and remove footnote definitions. """
+        """ Find, set, and remove footnote definitions. Find footnote references."""
```
waylan: Again, I'm assuming this change applies to a previous iteration and should be reversed now.

markdown/extensions/footnotes.py
```diff
@@ -401,6 +416,44 @@ def run(self, root: etree.Element) -> None:
                 root.append(footnotesDiv)
 
 
+class FootnoteReorderingProcessor(Treeprocessor):
+    """ Reorder list items in the footnotes div. """
```
waylan, line 429: I don't love that this is a separate Treeprocessor which is run after the footnotes div is built. However, I understand that (as documented in comments) the div needs to be built before inline processing so that inline processing happens on the contents of the footnotes, but the order of footnotes is not known until after inline processing. I'll need to think about this.

markdown/treeprocessors.py
```diff
@@ -368,7 +368,7 @@ def run(self, tree: etree.Element, ancestors: list[str] | None = None) -> etree.
         stack = [(tree, tree_parents)]
 
         while stack:
-            currElement, parents = stack.pop()
+            currElement, parents = stack.pop(0)
```
waylan, line 371: Let's not forget to specifically call out this change in the changelog.

tests/test_syntax/extensions/test_footnotes.py
```diff
@@ -336,3 +336,257 @@ def test_superscript_text(self):
...
+            self.dedent(
+                '''
+                * Reference to [^first]
+
+                * Reference to [^second]
+
+                [^first]: First footnote definition
+                [^second]: Second footnote definition
+                '''
+            ),
```
waylan, line 547: Let's reformat the new tests to all use this format, which is much easier to read. I understand that the expected output often includes lines which are too long, so we don't use string literal there, but the input doesn't have that problem.

waylan (changes requested):

tests/test_syntax/extensions/test_footnotes.py
waylan: To clarify, if the input for a test is only a single line (no line breaks), this is unnecessary. However, for any input which contains multiple lines, this is the preferred way. That said, I do not expect you to alter existing (otherwise unmodified) tests to match this format, only tests you are adding/modifying.

markdown/extensions/footnotes.py
aeskildsen: Correct. However, the original docstring seems incorrect. The block processor finds footnotes _definitions_, not footnote _references_. So it should be changed to this: `""" Find all footnote definitions and store for later use. """`

markdown/extensions/footnotes.py
aeskildsen: Correct.

markdown/extensions/footnotes.py
aeskildsen: Exactly. For the reasons you list here, I don't believe it is possible to do both in one Treeprocessor. I'll be happy to change it if you can think of a better approach.

tests/test_syntax/extensions/test_footnotes.py
aeskildsen: Agreed, will do.

I assume this pertains to `self.dedent` and not single vs double quotes.

facelessuser: Yes, dedent

aeskildsen: I also assume we shouldn't use `self.dedent` on single line inputs.

facelessuser: Yes, I believe that was the guidance.

waylan: This pertains to `dedent` and triple-quote or multi-line strings. Triple-quote strings can either use single or double quotes (`'''` or `"""`). 

Regarding the use of single or double quotes in code (`'` or `"`), whether that pertains to triple-quote strings or not, we do not enforce any particular usage. \<rant>However, personally, I almost always use and prefer single quotes as they are easier to type (no need to use the shift key) and add less noise (one tick mark vs. two). I only use double quotes when I'm avoiding the need to escape quotes within a string (`"Bob's"` vs. `'Bob\'s'`). Personally, I have always assumed that those who default to using double quotes picked up that habit from other languages which impute different meaning to single and double quotes. As that difference doesn't exist in Python, why would anyone choose to use double quotes over single quotes? \</rant> In any event, as I said, we don't enforce any particular preference here.

markdown/extensions/footnotes.py
waylan: Something that we could do is build the footnote div after inline processing (with the correct order out of the gate) and them explicitly run inline processing on the div before inserting it into the document. However, that could cause the body of those footnotes to miss out on any other treeprocessors run beforehand, including from any random third-party extension. Upon further consideration, I think the current implementation is probably the least disruptive. We can leave this as-is.

markdown/treeprocessors.py
aeskildsen: I have added an entry, but I am not sure whether it is sufficiently detailed:

* Ensure inline processing iterates through elements in document order.

waylan (approved):
Looks good. Thank you for all your work on this.

aeskildsen:
> Looks good. Thank you for all your work on this.

I'm happy to help. Thank you for a nice process.

One small thing that keeps lingering at the back of my mind: I wonder if the change to `stack.pop(0)` could have a slight impact on performance. Would it be worth benchmarking?

docs/changelog.md
```diff
@@ -15,6 +15,8 @@ See the [Contributing Guide](contributing.md) for details.
 ### Fixed
 
 * Fix handling of incomplete HTML tags in code spans in Python 3.14.
+* Fix issue with footnote ordering (#1367).
```
waylan: I'm inclined to classify this as a change in behavior rather than a bug fix. While it is unclear if PHP Markdown changed their behavior at some point or not, it is clear from reading our code (and code comments) that we were intentionally ordering footnotes based on the order of the definitions. Therefore, I think it would be a stretch to call this a bug fix.

If that is the case. then this entry in the changelog would need to be removed from the `Fixed` section as added to a new `Changed` section. And that would mean that the next release would be a point release (presumably 3.9) rather than a bugfix release (such as 3.8.3).

On that note, it also occurs to me that some users may prefer the previous behavior, which raises the question: would it make sense to add a config setting to the extension which allows the user to select their preferred behavior? Presumably, such a setting would either include or exclude the `FootnoteReorderingProcessor`.

Thoughts?

waylan: Sorry I didn't raise this earlier, but I hadn't fully formed my thoughts until now. And even now, I am unsure how we should proceed. I'm just looking for feedback.

waylan:
> One small thing that keeps lingering at the back of my mind: I wonder if the change to `stack.pop(0)` could have a slight impact on performance. Would it be worth benchmarking?

Perhaps. I briefly thought about that as well, but what could we do to address any potential performance issue? Would using [collections.deque](https://docs.python.org/3/library/collections.html#collections.deque) for the stack make sense here? If you are so inclined, feel free to benchmark and report back.

docs/changelog.md
facelessuser: It's an enhancement. I wouldn't classify it as a bug fix.

aeskildsen: I see what you mean. I'm fine calling this a change in behavior/an enhancement.

As for a config setting: Whether some users rely on the current behavior, I can't really say. I don't think it would be complicated to implement though. There is a little more to it than in-/excluding the `FootnoteReorderingProcessor`, but not much, I think. I'd be happy to give it a shot.

aeskildsen:
> > One small thing that keeps lingering at the back of my mind: I wonder if the change to `stack.pop(0)` could have a slight impact on performance. Would it be worth benchmarking?
> 
> Perhaps. I briefly thought about that as well, but what could we do to address any potential performance issue? Would using [collections.deque](https://docs.python.org/3/library/collections.html#collections.deque) for the stack make sense here? If you are so inclined, feel free to benchmark and report back.

That's what I was thinking. I should be able to do some benchmarking in the next day or two.

aeskildsen:
So I managed to do a bit of benchmarking.

I benchmarked the processing of two markdown documents using `pyperf`:

- *sample_doc.md:* A document attempting to simulate a big chunk of realistic README-like markdown. Contains all the markdown from the top level of Python-Markdown's docs directory.
- *nested_doc.md:* A document with 22 examples of nested structures which requires the 'extra' extension for `footnotes`, `md_in_html`, etc. It contains the loose list example @waylan provided, the contents of the aforementioned raw-html.txt test file, a bunch of AI-generated markdown with nested structures, and a single line containing 2000 footnote references (adapted from the original test file `footnote_many_footnotes.txt`), a few of which are nested within `**` and `__`.

Let me know if you want to see the contents of these files.

The `stack.pop(0)` fix seems to have a very negligible impact on performance. I'm no statistician, but I would guess the difference is within the margin of error for the benchmarks I ran. `collections.deque` seems to show a very minor performance improvement in the case of the nested document, but again, likely not far from the margin of error.

I am no expert on how the inline processor works, but I think the reason we are not seeing big differences here is that inline nesting mostly does not go very deep. I believe this is what @waylan mentioned earlier. As I understand it, `list.pop(0)` is O(n) and `deque.popleft()` is O(1), but when the list is short the difference is probably negligible.

## Data

### sample_doc.md, default pyperf timeit benchmark

v3.8.2 with no changes: Mean +- std dev: 50.6 ms +- 0.9 ms
commit d8ecae6 (stack.pop(0) fix): Mean +- std dev: 51.4 ms +- 1.9 ms
commit d8ecae6 + collections.deque instead of list: Mean +- std dev: 51.9 ms +- 2.3 ms

### nested_doc.md, default pyperf timeit benchmark

Here I used `extensions=['extra']`.

v3.8.2 with no changes: Mean +- std dev: 156 ms +- 3 ms
commit d8ecae6 (with stack.pop(0) fix): Mean +- std dev: 160 ms +- 5 ms
commit d8ecae6 + collections.deque instead of list: Mean +- std dev: 158 ms +- 3 ms

### nested_doc.md, rigorous pyperf timeit benchmark

Here I used `extensions=['extra']`.

v3.8.2 with no changes: Mean +- std dev: 160 ms +- 9 ms
commit d8ecae6 (with stack.pop(0) fix): Mean +- std dev: 164 ms +- 11 ms
commit d8ecae6 + collections.deque instead of list: Mean +- std dev: 160 ms +- 7 ms

## TLDR

There does not appear to be a noteworthy performance impact from the `stack.pop(0)` fix. I think it can be merged without major concerns about performance.

It is possible to use `collections.deque` as a drop-in replacement for the list, which could offer a slight improvement in documents with very complex nested inline structures. We could consider testing how it performs on a wider range of documents first, and so perhaps a different PR would be more appropriate for that.

docs/changelog.md
waylan: Let's go forward with this then.

aeskildsen: Okay. Since this is a binary choice I guess a boolean config setting is in order? Maybe along the lines of `LEGACY_ORDERING` or `ORDER_BY_DEFINITION`...

But for the latter, we should be aware that the docs operate with a different terminology from the code when it comes to footnote "references" and "definitions" (although not consistently). That's not a problem for me personally, but the terminology around the config setting should probably align with the terms used in [the docs](https://python-markdown.github.io/extensions/footnotes/#syntax):

> A footnote label must start with a caret ^ and may contain any inline text (including spaces) between a set of square brackets []. Only the first caret has any special meaning.
> A footnote content must start with the label followed by a colon and at least one space.

I wouldn't mind going over the docs and changing from label/content to reference/definition if you think that is better.

waylan: Yes, "the terminology around the config setting should probably align with the terms used in the docs." I see that PHP uses 3 words:

1. **marker**: what we have been calling a label or reference.
2. **definition**: same as our definition/content.
3. **name**: the identifying string in both the marker and definition (`foo` of `[^foo]`).

I don't really have a strong opinion, but it might be reasonable for the terms in our docs to match PHP's.

And yes, something like `ORDER_BY_DEFINITION` would be the correct way to go. Just be careful to not insinuate that they would be sorted (for example alphanumerically by name). Rather, the order is determined by the order of definitions in the source. Maybe something like `USE_DEFINITION_ORDER` or `PRESERVE_DEF_ORDER`. I'm open to better suggestions.

aeskildsen: I think `USE_DEFINITION_ORDER` is as clear as can be. I'll explain it in the docs.

I think it makes sense to match the PHP terminology, except I think we should avoid the term "marker" as we have a config setting called `PLACE_MARKER`. They also use "reference" as a synonym for "marker", so I suggest this (which luckily also aligns well with our code comments and variable naming):

1. reference
2. definition
3. name (we call this `id` in the code, but that's fine I think)

docs/extensions/footnotes.md
```diff
@@ -24,26 +24,32 @@ the output.
...
-follow the label either on the same line or on the next line. The content may
-contain multiple lines, paragraphs, code blocks, blockquotes and most any other
-markdown syntax. The additional lines must be indented one level (four spaces or
-one tab).
-
-When working with multiple blocks, it may be helpful to start the content on a
-separate line from the label which defines the content. This way the entire block
+A **footnote name**  is a string that uniquely identifies a footnote within the
+document. It may contain any character which is valid for an HTML id attribute
+(including spaces). Examples: `1` in `[^1]` and `@#$%` in `[^@#$%]`.
```
waylan: This is a bit if a nitpick, but I wonder if perhaps we should add an example which uses a word. I expect that names like `@#$%` are very unlikely to be used, whereas names which consist of a word will be more common. Someone might get the impression that word-based names are not allowed by the current examples (despite the text clearly explaining that they are allowed). At the same time, I think we should keep `@#$%` and `1` as examples to avoid the same issue with them.

In all other respects, this is some great documentation.

aeskildsen: Actually, I had the same thought but ended up with the examples that were already there. Added an example now.
And thanks!

waylan (approved):
Excellent! This looks good to me. Thanks for doing the work on this.

Issue #1462, opened 2024-04-25
Separate extensions from core


waylan:
From time-to-time we have mentioned possibly separating the built-in extensions from the core. This would involve breaking each of the 18 existing extensions out into a separate repo and Python package. This has a few advantages.

1. Development of core would not hold up release of an update to an extension or visa versa.
2. Issue tracking would be more focused as each repo would have its own issue tracker.
3. It would provide a clearer distinction between extensions and core behavior, which would (hopefully) reduce requests for additional extensions to be added to the core.
4. Some existing extensions could potentially be clearly marked as being in "maintenance mode" without any effect on the core or other extensions. For example, an extension could be clearly marked as only receiving updates to continue working with future versions, but to not receive any new feature additions. Or an extension could be marked end-of-life, which means no updates at all going forward. But it could continue to be used so long as it supports the dependencies the user has installed.

I propose that each extension and its related tests would be moved to a new repo (under the `Python-Markdown` org) and be distributed as a separate Python package. Presumably, each extension would be packaged as a [namespace package](https://packaging.python.org/en/latest/guides/packaging-namespace-packages/), Therefore they would still be importable from the same path (`markdown.extensions.{extension_name}`).

There is one major issue that still needs addressing, which is what has stopped us from moving forward with this thus far. Specifically, most existing users expect to get the existing extensions when installing `markdown`. Therefore, we still need to ensure each separate package gets installed with it. There are a few different ways we could approach this.

### Option 1 (the namespace method)
Keep the core functionality in the existing  `Python-Markdown/markdown` repo and list all 18 of the extensions as dependencies which get installed as namespace packages. There would be no way to install the core without all 18 extensions.

This option would be simple to implement, and would not even be noticed by most users. However, it might be confusing for a user who was inclined to report an issue or make a feature request. They would have to track down the repo for the relevant package that they had not explicitly installed . 

On the other hand, releases would be easy here. Any updates to core or an extension could just be released. Individual extensions would simply indicate the earliest version of `markdown` that they support, which could differ from extension to extension over time. It might be worth exploring if a future namespace package would override/replace a submodule of the existing (non-separated) version of `markdown`. We might need to limit to specific versions in an extension's dependencies to avoid weirdness here.

### Option 2 (the core method)
[...]

PR #1553, opened 2025-09-17
Video extension

jonata:
Add `video` Markdown extension

This PR introduces a new Markdown extension, `video`, that automatically transforms image links pointing to video files into HTML5 `<video>` elements.

Features:

* Supports common video formats (`mp4`, `webm`, `ogg`, `mov`) with configurable extensions.
* Parses query parameters in Markdown links to set HTML5 attributes (`autoplay`, `controls`, `loop`, `muted`).
* Supports the `poster` attribute for video thumbnails.
* Preserves `alt` text as the video `title`.
* Works as a standard Markdown extension:

  ```python
  import markdown
  md = markdown.Markdown(extensions=['video'])
  html = md.convert('![Demo](video.mp4?poster=thumb.png&autoplay=1&controls=1)')
  ```

This keeps the extension consistent with the existing project structure and allows users to easily add videos in Markdown without custom HTML.

### DIFF
--- a/docs/extensions/index.md
+++ b/docs/extensions/index.md
@@ -55,6 +55,7 @@ Extension                          | Entry Point    | Dot Notation
 [SmartyPants]                      | `smarty`       | `markdown.extensions.smarty`
 [Table of Contents]                | `toc`          | `markdown.extensions.toc`
 [WikiLinks]                        | `wikilinks`    | `markdown.extensions.wikilinks`
+[Video]                            | `video`        | `markdown.extensions.video`
 
 [Extra]: extra.md
 [Abbreviations]: abbreviations.md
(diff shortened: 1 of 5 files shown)


facelessuser:
In general, it is encouraged that users who have new, innovative extensions host them as a 3rd party. It was one of the reasons for the creation of the extension system. That saves us from having to support any and all extensions of the community, which we do not have the resources to support.

AFAIK, Python Markdown is not accepting new extensions.

waylan:
@facelessuser is correct. We are not adding any new extensions. In fact, there is a proposal to separate all extensions from the core (see #1462). I suggest that you create a separate repo for your extension as a third-party extension, release it to PyPI and list it in our [wiki](https://github.com/Python-Markdown/markdown/wiki/Third-Party-Extensions).

PR #1572, opened 2025-12-01
Ensure nested elements inside inline comments are properly unescaped.

### LINKED ISSUE #1571: Placeholder in HTML markup
When the markdown source contains an inline HTML comment which contains a markdown link whose text contains markdown formatting, the resulting HTML contains the placeholder for the formatted text.

```py
>>> markdown.markdown("For example <!-- [**Bold link**](http://example.com) -->")
'<p>For example <!-- <a href="http://example.com">\x02klzzwxh:0000\x03</a> --></p>'
```

This is similar to #42, but only occurs when there is formatting in the link text.

Also notable is that this behavior for comments only occurs when the comment does not start a line — HTML comments at the start of a line are not processed for markdown syntax at all.

For a real-world example of this bug, see https://github.com/rust-lang/this-week-in-rust/issues/7299

waylan:
When inline Markdwon elements are nested, each nested level contains its own placeholder. Therefore, when unescaping, the regex substitution needs to be run again each nested level. As this is nested inside the match, it ensures only one non-matching regex will be run in each instace.

Closes #1571.

### DIFF
--- a/tests/test_syntax/inline/test_raw_html.py
+++ b/tests/test_syntax/inline/test_raw_html.py
@@ -34,3 +34,9 @@ class TestRawHtml(TestCase):
 
     def test_noname_tag(self):
         self.assertMarkdownRenders('<span></></span>', '<p><span>&lt;/&gt;</span></p>')
+
+    def test_markdown_nested_in_inline_comment(self):
+        self.assertMarkdownRenders(
+            'Example: <!-- [**Bold link**](http://example.com) -->',
+            '<p>Example: <!-- <a href="http://example.com"><strong>Bold link</strong></a> --></p>'
+        )
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -10,7 +10,7 @@ and this project adheres to the
 [Python Version Specification](https://packaging.python.org/en/latest/specifications/version-specifiers/).
 See the [Contributing Guide](contributing.md) for details.
 
-## [unreleased]
+## [Unreleased]
 
 ### Fixed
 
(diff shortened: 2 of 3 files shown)


waylan:
The checklinks and checkspelling linters are failing with a mkdocs build error. We can ignore them for this issue, but we will need to address that separately.

tests/test_syntax/inline/test_raw_html.py
```diff
@@ -34,3 +34,9 @@ def test_inline_html_backslashes(self):
 
     def test_noname_tag(self):
         self.assertMarkdownRenders('<span></></span>', '<p><span>&lt;/&gt;</span></p>')
+
+    def test_markdown_nested_in_inline_comment(self):
+        self.assertMarkdownRenders(
+            'Example: <!-- [**Bold link**](http://example.com) -->',
+            '<p>Example: <!-- <a href="http://example.com"><strong>Bold link</strong></a> --></p>'
```
facelessuser, line 41: Are links converted in comments because comments are processed before links are? I would have thought that the content in comments would skip the parser, at least ideally.

waylan: In short, we are copying the behavior of the reference implementation. More details [here](https://github.com/Python-Markdown/markdown/issues/1571#issuecomment-3596757609). Personally, I would have expected no parsing as well, but after looking into it I recall working this out before.

facelessuser: Gotcha

facelessuser (approved):

PR #1575, opened 2025-12-11
Make the docs build successfully with mkdocstrings-python 2.0

mitya57:
This fixes CI failure which can be seen [here](https://github.com/Python-Markdown/markdown/actions/runs/20112704275/job/57714086708).

In mkdocstrings/python@5ebeda6fce1b1bc7cb3f5e27a5a90ac394a3de0c (1.14 release), two changes happened:
- `import` was renamed to `inventories`;
- passing extra options directly under `options` was deprecated, and they should be now passed in `options.extra`.

The old config kept working in 1.x releases, but since mkdocstrings/python@c10afdb98d590a23c8840c7c0cdd6c358094dc2c (2.0 release), it no longer works.

I bumped mkdocstrings dependency to 0.28.3, because it in turn depends on mkdocstrings-python new enough to contain the first commit (see mkdocstrings/mkdocstrings@ba9003e96c8e5e01900743d5c464cbd228d732f4).

---

While working on this, I noticed that links to code are broken in the current release, because the tag name is `3.10.0` but `_get_version()` returns just `3.10`. We should make this consistent, but that’s a separate issue.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,7 @@ See the [Contributing Guide](contributing.md) for details.
 ### Fixed
 
 * Ensure nested elements inside inline comments are properly unescaped (#1571).
+* Make the docs build successfully with mkdocstrings-python 2.0 (#1575).
 
 ## [3.10.0] - 2025-11-03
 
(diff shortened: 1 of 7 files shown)


waylan (approved):
Thanks for working through this. I just haven't had the time to look into it myself.

PR #1579, opened 2025-12-28
Fix infinite loop when text contains multiple unclosed comments

### LINKED ISSUE #1578: Infinite loop when inline code contains two or more ```</```
Text containing two or more `` `</` `` sequences causes an infinite loop.

```python
import markdown
markdown.markdown("`</` and `</`")  # Hangs
```

The same can also be reproduced with opening a second html comment without closing the first, eg:

```python
import markdown
markdown.markdown("<!-- <!--")  # Hangs
```

## Versions

- **Affected:** markdown 3.8.2+ (introduced in 9980cb5b)
- Tested on Python 3.10, 3.11, 3.12, 3.13, 3.14

## Cause

Commit 9980cb5b added handling for unclosed comments to fix Python 3.14 compatibility. The intent was: when an unclosed comment is detected, emit `<` as literal text and restart parsing. The implementation adds:

```python
def handle_comment(self, data: str):
    if self.rawdata[i:i + 3] != '-->':  # unclosed comment?
        self.handle_data('<')
        self.override_comment_update = True
        return

def updatepos(self, i: int, j: int) -> int:
    if self.override_comment_update:
        self.override_comment_update = False
        i = 0  # reset to start
        j = 1
    return super().updatepos(i, j)
```

The problem: resetting to position 0 restarts parsing from the beginning. With two `</\`` sequences, the parser:

1. Hits first `</`` → resets to position 1 (after emitting `<`)
2. Continues, hits second `</`` at position 9
3. Resets to position 1 again
4. Second `</`` is still at position 9 → infinite loop

Why `</`` triggers this: per HTML5, `</` followed by a non-letter is a "bogus comment" (Python's html.parser uses `endtagopen = re.compile('</[a-zA-Z]')`). The stdlib `HTMLParser.goahead()` flow is:

1. [`endtagopen.match()` fails](https://github.com/python/cpython/blob/main/Lib/html/parser.py#L262) for `</` not followed by a letter, so `parse_endtag()` is never called.
2. The [fallback "bogus comment" path](https://github.com/python/cpython/blob/main/Lib/html/parser.py#L265-L279) calls `handle_comment(rawdata[i+2:])` (i.e., data after `</`).
3. The custom [`handle_comment()`](https://github.com/Python-Markdown/markdown/blob/89112c293f7b399ae8808f3a06306f46601e9684/markdown/htmlparser.py#L273-L281) interprets this as "unclosed comment" (no `-->`) and sets `override_comment_update = True`.
4. [`updatepos()`](https://github.com/Python-Markdown/markdown/blob/89112c293f7b399ae8808f3a06306f46601e9684/markdown/htmlparser.py#L283-L288) then resets to `(0, 1)`, which rewinds the scan to the start of the buffer.

So the loop is caused by the global rewind in `updatepos()`, not by bogus comments themselves. The fix should preserve the "emit `<` and retry" intent but only rewind to just after the `<` that triggered the bogus comment (e.g., store the triggering index and advance to `index + 1`), instead of resetting to the start of the buffer.

btucker:
The updatepos() override was resetting the parser position to 0, causing infinite loops when multiple bogus/unclosed comments appeared in the input (e.g., `</` followed by non-standard characters or `<!--` ). Now tracks the position of the triggering `<` and advances past it instead of rewinding to the start of the buffer.

Fixes #1578

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -16,6 +16,7 @@ See the [Contributing Guide](contributing.md) for details.
 
 * Ensure nested elements inside inline comments are properly unescaped (#1571).
 * Make the docs build successfully with mkdocstrings-python 2.0 (#1575).
+* Fix infinite loop when multiple bogus or unclosed HTML comments appear in input (#1578).
 
 ## [3.10.0] - 2025-11-03
 
--- a/markdown/htmlparser.py
+++ b/markdown/htmlparser.py
@@ -111,6 +111,7 @@ class HTMLExtractor(htmlparser.HTMLParser):
         self.lineno_start_cache = [0]
 
         self.override_comment_update = False
+        self.override_comment_start = 0
 
         # This calls self.reset
         super().__init__(*args, **kwargs)
@@ -124,6 +125,7 @@ class HTMLExtractor(htmlparser.HTMLParser):
         self._cache: list[str] = []
         self.cleandoc: list[str] = []
         self.lineno_start_cache = [0]
+        self.override_comment_start = 0
 
         super().reset()
 
@@ -276,6 +278,8 @@ class HTMLExtractor(htmlparser.HTMLParser):
         i = self.line_offset + self.offset + len(data) + 4
         if self.rawdata[i:i + 3] != '-->':
             self.handle_data('<')
+            pos = self.line_offset + self.offset
+            self.override_comment_start = pos - 1 if self.rawdata[pos - 1:pos] == '<' else pos
             self.override_comment_update = True
             return
         self.handle_empty_tag('<!--{}-->'.format(data), is_block=True)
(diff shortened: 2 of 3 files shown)


facelessuser:
Good catch! One question though, I'm not following how this fixed this case: 

```py
markdown.markdown("`</` and `</`")
```

It seems the reset logic is only triggered when hitting `-->`. In the above example, that isn't happening. I'm sure I'm just missing something.

facelessuser:
Never mind, it's `!= '-->'` 🤦🏻

facelessuser (approved):
LGTM. Changelog should get updated, though.

btucker:
Yeah, exactly. `</` being treated as the start of a comment was news to me.

PR #1581, opened 2025-12-30
Revert "Allow reference links with backticks"

facelessuser:
This reverts commit 07dfa4eb43f7a2ab3181b4f842a960a03a6c1221.

This does not fully work as advertised.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -17,8 +17,6 @@ See the [Contributing Guide](contributing.md) for details.
 * Ensure nested elements inside inline comments are properly unescaped (#1571).
 * Make the docs build successfully with mkdocstrings-python 2.0 (#1575).
 * Fix infinite loop when multiple bogus or unclosed HTML comments appear in input (#1578).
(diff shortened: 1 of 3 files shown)


facelessuser:
@waylan I am requesting this gets reverted. This was done with only a little testing, and once more testing is done, it is revealed this does not fully work. For instance, this works:

```
[`Text` and `Test`]

[`Text` and `Test`]: http://example.com
```

But this doesn't:

```
Start [`Text` and `Test`] End

[`Text` and `Test`]: http://example.com
```

There are some incorrect assumptions being made in the original commit that do not hold up.

facelessuser:
To be clear, an assumption was made in the code change that assumes the Backticks processor will get the reference, enclosed with `[]` and will be able to reject it. But this only works in the very simple case where ``[`Text`]`` is the only text. As soon as you add any content before or after the reference, the logic no longer works. While the provided tests do work, in the real world, references will be embedded in paragraphs, and this will not actually solve the issue for most people.

Issue #1590, opened 2026-02-02
MSO conditional comments truncate table content in 3.10.1


amureki:
Hey there!
We discovered a regression in the new version of the library (3.10.1):
HTML content after MS Office conditional comments is truncated when inside a `<table>` element.

## Reproduction

```python
import markdown

html = """
<table>
<!--[if mso]>-->
<td>keep</td>
<!--<![endif]-->
<td>also keep</td>
</table>
"""

markdown.markdown(html)
```

**Expected:**
```html
<table>
<!--[if mso]>-->
<td>keep</td>
<!--<![endif]-->
<td>also keep</td>
</table>
```

**Actual:**
```html
<table>
</table>
```

## Context

These are standard MS Office/Outlook conditional comments used in HTML emails for Outlook compatibility.

## Regression

Works in 3.10, broken in 3.10.1.
I think, it was introduced in f925349

amureki:
I created a PR with a regression test in https://github.com/Python-Markdown/markdown/pull/1591

waylan:
This appears to only be an issue inside `<table>` for some reason. Although, to be fair, I didn't try every possible tag so it might occur in some other scenarios.

```
>>> src = '''
... <div>
... <!--[if mso]>-->
... <div>bar</div>
... <!--<!endif]-->
... <div>baz</div>
... </div>
... '''
>>> markdown.markdown(src)
'<div>\n<!--[if mso]>-->\n<div>bar</div>\n<!--<!endif]-->\n<div>baz</div>\n</div>'
>>> src = '''
... <table>
... <!--[if mso]>-->
... <td>foo</td>
... <!--<!endif]-->
... <td>bar</td>
... </table>
... '''
>>> markdown.markdown(src)
'<table>\n</table>'
```

facelessuser:
😢 These cases are getting a bit exhausting. At least, since it is table-specific, it is likely to be easier to get to the bottom.

facelessuser:
3rd, 4th, or 5th time's a charm (lost count) 🤞🏻. PR in #1592.

facelessuser:
@waylan, don't release anything. I want to dig more into this.

facelessuser:
I think I have some more fixes coming.

facelessuser:
[...]

PR #1591, opened 2026-02-02
Ref #1590 -- Add regression test for MSO conditional comments bug

amureki:


facelessuser:
We'll likely just add the regression test with the solution.

amureki:
@facelessuser yeah, this meant only to be a demo of a problem, no further action on this PR needed!

PR #1593, opened 2026-02-03
More reliable fix for `</`

facelessuser:


markdown/htmlparser.py
```diff
@@ -48,6 +47,8 @@
 # throwing it away. When we see it, we will process it as data.
 htmlparser.starttagopen = re.compile('<[a-zA-Z]|</>')
 
+htmlparser.endtagopen = re.compile('</[a-zA-Z]|</')
```
mitya57: Maybe it can be simplified to `'</[a-zA-Z]?'`?

facelessuser: Good call

facelessuser: Done

mitya57 (approved):

waylan (approved):

PR #1600, opened 2026-04-22
fix(htmlparser): handle zipimporter without exec_module for embedded Python

### LINKED ISSUE #1132: Monkeypatching html.parser fails when importing with zipimporter
When trying to monkeypatch a copy of `html.parser`, markdown fails when `html.parser` comes from a zip file and an instance of `zipimporter` is used as the loader.

## To reproduce

1. Download an embeddable package version of Python like the latest [*Windows embeddable package (64-bit)* (Python 3.9.5)](https://www.python.org/ftp/python/3.9.5/python-3.9.5-embed-amd64.zip)
2. Unzip it
3. Copy the markdown package into the python folder
4. Run python
5. Import markdown
6. Notice the following traceback:

```
$ python
Python 3.9.5 (tags/v3.9.5:0a7dcbd, May  3 2021, 17:27:52) [MSC v.1928 64 bit (AMD64)] on win32
>>> import markdown
  File "C:\Users\Rémy Roy\Projects\CDDA-Game-Launcher\build\test\markdown\__init__.py", line 29, in <module>
    from .core import Markdown, markdown, markdownFromFile  # noqa: E402
  File "C:\Users\Rémy Roy\Projects\CDDA-Game-Launcher\build\test\markdown\core.py", line 27, in <module>
    from .preprocessors import build_preprocessors
  File "C:\Users\Rémy Roy\Projects\CDDA-Game-Launcher\build\test\markdown\preprocessors.py", line 29, in <module>
    from .htmlparser import HTMLExtractor
  File "C:\Users\Rémy Roy\Projects\CDDA-Game-Launcher\build\test\markdown\htmlparser.py", line 31, in <module>
    spec.loader.exec_module(htmlparser)
AttributeError: 'zipimporter' object has no attribute 'exec_module'
```

The main issue is that the `zipimporter` loader does not have an `exec_module` method.

armorbreak001:
When Python is installed as an embeddable package (e.g. Windows embeddable zip), html.parser is loaded via zipimporter which does not have an exec_module method. This causes an AttributeError on import. The fix adds a hasattr check before calling exec_module(), falling back to importlib.import_module() for loaders that do not support it. Fixes #1132

### DIFF
--- a/markdown/htmlparser.py
+++ b/markdown/htmlparser.py
@@ -40,7 +40,11 @@ commentclose = re.compile(r'--!?>')
 # Users can still do `from html import parser` and get the default behavior.
 spec = importlib.util.find_spec('html.parser')
 htmlparser = importlib.util.module_from_spec(spec)
-spec.loader.exec_module(htmlparser)
+if hasattr(spec.loader, 'exec_module'):
+    spec.loader.exec_module(htmlparser)
+else:
+    import importlib
+    htmlparser.__dict__.update(importlib.import_module('html.parser').__dict__)
 sys.modules['htmlparser'] = htmlparser
 
 # This is a hack. We are sneaking in `</>` so we can capture it without the HTML parser


waylan:
According to my comment [here](https://github.com/Python-Markdown/markdown/issues/1132#issuecomment-1222395487) this issue was fixed upstream for Python 3.10. As we no longer support anything lower than Python 3.10, that would suggest that this is no longer an issue and there is no need to do anything (except close the open issue).

If, however, this is still an unresolved issue upstream, then any fix here would need tests to demonstrate both that the fix works as expected (and fails without the changes) and that the user can still import the standard (not monkeypatched) html lib.

waylan:
In #1132 it has been confirmed that this issue was fixed upstream. Therefore, I am closing this as unnecessary.

PR #1609, opened 2026-06-24
fix: resolve reference links when label contains code spans (fixes #495)

### LINKED ISSUE #495: Reference links don't work if the link label contains backticks
# Reference Links with Formatting Test

An [_inline_ link **can** contain various `Markdown` formatting.](https://github.com/)

A reference link [works when you don't have any formatting](https://example.com/).

A [reference link _with_ formatting](https://example.com/ref-link-with-formatting) also renders properly.

If that [formatting contains `code` backticks](https://example.com/ref-link-with-code) then it should, but the Python renderer doesn't handle it correctly.

Source:

```
# Reference Links with Formatting Test

An [_inline_ link **can** contain various `Markdown` formatting.](https://github.com/)

A reference link [works when you don't have any formatting][].

[works when you don't have any formatting]: https://example.com/

A [reference link _with_ formatting][] also renders properly.

[reference link _with_ formatting]: https://example.com/ref-link-with-formatting

If that [formatting contains `code` backticks][] then it should, but the Python renderer doesn't handle it correctly.

[formatting contains `code` backticks]: https://example.com/ref-link-with-code
```

Output of `python -m markdown` with Markdown 2.6.2 module (verified with Python 2.7.12 and Python 3.5.2 both on Linux):

```
<h1>Reference Links with Formatting Test</h1>
<p>An <a href="https://github.com/"><em>inline</em> link <strong>can</strong> contain various <code>Markdown</code> formatting.</a></p>
<p>A reference link <a href="https://example.com/">works when you don't have any formatting</a>.</p>
<p>A <a href="https://example.com/ref-link-with-formatting">reference link <em>with</em> formatting</a> also renders properly?</p>
<p>If that [formatting contains <code>code</code> backticks][] then it doesn't?</p>[mduo13@Manatsu ~]$ cat md_reflink_test.md | python -m markdown
<h1>Reference Links with Formatting Test</h1>
<p>An <a href="https://github.com/"><em>inline</em> link <strong>can</strong> contain various <code>Markdown</code> formatting.</a></p>
<p>A reference link <a href="https://example.com/">works when you don't have any formatting</a>.</p>
<p>A <a href="https://example.com/ref-link-with-formatting">reference link <em>with</em> formatting</a> also renders properly.</p>
<p>If that [formatting contains <code>code</code> backticks][] then it should, but the Python renderer doesn't handle it correctly.</p>
```

As you can see, in the Python output, it retains the [brackets around link label][] in the output HTML.

gaoflow:
Reference links whose label contains backtick code spans
(`[text with \`code\`][]`) fail to resolve even when the
matching reference definition exists. The raw bracket syntax
appears in the output instead of a resolved `<a>` tag.

Fixes #495 (open since 2016).

**Before:**
```md
[formatting contains `code` backticks][]
[formatting contains `code` backticks]: https://example.com/ref
→ <p>[formatting contains <code>code</code> backticks][]</p>
```

**After:**
```html
→ <p><a href="https://example.com/ref">formatting contains
  <code>code</code> backticks</a></p>
```

**Root cause:** Inline processing runs in priority order. The backtick
processor (priority 190) replaces code spans with opaque placeholders
before the reference processor (priority 170) attempts lookup. The
stored reference definition key contains literal backtick text,
creating a mismatch against the placeholder-containing ID.

**Fix (+38 lines in 2 files):**
1. `inlinepatterns.py` — `_unescapeId()` method resolves inline
   placeholders back to plain text before reference lookup, using
   `itertext()` to strip markup from stashed elements.
2. `blockprocessors.py` — `ReferenceProcessor.run()` also stores
   each definition under a backtick-stripped key for normalization.

The sibling inline-link form `[text with \`code\`](url)` works
correctly — this fix brings reference links to parity.

761/761 tests pass (0 regressions).

This pull request was prepared with the assistance of AI, under my
direction and review.

### DIFF
--- a/markdown/blockprocessors.py
+++ b/markdown/blockprocessors.py
@@ -591,6 +591,13 @@ class ReferenceProcessor(BlockProcessor):
             link = m.group(2).lstrip('<').rstrip('>')
             title = m.group(5) or m.group(6)
             self.parser.md.references[id] = (link, title)
+            # Also store under a backtick-stripped key so that inline
+            # reference lookups work when the label contains code spans.
+            # Backtick code spans are processed before reference resolution,
+            # and the stashed code element does not preserve the backtick count.
+            id_stripped = id.replace('`', '')
+            if id_stripped != id:
+                self.parser.md.references[id_stripped] = (link, title)
             if block[m.end():].strip():
                 # Add any content after match back to blocks as separate block
                 blocks.insert(0, block[m.end():].lstrip('\n'))
--- a/markdown/inlinepatterns.py
+++ b/markdown/inlinepatterns.py
@@ -879,6 +879,35 @@ class ReferenceInlineProcessor(LinkInlineProcessor):
 
     RE_LINK = re.compile(r'\s?\[([^\]]*)\]', re.DOTALL | re.UNICODE)
 
+    def _unescapeId(self, id: str) -> str:
+        """Unescape inline placeholders in a reference ID for lookup.
+
+        Inline processing (e.g. backtick code spans at priority 190) runs before
+        reference resolution (priority 170). Code spans are replaced with
+        placeholders that would not match the raw-text reference definition key.
+        This method reverses that transformation for reference matching.
+
+        Backtick delimiters are stripped during normalization since the backtick
+        count is not preserved in the stashed element. Reference definitions are
+        stored under a similarly backtick-stripped key.
+        """
+        try:
+            stash = self.md.treeprocessors['inline'].stashed_nodes
+        except (KeyError, AttributeError):
+            return id
+
+        def _replace_placeholder(m: re.Match[str]) -> str:
+            sid = m.group(1)
+            if sid in stash:
+                value = stash.get(sid)
+                if isinstance(value, str):
+                    return value
+                else:
+                    return ''.join(value.itertext())
+            return m.group(0)
+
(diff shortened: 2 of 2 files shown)


waylan:
Thank you for your submission. A few things to note.

First of all, this does not include any tests to verify that it works as expected.

Second, thank you for noting that AI assisted. However, it is not clear if you mean that AI wrote the code and you reviewed it or that AI assisted in finding where the error occurred and you wrote the code. This hasn't really come up yet, so we don't have a clear policy, but my inclination is to only accept human-written code.

waylan:
With a cursory look at the code, it appears that it is simply removing backticks from reference link labels. While that may seem to work in simple cases, it is **_not_** a viable solution. Consider this example:

``` markdown
[link with `code`] and [link with code]

[link with `code`] https://example.com
[link with code] https://notexample.com
```

If I understand the code correctly; both reference links would resolve to `https://example.com` which would be wrong.

The correct solution is to just require the document author to use better references. For example:

``` markdown
[link with `code`][1] and [link with code][2]

[1] https://example.com
[2] https://notexample.com
```

And that works without any changes to the code.

gaoflow:
Thanks for taking a look. You're right: normalizing the label by stripping code-span backticks can conflate distinct references such as `[link with `code`]` and `[link with code]`, which is not a safe fix. I'll close this PR rather than push this direction further.

PR #1611, opened 2026-06-26
Create AI Policy

### LINKED ISSUE #1610: AI Policy?
We have received a few recent PRs which I suspect were AI generated. It is likely we will only continue to receive more. I am not opposed to AI in general, but I do have some thoughts about what I expect for this project. All of this hangs on the fact that I have limited time to devote to this project as a volunteer. The more that my valuable time is wasted, the more the project suffers. Therefore, in the interest of maintaining a high-quality project, I feel that we need to create an AI policy.

First, I would prefer for the documentation to not contain what readers would perceive as an "AI" voice. That said, AI can be a real asset to contributors for whom English is not their first language.  My inclination is to suggest that they put forth their best (human) effort, and I would be happy to spend some time helping them improve their English prose. I would not consider that to be a waste of my time. However, it would be a waste of my time to edit AI generated prose.

Second, I would generally prefer that contributors be able to fully understand the code they are submitting. If a contributor cannot fully explain what their code is doing, then it is a waste of my time to review it. Yes, there are some very skilled programmers who may be able to provide an AI model with a detailed spec and have it generated good code which they have competently reviewed. However, it seems that most open-source submissions which are AI generated are slop from inexperienced contributors who are not able to understand the code, let alone understand why it is slop. Reviewing that slop is a waste of my time.

Therefore, I am inclined to implement a policy that all contributions must be human generated. If a contributor wants to use AI to assist with debugging, that's fine, but the code, documentation, commit message, and comments must be human generated. That said, I would be open to exceptions for specific known developers who I trust so long as they are upfront about their AI use.

I think that if we do this, we not only need to document the policy, but include a PR template which includes a checkbox where the contributor must explicitly indicate that no AI was used to generate the included code, documentation, commit message, and comments. I have generally been opposed to templates because I find that they are often too restrictive as they cannot address every possible situation. However, most recent PRs seem to completely ignore our contributing guidelines. A simple set of checkboxes might be sufficient.

Thoughts?

waylan:
Any feedback is welcome. I think this addresses our concerns in the least restrictive way possible.

Commit Message (which memorializes our thinking):

> The motivation for adopting this policy is to avoid wasting the maintainers' time. We work on this project as volunteers and it is an unproductive waste of time to interact with contributors who do not understand their contributions.
> 
> While we generally would prefer to receive contributions which are not AI generated, we understand that such tools can provide value in limited ways. Therefore, we have attempted to create a policy which allows use of such tools while also holding the contributor accountable. Hopefully this accomplishes our goal. Closes #1610.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,15 @@
+#### Description
+<!-- Provide a concise overview of the issue or rationale behind the proposed changes. -->
+
+#### AI Assistance Disclosure
+<!-- Please select exactly ONE of the following: -->
+- [ ] **No AI tools were used** in preparing this PR.
+- [ ] **If AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.
+
+#### Checklist
+- [ ] This PR follows the [contribution guidelines](https://python-markdown.github.io/contributing/).
+- [ ] The code follows the [Code Style Guide](https://python-markdown.github.io/contributing/#code-style-guide).
+- [ ] The commit message follows the [Commit Message Style Guide](https://python-markdown.github.io/contributing/#commit-message-style-guide).
+- [ ] I have added or updated relevant docs, including release notes if applicable which follow the [Documentation Style Guide](Documentation Style Guide).
+- [ ] I have added or updated relevant tests.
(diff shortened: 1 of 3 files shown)


mitya57 (approved):
I like this very much.

facelessuser:
LGTM

facelessuser (approved):

PR #1621, opened 2026-08-11
Fix quadratic rendering time for many inline links

### LINKED ISSUE #1619: Super-linear (quadratic) rendering on many inline links: '[link](x)' * N
python-Markdown renders repeated inline links super-linearly (quadratic). measure: `'[link](x)'*512`=0.02s, `*2048`=0.25s, `*8192`=3.75s (4x input ~15x time). A ~300KB link-heavy document takes ~60s. No extant advisory covers a link-reference ReDoS (only GHSA-5wmx-573v-2qwq UncaughtException, PYSEC-2026-89). Repro: `Markdown().convert('[link](x)'*8192)`.

aryansk:
#### Description

Rendering a paragraph containing many inline links took quadratic time:
each match replaced the link with a placeholder and then rescanned the
whole remaining text from index 0, so N links cost O(N²) pattern work.
`'[link](x)' * 8192` took ~7s before this change and ~0.3s after.

`__applyPattern` now returns the index just past the inserted placeholder
so the next scan starts at the unprocessed tail.

Fixes #1619.

#### AI Assistance Disclosure

Claude (via Command Code) assisted with implementation and testing; the change was fully reviewed and verified by a human.

#### Checklist

- [x] I have added or updated relevant docs, including release notes if applicable.

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -12,10 +12,14 @@ See the [Contributing Guide](contributing.md) for details.
 
 ## [Unreleased]
 
+### Changed
+
+* Inline processors now resume searching after the previous match, improving
+  performance for repeated inline patterns (#1619).
+
 ### Fixed
 
 * Fix an issue with excessive backtracking when matching inline code blocks (#1617).
-* Fix quadratic rendering time when a paragraph contains many inline links (#1619).
 
 ## [3.10.3] - 2026-07-30
 
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -15,6 +15,7 @@ See the [Contributing Guide](contributing.md) for details.
 ### Fixed
 
 * Fix an issue with excessive backtracking when matching inline code blocks (#1617).
+* Fix quadratic rendering time when a paragraph contains many inline links (#1619).
 
 ## [3.10.3] - 2026-07-30
 
--- a/markdown/treeprocessors.py
+++ b/markdown/treeprocessors.py
@@ -323,12 +323,17 @@ class InlineProcessor(Treeprocessor):
         placeholder = self.__stashNode(node, pattern.type())
 
         if new_style:
+            # Return the index just past the inserted placeholder so the
+            # next call scans only the unprocessed tail. Scanning from 0
+            # after every match makes repeated inline patterns quadratic.
             return "{}{}{}".format(data[:start],
-                                   placeholder, data[end:]), True, 0
+                                   placeholder, data[end:]), True, start + len(placeholder)
         else:  # pragma: no cover
             return "{}{}{}{}".format(leftData,
                                      match.group(1),
-                                     placeholder, match.groups()[-1]), True, 0
+                                     placeholder, match.groups()[-1]), True, (
+                len(leftData) + len(match.group(1)) + len(placeholder)
+            )
 
     def __build_ancestors(self, parent: etree.Element | None, parents: list[str]) -> None:
         """Build the ancestor list."""
--- a/tests/test_syntax/inline/test_links.py
+++ b/tests/test_syntax/inline/test_links.py
@@ -21,8 +21,6 @@ License: BSD (see LICENSE.md for details).
 
 from markdown.test_tools import TestCase
 
-import time
-
 
 class TestInlineLinks(TestCase):
 
@@ -436,21 +434,3 @@ class TestReferenceLinks(TestCase):
                 """
             )
         )
-
-    def test_many_repeated_links(self):
-        # Regression test for #1619: rendering many inline links in one
-        # paragraph used to rescan the unprocessed text from the start after
-        # every match, making conversion quadratic in the number of links.
-        from markdown import Markdown
-
-        text = "[link](x)" * 8192
-        start = time.monotonic()
-        html = Markdown().convert(text)
-        elapsed = time.monotonic() - start
-
-        # The old implementation takes several seconds (or more on slow CI)
-        # for this input; the linear implementation finishes well under a
-        # second on any machine. Allow a generous ceiling to avoid flakes.
-        self.assertLess(elapsed, 5)
-        self.assertEqual(html.count("<a "), 8192)
-        self.assertEqual(html.count("</a>"), 8192)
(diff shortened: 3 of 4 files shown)


tests/test_syntax/inline/test_links.py
```diff
@@ -434,3 +436,21 @@ def test_ref_round_brackets(self):
...
+        start = time.monotonic()
+        html = Markdown().convert(text)
+        elapsed = time.monotonic() - start
+
+        # The old implementation takes several seconds (or more on slow CI)
+        # for this input; the linear implementation finishes well under a
+        # second on any machine. Allow a generous ceiling to avoid flakes.
+        self.assertLess(elapsed, 5)
+        self.assertEqual(html.count("<a "), 8192)
+        self.assertEqual(html.count("</a>"), 8192)
```
waylan: I'm not sure about measuring time in a test. That assumes certain capabilities of the system which is running the test. I realize you built in a cushion, but it is still possible that this test could fail just because a system is really slow. Seems like there could be a better approach here.

markdown/treeprocessors.py
waylan, line 1: This changes the API for all inline patterns and processors. @facelessuser I think you have a better handle on this part of the code. Any input on that change here?

docs/changelog.md
```diff
@@ -15,6 +15,7 @@ See the [Contributing Guide](contributing.md) for details.
 ### Fixed
 
 * Fix an issue with excessive backtracking when matching inline code blocks (#1617).
+* Fix quadratic rendering time when a paragraph contains many inline links (#1619).
```
waylan: As this changes the API for all inline patterns and processors, it is more than a fix.  It is a change and should be identified as such in the changelog (A `Changed` section should be added and the comment moved there.). This will require the next release to be at least a point release (3.11), not a bugfix release.

waylan (changes requested):
Thanks for working on this. I have a few concerns as outlined below.

tests/test_syntax/inline/test_links.py
mitya57: I agree with @waylan. In Debian, we run the tests on 9 different CPU architectures, and some of them are really slow (like software-emulated riscv64). The tests should not assume anything about time.

waylan: In fact, the [pypy311](https://github.com/Python-Markdown/markdown/actions/runs/31495311819/job/93800281532?pr=1621) test is failing for this very reason in our CI tests for this PR.

aryansk: Agreed. I removed the elapsed-time assertion and replaced it with a deterministic test that verifies repeated matches advance the compiled-pattern search offset. Commit b745bba; the full unittest suite passes with 1,089 tests and 13 skipped.

aryansk: The timing-based test is gone in b745bba. The replacement checks the search offset directly, so it does not depend on machine speed or architecture; the full unittest suite passes with 1,089 tests and 13 skipped.

aryansk: Thanks for confirming the failure. I removed the timing-based test in b745bba and replaced it with a deterministic search-offset assertion, so the PyPy slowdown no longer affects the test.

markdown/treeprocessors.py
aryansk: Thanks. I kept the search-index behavior change and documented it under Unreleased > Changed in b745bba, treating it as an inline-processor API/behavior change rather than a Fixed-only entry.

docs/changelog.md
aryansk: Done in b745bba: #1619 is now under Unreleased > Changed, and the timing regression test was replaced with a deterministic search-offset test.

aryansk:
Thanks for the detailed review. Commit b745bba addresses the points raised: it removes the wall-clock regression test, replaces it with a deterministic search-offset assertion, and moves #1619 under Unreleased > Changed to document the inline-processor behavior/API impact. The full unittest suite passes with 1,089 tests and 13 skipped; changed-file flake8 and git diff --check pass.

waylan:
Thank you for addressing my concerns. I am going to let @facelssuser review this as he has a better grasp of the affected code.

aryansk:
Thanks, @waylan. I appreciate the follow-up; I’ll wait for @facelssuser’s review of the affected code.

facelessuser:
I will hopefully get to this sometime this week. I've become somewhat exhausted with the onslaught of everyone pointing AI at every project on the internet. I'm already dealing with similar issues on numerous projects I support.

aryansk:
Understood. I will leave the draft unchanged while you assess it; no further follow-up is needed from me right now.

markdown/treeprocessors.py
facelessuser: I do have some mild concerns about this change. We are assuming that because the API changed here without breaking things, others using this API won't be broken.

When we changed the API here in the first time, we did so in a way that was non-breaking, one you could opt into. This was done by providing a different class that could be checked.

It is possible that we may need to employ some method ot indicate this is a different version of the new way. I really need to do some testing to understand the implications of the change. I will have to do some testing with this over in Pymdown Extensions, where we have a number of plugins using the new style, so I can get a better idea of what the impact here is.

facelessuser: It seems that all the tests over on Pymdown Extensions passed with these changes, so that is a good sign.

waylan: @facelessuser did you intend to provide more input on this? Itis not clear to me if you had finished reviewing this.

facelessuser: Yeah, I've just been busy. My initial pass indicated that I don't "think" it'll break anything.

facelessuser: I can try to formally finish the review soon.

facelessuser (approved):

markdown/treeprocessors.py
facelessuser: I've approved the review; I think this makes sense.

NOVATechnocrat:
Thanks for all of your your hard work, I know you all are busy. Let me know if I can assist.

PR #1624, opened 2026-08-27
fix: treat custom element names as block-level HTML

### LINKED ISSUE #1246: Extension md_in_html does not recognize tags with hyphens
Web components are custom HTML components that are required to have `-` in their names. This breaks current HTML handling since these elements are not considered. IMHO they should be treated the same as `<div>` ("block" elements, if I'm not mistaken).

The following was tested in current `main` with the extension `md_in_html` active. 

**input**
```
<a-b>

asdf

</a-b>
```

**output**:
```
<p><a-b></p>
<p>asdf</p>
<p></a-b></p>
```

**expected**:
```
<a-b>
<p>asdf</p>
</a-b>
```

I went through the code and might know how to add this, but I would like the maintainers' input before proceeding.

BetterAndBetterII:
## Summary

Custom element names are required to contain a hyphen. `is_block_level()` and `md_in_html` previously only recognized a fixed list of block tags, so a standalone `<a-b>…</a-b>` was wrapped in `<p>` and the `markdown` attribute on those tags was ignored.

Treat any tag name that contains a hyphen as block-level (same treatment as `<div>`). Inner Markdown still requires the `markdown` / `markdown="1"` attribute. Inline use mid-paragraph stays span-level.

## Test plan

- [x] `python -m unittest tests.test_syntax.blocks.test_html_blocks.TestHTMLBlocks.test_raw_custom_element tests.test_syntax.extensions.test_md_in_html.TestMdInHTML.test_custom_element_markdown_attr tests.test_syntax.extensions.test_md_in_html.TestMdInHTML.test_custom_element_md1 tests.test_syntax.extensions.test_md_in_html.TestMdInHTML.test_custom_element_inline`
- [x] Full suite previously green under grok (`python -m unittest discover tests`)

Fixes #1246

### DIFF
--- a/markdown/core.py
+++ b/markdown/core.py
@@ -303,12 +303,15 @@ class Markdown:
         """
         Check if the given `tag` is a block level HTML tag.
 
-        Returns `True` for any string listed in `Markdown.block_level_elements`. A `tag` which is
-        not a string always returns `False`.
+        Returns `True` for any string listed in `Markdown.block_level_elements`.
+        Also returns `True` for valid custom element names (tag names that contain
+        a hyphen; see https://html.spec.whatwg.org/#valid-custom-element-name).
+        A `tag` which is not a string always returns `False`.
 
         """
         if isinstance(tag, str):
-            return tag.lower().rstrip('/') in self.block_level_elements
+            tag = tag.lower().rstrip('/')
+            return tag in self.block_level_elements or '-' in tag
         # Some ElementTree tags are not strings, so return False.
         return False
 
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -34,6 +34,17 @@ if TYPE_CHECKING:  # pragma: no cover
     from markdown import Markdown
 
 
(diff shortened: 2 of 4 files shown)


waylan:
If I am not mistaken, this is not the fix suggested in #1246. This accepts any tag name which contains a hyphen, whereas the fix discussed in #1246 was to allow the user to add specific tags and only those tags would be recognized as block-level. As noted, users can already add tags to the list of recognized tags. However, due to a bug, the md-in-html extension does not recognize them.

Note that, [as documented](https://python-markdown.github.io/extensions/#:~:text=All%20of%20the%20included%20extensions%20are%20in%20maintenance%20mode%2E), all first-party extensions have entered maintenance mode and are no longer accepting new features. Adding support for all hyphenated tags would be a new feature. However, fixing the bug which prevents the extension from recognizing a user provided tag (regardless of whether it was a web-component) would not be a new feature, but a bug fix (as that is how it was intended to work from the beginning). Therefore, that is the only fix we will accept. Anything more than that would require a new or forked third-part extension to implement.

PR #1626, opened 2026-09-02
fix: number nested-block footnotes in document order

### LINKED ISSUE #1561: Nested block footnotes rendered in wrong order
(**Python Markdown 3.9**; mkdocs 1.6.1 with footnotes extension; Python 3.12)

While using mkdocs I stumbled upon this problem:

Footnote anchors and footnotes themselves are rendered out of order when, in the source and in addition to other document-level footnotes, one ore more footnotes are being referenced from within a nested block, such as a list element or a quote block.

If Python Markdown is downgraded to version 3.8.2, footnotes will be rendered correctly.

Here's a reproducible Markdown case:

```
# Heading

The following footnote anchor is the first one.[^1]

## Subheading

The following footnote anchor

1. should be the second one, but mkdocs renders it in fourth place.[^2]
1. Note that every anchor still points to its correct, matching footnote.

The following footnote anchor is the third one, but will be rendered in second place. [^3]

> Now in a quote block, possibly reproducible in any other nested block. Fourth yet shown as fifth footnote anchor.[^4]

Last, fifth footnote anchor.[^5]

### Footnotes rendered out of order:

[^1]: First
[^2]: Second
[^3]: Third
[^4]: Fourth
[^5]: Fifth

```

For reference, here is what I see via `mkdocs serve`:

Image

BetterAndBetterII:
## Summary
With `USE_DEFINITION_ORDER=False`, footnote reference numbers followed the breadth-first inline parse order rather than source/document order. Refs inside lists and blockquotes therefore got the wrong superscripts (e.g. 1, 4, 2, 5, 3 instead of 1–5).

This rebuilds `footnote_order` from the already-rendered tree in document order, renumbers the superscripts, then reorders the footnote list — without changing core inline walking.

Fixes #1561

## Test plan
- [x] `python -m unittest tests.test_syntax.extensions.test_footnotes` (28 OK)
- [x] New cases: nested list/blockquote refs; nested refs where definition order ≠ document order

### DIFF
--- a/.spell-dict
+++ b/.spell-dict
@@ -50,6 +50,7 @@ GSoC
 hacky
 HeaderId
 HTTPS
+href
 html
 implementers
 InlineProcessor
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -21,6 +21,7 @@ See the [Contributing Guide](contributing.md) for details.
 
 ### Fixed
 
+* Number nested-block footnotes in document order when `USE_DEFINITION_ORDER` is `False` (#1561).
 * Fix an issue with excessive backtracking when matching inline code blocks (#1617).
 
 ## [3.10.3] - 2026-07-30
--- a/markdown/extensions/footnotes.py
+++ b/markdown/extensions/footnotes.py
@@ -448,11 +448,42 @@ class FootnoteReorderingProcessor(Treeprocessor):
     def run(self, root: etree.Element) -> None:
         if not self.footnotes.footnotes:
             return
+        # Rebuild from the tree: inline processing is not document order.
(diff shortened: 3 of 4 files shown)


waylan:
You seem to have deleted the [pull request template](https://raw.githubusercontent.com/Python-Markdown/markdown/refs/heads/master/.github/pull_request_template.md) from your submission. Could you please copy that into a comment and provide the requested information before we use our time to review your submission?

waylan:
This appears to be a bug-fix, not an enhancement. Therefore, the feature freeze on extensions does not apply. However, I will not review this any further until the requested information is provided. Specifically, we would like to know about any AI use which may or may not have contributed to this PR.

PR #1629, opened 2026-09-05
Honor tags added to block_level_elements after md_in_html loads

### LINKED ISSUE #1246: Extension md_in_html does not recognize tags with hyphens

gyanu2507:
## Summary
`md_in_html` copied `Markdown.block_level_elements` when the extractor was constructed. Tags registered later with `md.block_level_elements.append('a-b')` were therefore ignored, so custom elements never got block-level treatment or `markdown="1"` parsing.

The extractor now reads the live list. That is the change discussed in #1246 (not treating every hyphenated name as a block tag).

## Test plan
- [x] `python3 -m unittest tests.test_syntax.extensions.test_md_in_html` (198 tests)

Fixes #1246

### DIFF
--- a/docs/changelog.md
+++ b/docs/changelog.md
@@ -24,6 +24,8 @@ See the [Contributing Guide](contributing.md) for details.
 ### Fixed
 
 * Fix an issue with excessive backtracking when matching inline code blocks (#1617).
+* `md_in_html` now honors tags added to `Markdown.block_level_elements` after
+  the extension is loaded (#1246).
 
 ## [3.10.3] - 2026-07-30
 
--- a/markdown/extensions/md_in_html.py
+++ b/markdown/extensions/md_in_html.py
@@ -41,8 +41,6 @@ class HTMLExtractorExtra(HTMLExtractor):
     """
 
     def __init__(self, md: Markdown, *args, **kwargs):
-        # All block-level tags.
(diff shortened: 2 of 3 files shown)


waylan (approved):
This looks good! Unfortunately, you deleted the PR template and now we don’t have a disclosure statement from you about AI use, if any. Please add a comment.

gyanu2507:
Sorry about dropping the template.

#### AI Assistance Disclosure

I used an LLM coding assistant to help draft the change. I reviewed the diff, followed the approach discussed in #1246 (read the live `block_level_elements` list rather than treating every hyphenated name as a block tag), and ran `python3 -m unittest tests.test_syntax.extensions.test_md_in_html`.

waylan (approved):

PR #1635, opened 2026-09-10
Add Markdown renderer as superfences formatter

waylan:
#### Description

If a code block is demonstrating Markdown syntax, if can be assigned the `md-render` attribute, and both the Markdown source and HTML output will be rendered in a nested set of code blocks. This was inspired by PyMdown's custom formatter. The difference is that both the source and output are rendered from a single block, whereas PyMdown's documentation requires defining the two blocks separately. Examples of what the output looks like can be found throughout PyMdown's documentation.

The Documentation Style Guide was updated for this and various other changes related to the switch to Zensical.

#### AI Assistance Disclosure

- [x] **No AI tools were used** in preparing this PR.
- [ ] **If AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.

#### Checklist
- [x] This PR follows the [contribution guidelines](https://python-markdown.github.io/contributing/).
- [x] The code follows the [Code Style Guide](https://python-markdown.github.io/contributing/#code-style-guide).
- [x] The commit message follows the [Commit Message Style Guide](https://python-markdown.github.io/contributing/#commit-message-style-guide).
- [x] I have added or updated relevant docs, including release notes if applicable which follow the [Documentation Style Guide](Documentation Style Guide).
- [x] I have added or updated relevant tests.
- [x] I have not requested, and will not request, an automated AI review for this PR.

PR #1637, opened 2026-09-15
Add py-render custom code block formater

waylan:
#### Description

This is similar to the previous markdown renderer (#1635) except that it runs Python code blocks in an isolated environment and renders the output as a result block. The Contributing Guide has been updated with an explanation of the feature.

#### AI Assistance Disclosure

- [x] **No AI tools were used** in preparing this PR.
- [ ] **If AI tools were used**, I have disclosed which ones, and fully reviewed and verified their output.

#### Checklist
- [x] I have added or updated relevant docs, including release notes if applicable which follow the [Documentation Style Guide](https://python-markdown.github.io/contributing/#documentation-style-guide).


waylan:
Contributing-to-Python-Markdown-Python-Markdown-09-15-2026_04_16_PM

mitya57:
Are the backticks intended in `` Value of `fragment` ``, or it should be `Value of <code>fragment</code>` instead?

waylan:
Good question. I had originally tried `Value of <code>fragment</code>` but the HTML gets escaped. I then tried the backticks and left them in as a compromise. I intended to file an issue with Superfences, but forgot about it. Thanks for the reminder.

facelessuser:
I'm not sure I'm following. What is the issue?

waylan:
Raw HTML in the title of a code block is escaped and Markdown rendering does not happen. Therefore, there is no way to include any markup in a title. For a simple example consider this plain superfence code block.

```` markdown
``` text { title='Content of <code>somefile.txt</code>' }
Text goes here...
```
````

It occurs to me that the above might not work as the raw HTML might trip up the parser. However, when using a custom formatter, I want to assemble a title which includes markup and can't because it is escaped.

**Comment by facelessuser** [comment]
Yeah, I think that was because when I originally put SuperFences together, the focus was getting fenced code to render in lists and other indented objects. Everything is run in a preprocessor and the stashed. It essentially becomes invisible to the parser except for the placeholder and unpacked at the end. We waited until the end as we don't have full context of whether the fenced code we parsed was under raw HTML or not.

I haven't looked into it, but I guess it could be possible to unpack them all in the treeprocessor, as we should be past the HTML processor at that point. We should have context of `markdown=true` and such to tell whether we are under an allowable HTML tag. But that means we need to consider changing how SuperFences stores content and give it more intelligence regarding the HTML tree than it currently has now.

Right now, we store fence HTML as strings, so we'd likely need to have them store content as etree Elements, or at least optionally for those who are okay with stuff in them getting processed.

So yeah, it was intentional, with the idea that code is handled and not seen until the end, unless you specifically go in and unpack them, but it is possible we could allow fenced code to be stored as etree Elements and unpack them before the inline treeprocessor, assuming no adverse effects.

**Comment by facelessuser** [comment]
It's probably fine to create a feature request, but I'll have to dig in and see how complicated it would be to pull off allowing that to get processed normally. I imagine people want access to all the normal inline stuff, so we'd need to ensure it gets inserted back into the tree at the proper time. I make no promises until I have a chance to dig into this as I haven't refactored this area in quite some time.

**Comment by waylan** [comment]
Turns out the issue is with Pygments, not Superfences. Supfences passes the `title` to the `filename` argument of `HtmlFormatter` and that value is HTML-escaped in the HTML returned by Pygments.

``` pycon
>>> from pygments import highlight
>>> from pygments.lexers import PythonLexer
>>> from pygments.formatters import HtmlFormatter
>>> print(highlight('# Code goes here', PythonLexer(), HtmlFormatter(filename='Content of <code>file.py</code>')))
<div class="highlight"><span class="filename">Content of &lt;code&gt;file.py&lt;/code&gt;</span><pre><span></span><span class="c1"># Code goes here</span>
</pre></div>
```

The only way to work around that would be to use a placeholder which could be swapped out later.

facelessuser:
Yeah, there is some complexity in this area, so it's not something I'm committed to providing until I have time to investigate and determine all the potential issues.

It's likely possible to get something working; I'm just not 100% sure it's possible without side effects, but a thorough investigation will help make that more clear.

waylan:
Turns out it was a simple fix for me to address in my custom formatter (see [#1640](https://github.com/Python-Markdown/markdown/pull/1640/changes)). I just passed the title to `md.htmlStash.store`. I didn't even need to do the reverse swap. It was already taken care of by the normal handling of Pygment's output.

facelessuser:
Apparently, I misunderstood what we were talking about. I thought it was desired that the inline code with `` ` `` would get parsed as code. I see now we just wanted to preserve raw HTML.

